Issue #32, opened 2020-10-28
Complete all tests


rishabhpoddar:
They are mentioned as TODOs in the various test files

PR #42, opened 2020-11-03
Added tests

jscyo:
## Related Issues
- #32 

## Changes
- Added two process states for testing
- Process state checks in querier.ts and sessionRecipe.ts
- fixed typo in utils.ts
- Tests added:
   - Config tests:
      - test various inputs for appInfo 
      - test using zero, one and two recipe modules 
      - test config for session module 
   - Handshake test
     - test that once the info is loaded, it doesn't query again 
   - Middleware test
     - check that disabling default API actually disables it (for session) 
   - Querier test
     - Test that if the querier throws an error from a recipe, that recipe's ID is there 
     - Check that once the API version is there, it doesn't need to query again 
     - Check that rid is added to the header iff it's a "/recipe" || "/recipe/*" request. 
   - RecipeModuleManager tests
     - Check that querier has been inited when we call supertokens.init 
     - Check that modules have been inited when we call supertokens.init
     - Test various inputs to routing (if it accepts or not) 
     -  Test various inputs to errorHandler (if it accepts or not) 
     - If an error handler in a recipe throws an error, that error next to go to the user's error handler (done)
     - Error thrown from APIs implemented by recipes must not go unhandled (done)
     - Disable a default route, and then implement your own API and check that that gets called(done)
     - Test getAllCORSHeaders
   - Session tests
     - check if output headers and set cookies for create session is fine 
     - check if output headers and set cookies for refresh session is fine 
     - check if input cookies are missing, an appropriate error is thrown
   - SessionExpress tests
     -  check if disableDefaultImplementation is true, the default refresh API does not work - you get a 404

## Migration:
  - none


test/config.test.js
```diff
@@ -42,6 +44,220 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
...
+                    supertokens: {
+                        connectionURI: "http://localhost:8080",
+                    },
+                    appInfo: {
+                        appName: "SuperTokens",
+                        websiteDomain: "supertokens.io",
+                    },
+                    recipeList: [Session.init()],
+                });
+            } catch (err) {
```
rishabhpoddar, line 106: You are supposed to throw an error before the catch too.. Otherwise this test will pass even if no error is thrown. In which case, it's supposed to fail. Please check this in all places.

test/config.test.js
```diff
@@ -42,6 +44,220 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
...
+                },
+                appInfo: {
+                    apiDomain: "api.supertokens.io",
+                    appName: "SuperTokens",
+                    websiteDomain: "supertokens.io",
+                },
+                recipeList: [Session.init()],
+            });
+
+            assert(SuperTokens.getInstanceOrThrowError().recipeModules.length === 1);
```
rishabhpoddar, line 208: Also get the session recipe instance via `Session.getInstanceOrThrowError()` (or whatever that function is called). If this function throws an error, then it's not properly working. Same for the test right below this one

test/emailpassword/config.test.js
```diff
@@ -42,4 +44,195 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
...
+            appInfo: {
+                apiDomain: "api.supertokens.io",
+                appName: "SuperTokens",
+                websiteDomain: "supertokens.io",
+            },
+            recipeList: [EmailPassword.init()],
+        });
+
+        let emailpassword = await EmailPasswordRecipe.getInstanceOrThrowError();
+        assert(emailpassword.config.signUpFeature.formFields.length === 2);
```
rishabhpoddar, line 64: I'm sure there are more checks that you can do here. This is incomplete checking.

test/emailpassword/config.test.js
```diff
@@ -42,4 +44,195 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
...
+        });
+
+        let formFields = await EmailPasswordRecipe.getInstanceOrThrowError().config.signUpFeature.formFields;
+
+        let defaultEmailValidator = formFields.filter((f) => f.id === "email")[0].validate;
+        assert((await defaultEmailValidator("aaaaa")) === "Email is invalid");
+        assert((await defaultEmailValidator("aaaaaa@aaaaaa")) === "Email is invalid");
+        assert((await defaultEmailValidator("random  User   @randomMail.com")) === "Email is invalid");
+        try {
+            await defaultEmailValidator();
```
rishabhpoddar, line 221: I believe that this should return a string too now.. how is this test not failing?

This is a bug in the function. Please do not manipulate your tests to make them pass..

test/emailpassword/config.test.js
```diff
@@ -42,4 +44,195 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
...
+
+        let defaultPasswordValidator = formFields.filter((f) => f.id === "password")[0].validate;
+        assert(
+            (await defaultPasswordValidator("aaaaa")) ===
+                "Password must contain at least 8 characters, including a number"
+        );
+        assert((await defaultPasswordValidator("aaaaaaaaa")) === "Password must contain at least one number");
+        assert((await defaultPasswordValidator("1234*-56*789")) === "Password must contain at least one alphabet");
+        try {
+            await defaultPasswordValidator();
```
rishabhpoddar, line 234: I believe that this should return a string.. How is this test not failing?

test/emailpassword/config.test.js
```diff
@@ -42,4 +44,195 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
...
+            assert(false);
+        } catch (error) {}
+    });
```
rishabhpoddar, line 237: Also check for good case scenario!

test/emailpassword/passwordreset.test.js
```diff
@@ -13,25 +13,329 @@
...
+                websiteDomain: "supertokens.io",
+            },
+            recipeList: [EmailPassword.init()],
+        });
+        let emailpassword = await EmailPasswordRecipe.getInstanceOrThrowError();
+        const app = express();
+
+        app.use(STExpress.middleware());
+
+        app.post("/reset-password-invalid-email", async (req, res) => {
```
rishabhpoddar, line 88: # Major issue
- You do not need to create your own API (unless the tests requires that). The middleware above adds it on its own. You simply need to call the default API below. 
- The request body must be given in the request.. in set in the API.
- Requests do not have a validation function in the formFields.. Please see the types properly..

I think you will need to redo all tests that are written this way.

rishabhpoddar (changes requested):

test/emailpassword/config.test.js
```diff
@@ -42,4 +44,206 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
...
+
+        let formFields = await EmailPasswordRecipe.getInstanceOrThrowError().config.signUpFeature.formFields;
+
+        let defaultEmailValidator = formFields.filter((f) => f.id === "email")[0].validate;
+        assert((await defaultEmailValidator("aaaaa")) === "Email is invalid");
+        assert((await defaultEmailValidator("aaaaaa@aaaaaa")) === "Email is invalid");
+        assert((await defaultEmailValidator("random  User   @randomMail.com")) === "Email is invalid");
+        assert((await defaultEmailValidator("*@*")) === "Email is invalid");
+        assert((await defaultEmailValidator("validmail@gmail.com")) === undefined);
+        await defaultEmailValidator();
```
rishabhpoddar, line 237: You need to check that `await defaultEmailValidator()` does not return `undefined`. And same for with password validator (everywhere)

rishabhpoddar (changes requested):

test/emailpassword/passwordreset.test.js
```diff
@@ -143,31 +143,27 @@ describe(`passwordreset: ${printPath("[test/passwordreset.test.js]")}`, function
...
+                            value: "random@gmail.com",
+                        },
+                    ],
+                })
                 .end((err, res) => {
                     if (err) {
                         resolve(undefined);
                     } else {
-                        resolve(res.body);
+                        resolve(res);
```
rishabhpoddar, line 166: This test is entirely wrong.

rishabhpoddar (changes requested):

test/emailpassword/signinFeature.test.js
```diff
@@ -39,3 +56,1315 @@
...
+                Session.init(),
+            ],
+        });
+        const app = express();
+
+        app.use(STExpress.middleware());
+
+        app.use(STExpress.errorHandler());
+
+        await new Promise((resolve) =>
```
rishabhpoddar, line 729: Check that the output of this is OK

test/emailpassword/signinFeature.test.js
```diff
@@ -39,3 +56,1315 @@
...
```
rishabhpoddar, line 825: Check that the output of this is OK

test/emailpassword/signinFeature.test.js
```diff
@@ -39,3 +56,1315 @@
...
+                .expect(200)
+                .end((err, res) => {
+                    if (err) {
+                        resolve(undefined);
+                    } else {
+                    }
+                    resolve(JSON.parse(res.text));
+                })
+        );
+        assert(response.status === "WRONG_CREDENTIALS_ERROR");
```
rishabhpoddar, line 874: Make sure that this error is cause of the validator error by having a counter (initially 0) increment in the validator function before it returns a string, and then check that the value of that counter is 1 here.

test/emailpassword/signinFeature.test.js
```diff
@@ -39,3 +56,1315 @@
...
+            },
+            recipeList: [EmailPassword.init(), Session.init()],
+        });
+        const app = express();
+
+        app.use(STExpress.middleware());
+
+        app.use(STExpress.errorHandler());
+
+        await new Promise((resolve) =>
```
rishabhpoddar, line 977: Test that the output is OK

test/emailpassword/signinFeature.test.js
```diff
@@ -39,3 +56,1315 @@
...
+                    if (err) {
+                        resolve(undefined);
+                    } else {
+                    }
+                    resolve(JSON.parse(res.text));
+                })
+        );
+        assert(response.status === "FIELD_ERROR");
+        assert(response.formFields[0].error === "Email is invalid");
+        assert(response.formFields[0].id === "email");
```
rishabhpoddar, line 1247: it should also have another element! formFields[1] should contain an error for the password! That's the point of this test.

test/emailpassword/signinFeature.test.js
```diff
@@ -39,3 +56,1315 @@
...
+                        resolve(undefined);
+                    } else {
+                        resolve(res);
+                    }
+                })
+        );
+        let userInfo = await emailpassword.getUserByEmail("random@gmail.com");
+
+        assert(userInfo.email === "random@gmail.com");
+        assert(userInfo.id !== undefined);
```
rishabhpoddar, line 1307: check that the ID is same as what is returned during sign up

test/emailpassword/signinFeature.test.js
```diff
@@ -39,3 +56,1315 @@
...
+                        resolve(undefined);
+                    } else {
+                        resolve(JSON.parse(res.text));
+                    }
+                })
+        );
+        let userInfo = await emailpassword.getUserById(response.user.id);
+
+        assert(userInfo.email === "random@gmail.com");
+        assert(userInfo.id !== undefined);
```
rishabhpoddar, line 1368: check that the ID is same as what is returned during sign up

test/middleware.test.js
```diff
@@ -39,6 +40,32 @@ describe(`middleware: ${printPath("[test/middleware.test.js]")}`, function () {
...
+                Session.init({
+                    sessionRefreshFeature: {
+                        disableDefaultImplementation: true,
+                    },
+                }),
+            ],
+        });
+
+        let apisHandled = await SessionRecipe.getInstanceOrThrowError().getAPIsHandled();
+        assert(apisHandled[0].disabled === true);
```
rishabhpoddar, line 66: No. Create an express app, add the middlewares and error handler, call the refresh API, and check that you get 404

test/querier.test.js
```diff
@@ -39,6 +40,110 @@ describe(`Querier: ${printPath("[test/querier.test.js]")}`, function () {
...
+        });
+        try {
+            await Session.getAllSessionHandlesForUser();
+            assert(false);
+        } catch (err) {
+            if (err.type !== ST.Error.GENERAL_ERROR || err.rId !== "session") {
+                throw err;
+            }
+        }
+    });
```
rishabhpoddar, line 66: Do similar for a function from emailpassword as well please (within this test)

test/querier.test.js
```diff
@@ -39,6 +40,110 @@ describe(`Querier: ${printPath("[test/querier.test.js]")}`, function () {
...
+        await q.getAPIVersion();
+        verifyState = await ProcessState.getInstance().waitForEvent(
+            PROCESS_STATE.CALLING_SERVICE_IN_GET_API_VERSION,
+            2000
+        );
+        assert(verifyState === undefined);
+    });
+
+    // * TODO: Check that rid is added to the header iff it's a "/recipe" || "/recipe/*" request.
+    it("test that rid is added to the header if it's a recipe request", async function () {
```
rishabhpoddar, line 103: What is the failure condition here? How have you tested that?

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
     });
+
+    // * TODO: Check that querier has been inited when we call supertokens.init
+    it("test that querier has been initiated when we call supertokens.init", async function () {
+        await startST();
+
+        try {
+            await Querier.getInstanceOrThrowError();
+            assert(false);
+        } catch (err) {}
```
rishabhpoddar, line 68: You do not check for the error type!

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
+
+    // TODO: Check that modules have been inited when we call supertokens.init
+    it("test that modules have been initiated when we call supertokens.init", async function () {
+        await startST();
+
+        try {
+            await SessionRecipe.getInstanceOrThrowError();
+            await EmailPasswordRecipe.getInstanceOrThrowError();
+            assert(false);
+        } catch (err) {}
```
rishabhpoddar, line 93: - Again, no checking of error type
- `await EmailPasswordRecipe.getInstanceOrThrowError();` never actually runs... cause `await SessionRecipe.getInstanceOrThrowError();` throws an error.. So you should have these two in different try catch

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
+                appName: "SuperTokens",
+                websiteDomain: "supertokens.io",
+            },
+            recipeList: [TestRecipe.init()],
+        });
+        const app = express();
+
+        app.post("/auth/user-api", async (req, res) => {
+            res.status(200).json({ message: "success" });
+        });
```
rishabhpoddar, line 133: Middleware should go before your custom APIs

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
+    });
+
+    /*
+        TODO: Test various inputs to routing (if it accepts or not)
+        - including when the base path is "/"
+        - with and without a rId
+        - where we do not have to handle it and it skips it (with / without rId)
+    */
+
+    it("test various inputs to routing when bass path is /", async function () {
```
rishabhpoddar, line 203: Typo in "bass"

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
+                },
+                recipeList: [TestRecipe.init()],
+            });
+            const app = express();
+
+            app.post("/user-api", async (req, res) => {
+                res.status(200).json({ message: "success" });
+            });
+
+            app.use(ST.middleware());
```
rishabhpoddar, line 224: Middleware to go before your custom API

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
+                .expect(200)
+                .end((err, res) => {
+                    if (err) {
+                        resolve(undefined);
+                    } else {
+                        resolve(res);
+                    }
+                })
+        );
+        assert(r1.text === "success TestRecipe1 /hello");
```
rishabhpoddar, line 334: Also add similar when the rid is the one that belongs to TestRecipe

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
+                        resolve(res);
+                    }
+                })
+        );
+        assert(r1.status === 400);
+        assert(r1.body.message === "Bad input error from TestRecipe");
+    });
+
+    // TODO: Error thrown from APIs implemented by recipes must not go unhandled
+    it("Error thrown from APIs implemented by recipes must not go unhandled", async function () {
```
rishabhpoddar, line 396: You are testing this API with a non-general error.. also call an API that throws a `new Error("...")` and make sure that your error handler (which you need to define) is called with that error. If there is no such API in the TestRecipe, create one.

rishabhpoddar:
# Generic issues:
- [ ] Remove all TODOS in the comments above the tests. It makes it impossible to search for things that are actually left TODO
- [ ] Add failure condition comment on all tests (and make sure that you have tested it). Keep these comments short.
- [ ] Abstract away common blocks of code so that tests are shorter and easier to read. An example is signup API call blacks in the signin tests

test/recipeModuleManager.test.js
```diff
@@ -45,4 +57,674 @@ describe(`recipeModuleManagerTest: ${printPath("[test/recipeModuleManager.test.j
...
+        ST.init({
+            supertokens: {
+                connectionURI: "http://localhost:8080",
+            },
+            appInfo: {
+                apiDomain: "api.supertokens.io",
+                appName: "SuperTokens",
+                websiteDomain: "supertokens.io",
+            },
+            recipeList: [TestRecipe1.init()],
```
rishabhpoddar, line 528: Do this test with multiple recipes. Each recipe should add something unique to the cors list, and make them both add the same value as well (to test that the final result doesn't have duplicates)

test/session.test.js
```diff
@@ -59,6 +59,228 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
...
+                        resolve(undefined);
+                    } else {
+                        resolve(res);
+                    }
+                })
+        );
+        assert(res.header["front-token"] !== undefined);
+        assert(res.header["access-control-expose-headers"] === "front-token, id-refresh-token, anti-csrf");
+        assert(res.header["id-refresh-token"] !== undefined);
+        assert(res.header["anti-csrf"] !== undefined);
```
rishabhpoddar, line 103: You also need to check for cookies.. as the test description says

test/session.test.js
```diff
@@ -59,6 +59,228 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
...
+                    } else {
+                        resolve(res);
+                    }
+                })
+        );
+        assert(res2.header["front-token"] !== undefined);
+        assert(res2.header["access-control-expose-headers"] === "front-token, id-refresh-token, anti-csrf");
+        assert(res2.header["id-refresh-token"] !== undefined);
+        assert(res2.header["anti-csrf"] !== undefined);
+    });
```
rishabhpoddar, line 166: You also need to test for cookies

test/session.test.js
```diff
@@ -59,6 +59,228 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
...
+
+        app.post("/create", async (req, res) => {
+            await Session.createNewSession(res, "", {}, {});
+            res.status(200).send("");
+        });
+
+        app.post("/auth/session/refresh", async (req, res) => {
+            await Session.refreshSession(req, res);
+            res.status(200).send("");
+        });
```
rishabhpoddar, line 132: No need to add the refresh API as the middleware should add it automatically. In fact, this API should not even have been called. The fact that you have this means you did not test for failure condition...

test/session.test.js
```diff
@@ -59,6 +59,228 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
...
+                .post("/auth/session/refresh")
+                .end((err, res) => {
+                    if (err) {
+                        resolve(undefined);
+                    } else {
+                        resolve(res);
+                    }
+                })
+        );
+        assert(res2.status === 500);
```
rishabhpoddar, line 222: This is not an "appropriate" error.. It should be unauthorised error right? status code 401, with a proper message..

test/session.test.js
```diff
@@ -59,6 +59,228 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
...
+
+        const app = express();
+        app.use(SuperTokens.middleware());
+
+        app.post("/create", async (req, res) => {
+            await Session.createNewSession(res, "", {}, {});
+            res.status(200).send("");
+        });
+
+        app.post("/auth/session/refresh", async (req, res) => {
```
rishabhpoddar, line 191: Again, you do not need this API

test/session.test.js
```diff
@@ -59,6 +59,228 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
...
```
rishabhpoddar, line 248: You do not need this API

rishabhpoddar (changes requested):

test/emailpassword/signinFeature.test.js
jscyo: Im testing the signInFeature, the signInFeature throws "FIELD_ERROR" for email syntax error not for password

rishabhpoddar: Okay. So then do this with signUp and check that it returns both the errors, and with sign in, the size of `formFields` is 1

test/emailpassword/signinFeature.test.js
```diff
@@ -610,9 +633,10 @@ describe(`signinFeature: ${printPath("[test/signinFeature.test.js]")}`, function
 
         app.use(STExpress.errorHandler());
 
-        await signUPRequest(app, "random@gmail.com", "valid");
+        let signUpResponse = await signUPRequest(app, "random@gmail.com", "valid");
+        assert(JSON.parse(signUpResponse.text).status === "OK");
```
rishabhpoddar, line 637: Check that validatorCtr is = 0

test/emailpassword/signinFeature.test.js
rishabhpoddar: @jscyo you missed this comment

test/querier.test.js
rishabhpoddar: @jscyo you missed this. Also, this test does not make sense to me. Please explain over a call

rishabhpoddar (changes requested):

test/emailpassword/signinFeature.test.js
jscyo: ive added the check that formFields should be 1, ill be doing the signUp check in the signUpFeature test

rishabhpoddar (approved):

test/emailpassword/signoutFeature.test.js
```diff
@@ -13,10 +13,332 @@
...
+                        resolve(res);
+                    }
+                })
+        );
+        assert(JSON.parse(response.text).status === "OK");
+        assert(response.header["set-cookie"] === undefined);
+    });
+
+    // Call the API without an expired access token session, and it should return TRY_REFRESH_TOKEN
+    it("test that calling the API wiithout the access token session, and it should return TRY_REFRESH_TOKEN", async function () {
```
rishabhpoddar, line 203: Please remove this test as the one below does the same thing and more.

test/emailpassword/signoutFeature.test.js
```diff
@@ -13,10 +13,332 @@
...
+                    } else {
+                        resolve(res);
+                    }
+                })
+        );
+        assert(response2.status === 401);
+        assert(JSON.parse(response2.text).message === "try refresh token");
+    });
+
+    //Call the API with an expired access and refresh token, and it should return "OK" (with clearing the cookies).
```
rishabhpoddar, line 246: Change the description of this test. You have not implemented what the description says.. Change the description to (NOT THE TEST, ONLY THE DESCRIPTION):
"Call the API with an expired access token, refresh, and call the API again to get OK and clear cookies"

Also, `await setKeyValueInConfig("refresh_token_validity", 2);` is useless since the units here is mins.. please remove it

test/emailpassword/signupFeature.test.js
```diff
@@ -45,3 +71,1399 @@
...
+                        resolve(undefined);
+                    } else {
+                        resolve(JSON.parse(res.text));
+                    }
+                })
+        );
+        assert(badInputResponse.message === "formFields must be an array");
+    });
+
+    it("test bad input, formFields is not an array in /signup API", async function () {
```
rishabhpoddar, line 382: this test is duplicated above.. Remove one of them

test/emailpassword/signupFeature.test.js
```diff
@@ -45,3 +71,1399 @@
...
+                            id: "email",
+                            value: "random@gmail.com",
+                        },
+                        {
+                            id: "testField",
+                            value: "",
+                        },
+                    ],
+                })
+                .end((err, res) => {
```
rishabhpoddar, line 938: expect 200 status code

rishabhpoddar (changes requested):

rishabhpoddar:
# General feedback:
- [ ] You have not checked for status code in many of your API calls.

rishabhpoddar (approved):

Issue #56, opened 2020-12-14
NextJS Middleware


NkxxkN:
-we want to expose 

```js
import SuperTokens from "supertokens-node";
import supertokensMiddleware from "supertokens-node/nextjs"
```

This is essentially this function: https://gist.github.com/NkxxkN/fd62a8aa5bf6c4f6c813143de9a8c3ec#file-supertokens-nextjs-demo-pages-api-auth-path-js-L50

TODO:

- [x] Use in `supertokens-nextjs-demo`
- [x] Update guides https://github.com/supertokens/main-website/pull/154
- [x] Edge cases?
- [x] Testing? 
   - Option 1: Implement tests in `supertokens-nextjs-demo` directly => Then run them when deploying `supertokens-node`
          ++ Most realistic test cases
          -- Need to update the version in `supertokens-nextjs-demo` when we bump `supertokens-node`
 - Option 2: Mock the req/res objects from `supertokens-nextjs-demo` for different simple scenarios:
     - Signup
     - SignIn
     - Reset Password
 And verify that result is as expected.
        ++ Longer to implement
        ++ No dependency issues

PR #58, opened 2020-12-14
3.2.nextjs

NkxxkN:
- Add `supertokensMiddleware` |
- Update to changelog
- Bump to 3.2


NkxxkN:
TODO: After this one is released https://github.com/supertokens/supertokens-nextjs-demo/pull/1

NkxxkN:
Fixes https://github.com/supertokens/supertokens-node/issues/56

rishabhpoddar:
- We need to export the next js middleware as a default and a non default export (just like other functions exposed).
- From a file structure point of view, I prefer not having `nextjs.ts` inside a helpers folder. It should be in the root of the `ts` folder imo. The reason is that it's not a helper to anything in the source code. It's a function that the user uses directly.

rishabhpoddar:
We also need to create a new version of the docs for supertokens-node -> 3.2.X. But feel free to commit this please. Looks good to me.

PR #69, opened 2021-01-11
useSuperTokensFromNextJs

NkxxkN:
Expose a general middleware that can be used with any methods:

```js
import {superTokensNextWrapper} from "supertokens-node/nextjs";

// Example for Middleware
superTokensNextWrapper(async (next) => {
  return SuperTokens.middleware(request, response, next)
}, request, response);

// Example for Verify Session
superTokensNextWrapper(async (next) => {
  return SuperTokens.verifySession(request, response, next)
}, request, response);
```

Used in demo repo https://github.com/supertokens/supertokens-nextjs-demo/pull/5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Email verification feature
 - Change the User object to include timeJoined
 - Sends emails to our APIs only if not testing mode
-- Add useSuperTokensFromNextJs generic express middleware wrapper
+- Add superTokensNextWrapper generic express middleware wrapper
 
 ## [3.2.2] - 2020-12-18
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [3.3.0] - 2021-01-09
 ### Added
 - Email verification feature
-- Change to User object to include timeJoined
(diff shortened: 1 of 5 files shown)


NkxxkN:
Rename from `useSuperTokensFromNextJs` to `superTokensNextWrapper`.

The reason is that `eslint(react-hooks/rules-of-hooks)` prevents using any functions starting with `use` (such as `useState`, `useEffect` etc...) from some places. 

`useX` is a reserved prefix by React hooks :/ 

image

rishabhpoddar:
Due to the rename, your first comment will change to
```js
import {superTokensNextWrapper} from "supertokens-node/nextjs";

// Example for Middleware
superTokensNextWrapper(async (next) => {
  return SuperTokens.middleware(request, response, next)
}, request, response);

// Example for Verify Session
superTokensNextWrapper(async (next) => {
  return SuperTokens.verifySession(request, response, next)
}, request, response);
```

Right? If yes, please feel free to merge. Also, we need to make changes to the docs for this.

NkxxkN:
Updated my comment. There will be changes to the docs.

The code snippets in the docs should be replaced with the content of that PR https://github.com/supertokens/supertokens-nextjs-demo/pull/5.
They are hard to maintain though. I think we should encourage to clone the repo, and in case they don't, we should link to the files from the repo so that we do not have to make doc updates every time something changes.

rishabhpoddar:
> docs should be replaced with the content of that PR

I don't think users will like to go through lots of code in a demo app to learn how to use us with NextJS. There will need to be a page in the docs for it for sure. The demo app probably only helps as a reference...

Issue #94, opened 2021-02-24
Accept and pass additional params to third party authorisation url


sbauch:
I'm hoping to add a little bit of functionality to 3rd party oauth providers that allows a developer to get a value from user input and include that as a query param in the service's authorisation url.

NextAuth.js handles this well IMO, and [provides some examples](https://next-auth.js.org/getting-started/client#additional-params) for when this would be useful and further points at an [OIDC spec for more examples](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).

A further example that better illustrates needing user input is Google OAuth's `hd` param, which restricts Google users to using accounts from the domain provided to the `hd` param in the authorization URL.

I maintain an open source OAuth service [Osso](https://github.com/enterprise-oss/osso) that wraps SAML auth in an OAuth flow. SAML authentication is instance based - you must send a user to the expected IDP instance, and so you need to determine who the user is in order to route them to the right IDP.

So we want to be able to craft a login form that asks for the user's email and includes it as a param to the third party authorisation URL.

It seems like this URL construction is happening here - https://github.com/supertokens/supertokens-node/blob/master/lib/ts/recipe/thirdparty/api/authorisationUrl.ts#L68-L70

A naive approach might be to append the other params that are included in the request. I suppose there is some danger in including wayward params however.

@rishabhpoddar suggested on discord that the params type be extended to accept a function. 

I think the suggestion is to expand this type definition to accept functions as values - https://github.com/supertokens/supertokens-node/blob/master/lib/ts/recipe/thirdparty/types.ts#L35

I might suggest keeping that as a `{ [key: string]: string }` typed object, and instead add an additional key `dynamicParams`.

So something like this?

```ts
export type TypeProviderGetResponse = {
    accessTokenAPI: {
        url: string;
        params: { [key: string]: string }; // Will be merged with our object
    };
    authorisationRedirect: {
        url: string;
        params: { [key: string]: string };
        dynamicParams: { [key: string]: (request: Request) => string; } 
    };
    getProfileInfo: (authCodeResponse: any) => Promise<UserInfo>;
};
```

Then, if we were to update the third party strategies to pass these along - 

```ts
let dynamicParamsString = Object.entries(providerInfo.authorisationRedirect.dynamicParams).reduce(
  (dynamicParams, [key, getter]) => ({
    ...dynamicParams,
    [key]: getter(request),
  }
), {}).toString()

[...]

PR #95, opened 2021-02-28
Accept dynamic params via function for authorisation url

sbauch:
per #94 this allows thirdparty providers to include dynamic params when constructing an authorisation url.

The params value type on the thirdparty now accepts a string or a function, where the function is provided the request object in order to read params, locales, etc.

This can be used for example to collect a user's domain, and use that in the `hd` parameter for google oauth.

Also updates a test.


rishabhpoddar (approved):

rishabhpoddar:
Thanks @sbauch . I have added you as a contributor in supertokens-core. I will release this version very soon!

PR #106, opened 2021-03-31
Redwood Graphql Handler

bhumilsarvaiya:
## Related to:
- https://github.com/supertokens/supertokens-core/issues/129

supertokens redwood graphql handler added


lib/ts/redwood.ts
```diff
@@ -0,0 +1,169 @@
...
+                len = chunk.length;
+            } else {
+                len = Buffer.from(chunk).length;
+            }
+
+            this.set("Content-Length", len);
+        }
+
+        // strip irrelevant headers
+        if (204 === this.statusCode || 304 === this.statusCode) {
```
rishabhpoddar, line 85: Why specifically these status codes only?

lib/ts/redwood.ts
```diff
@@ -0,0 +1,169 @@
...
+        }
+        this.end(chunk);
+
+        return this;
+    };
+}
+
+// TODO: take serverless app as argument --DONE
+// TODO: this should be imported from supertokens-node/redwood --DONE
+// TODO: supertokensRedwoodGraphQLHandler -> supertokensGraphQLHandler --DONE
```
rishabhpoddar, line 99: Remove these todo comments

rishabhpoddar (changes requested):

lib/ts/redwood.ts
bhumilsarvaiya: Took this part from express response. Should I remove this?

lib/ts/redwood.ts
bhumilsarvaiya: okay

lib/ts/redwood.ts
rishabhpoddar: I don't know. See if it makes sense.. understand the code.. don't just blindly copy paste please.

bhumilsarvaiya: Response code 304 (Not Modified) indicates that the requested resource has not been modified since the previous transmission.
Response code 204 (No Content) success status indicates that a request has succeeded, but that the client doesn't need to navigate away from its current page.
For both these response codes, response is expected to have no content. So the above change just makes sure that any content related headers are stripped off.

rishabhpoddar (approved):

Issue #103, opened 2021-03-23
Give verifySession a flag that allows the API to be called even if there is no user


rishabhpoddar:
- In this case, `req.session` will be `undefined`.
- Something like `verifySession(true)`

TODO:
- [x] Implementaion
- [x] Issue: If headers are set for unauthorised call, then we need to make sure those are sent to the frontend as well (or that those are handled - @bhumilsarvaiya )
- [x] Test
- [x] Docs

bhumilsarvaiya:
Currently, verifySession accepts a boolean parameter `antiCsrfCheck` which is optional. i.e.

```ts
type requestHandler = (request: SessionRequest, response: Response, next: NextFunction) => Promise<void>;
verifySession(antiCsrfCheck?: boolean) => requestHandler;
```

If we add another optional parameter that allows the API to be called even if there is no session, let's call it `sessionRequired`. So the signature will become:

```ts
type requestHandler = (request: SessionRequest, response: Response, next: NextFunction) => Promise<void>;
verifySession(antiCsrfCheck?: boolean, sessionRequired?: boolean) => requestHandler;
```

Default value of `sessionRequired` will be `true`.

One drawback will be, for developers who just want to pass this a boolean that allows the API to be called even if there is no session, now also have to pass either `undefined` or `boolean` for `antiCsrfCheck` compulsorily. One way to counter this is to accept an object where user can pass possible flags. The verifySession function signature will then become like:

```ts
type verifySessionOptions?: {
    antiCsrfCheck?: boolean,
    sessionRequired?: boolean
};
type requestHandler = (request: SessionRequest, response: Response, next: NextFunction) => Promise<void>;
verifySession(options?: verifySessionOptions | boolean) => requestHandler;
```

Here, `options?: verifySessionOptions | boolean` is to make the changes backward compatible. In the documentation of the new `verifySession` function, we should only mention about `verifySessionOptions`.

rishabhpoddar:
I prefer the method in which the user has to provide that object with the two keys.

rishabhpoddar:
If `sessionRequired` is `false`, and no session is provided, I assume `req.session` will be `undefined`?

bhumilsarvaiya:
Yes, that's correct

rishabhpoddar:
@bhumilsarvaiya how would this interface be affected when we add anonymous sessions? We would ideally not want to do a breaking change by renaming this `sessionRequired` variable...

[...]

PR #108, opened 2021-04-02
sessionRequired flag in verifySession

bhumilsarvaiya:
related to issue #103


rishabhpoddar (approved):

lib/ts/recipe/emailverification/api/emailVerify.ts
```diff
@@ -67,7 +68,7 @@ export default async function emailVerify(recipeInstance: Recipe, req: Request,
                 }
             })
         );
-        let session = (req as SessionRequest).session;
+        let session = (req as SessionRequest).session as SessionClass;
```
rishabhpoddar, line 71: The point of doing as `SessionRequest` is that we don't have to add another as `SessionClass`. So please check your types...
If you are getting type of Session has `undefined | SessionClass`, then have an `undefined` check below and throw a GENERAL_ERROR in that case, with a message like "should never come here" or something

lib/ts/recipe/emailverification/api/generateEmailVerifyToken.ts
```diff
@@ -37,7 +38,7 @@ export default async function generateEmailVerifyToken(
             }
         })
     );
-    let session = (req as SessionRequest).session;
+    let session = (req as SessionRequest).session as SessionClass;
```
rishabhpoddar, line 41: See comment in emailVerify API

lib/ts/recipe/session/middleware.ts
```diff
@@ -13,31 +13,54 @@
...
         } catch (err) {
+            /**
+             * checking:
+             *  - it should not be the refresh API
+             *  - error thrown should be either unauthorised or try refresh token
+             *  - sessionRequired parameter is set to false in options
+             */
+            if (
+                !isRefreshAPICall(recipeInstance, request) &&
+                (err.type === STError.UNAUTHORISED || err.type === STError.TRY_REFRESH_TOKEN) &&
```
rishabhpoddar, line 57: This should only be `(err.type === STError.UNAUTHORISED)` instead of `(err.type === STError.UNAUTHORISED || err.type === STError.TRY_REFRESH_TOKEN)`. Cause if it throws try refresh token, then a session does exist, and we want to refresh that.. not give undefined as `req.session`. 

There should be a test for this case. That an expired access token is used when sessionRequired is false. That should yield a try refresh token 401 error.

test/middleware.test.js
```diff
@@ -126,10 +126,30 @@ describe(`middleware: ${printPath("[test/middleware.test.js]")}`, function () {
             res.status(200).json({ message: req.session.getUserId() });
         });
 
-        app.get("/user/handle", Session.verifySession(true), async (req, res) => {
+        app.get("/user/handleV0", Session.verifySession(true), async (req, res) => {
```
rishabhpoddar, line 129: There are way too many test changes, and they seem to follow a similar pattern. You will need to guide me through them.

rishabhpoddar (changes requested):

lib/ts/recipe/session/middleware.ts
rishabhpoddar, line 57: If headers are set for unauthorised call, then we need to make sure those are sent to the frontend as well (or that those are handled

rishabhpoddar (changes requested):

rishabhpoddar:
@bhumilsarvaiya we can ship this in version 4.3 since the unauthorised error part is unclear.

lib/ts/recipe/session/middleware.ts
rishabhpoddar: Removed `TRY_REFRESH_TOKEN` part, but need to add a test

rishabhpoddar (approved):

rishabhpoddar (approved):

lib/ts/recipe/session/middleware.ts
rishabhpoddar: This issue is addressed by letting optional session pass through only when there are no session tokens in the headers int he first place - do you think this makes sense?

rishabhpoddar:
# TESTS TO ADD (in all `sessionRequired` should be `false`:
- [ ] An expired access token is used and we see that try refresh token error is thrown
- [ ] A session exists, is refreshed, then is revoked, and then we try and use the access token (after first refresh), and we see that unauthorised error is called.
- [ ] A session doesn't exist, and we call verifySession, and it let's go through

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [4.3.0] - Unreleased
+### Changed
+- verifySession now accepts an object as a parameter. Two valid parameters are sessionRequired and antiCsrfCheck. sessionRequired can be by user if they intends to make the session requirement optional for the API.
```
rishabhpoddar, line 9: There is a typo in this sentence..

rishabhpoddar (changes requested):

CHANGELOG.md
bhumilsarvaiya: fixed

rishabhpoddar (approved):

PR #135, opened 2021-06-12
Fix email sending in serverless env

aldeed:
## Issue

In a serverless environment (specifically Vercel functions), the function program seems to stop shortly after a response is sent. This means that `createAndSendCustomEmail` does not always finish even though it is awaited. So usually an email is never sent even though the API responds "ok".

## Reproduction

1. Define a custom async `createAndSendCustomEmail` for either password reset or email verify. In it, wait some (5?) seconds before resolving, and log something after waiting.
2. Deploy to Vercel
3. Open the realtime functions log in Vercel dashboard
4. Hit the API route to verify email or reset password. Note that the API response is "ok" but note in Vercel logs that your final log message after waiting is never printed.

## Solution

I am not sure if this is a Vercel issue, lambda issue, or something else. Supposedly lambda functions do wait for the promise to resolve, but Vercel may be overriding this somehow. The solution in this PR does fix it for me and it seems fine to do this, but it may be considered more of a workaround than a fix.

The change is to check whether `isInServerlessEnv`, and if so, do not send the response until after awaiting `createAndSendCustomEmail`.

### DIFF
--- a/lib/build/recipe/emailpassword/api/generatePasswordResetToken.js
+++ b/lib/build/recipe/emailpassword/api/generatePasswordResetToken.js
@@ -85,10 +85,13 @@ function generatePasswordResetToken(recipeInstance, req, res, _) {
             token +
             "&rid=" +
             recipeInstance.getRecipeId();
-        // step 5
-        utils_1.send200Response(res, {
-            status: "OK",
-        });
+        // step 5 - respond before email sending if not in a serverless environment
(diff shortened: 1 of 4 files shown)


rishabhpoddar:
Thank you for this PR @aldeed. I have added your profile to supertokens-core readme :)

rishabhpoddar:
I have released this in v5.1.0

PR #152, opened 2021-07-17
Add ability to fetch all session info from database using a session handle

nkshah2:
### Summary
- Adds a new method to fetch all session info stored in the database for a given session handle
- Marks existing methods to fetch JWT and session data individually by session handle as deprecated

### Related issues
https://github.com/supertokens/supertokens-core/issues/255

### DIFF
--- a/lib/ts/recipe/session/sessionClass.ts
+++ b/lib/ts/recipe/session/sessionClass.ts
@@ -19,9 +19,6 @@ import STError from "./error";
 import NormalisedURLPath from "../../normalisedURLPath";
 import { SessionContainerInterface } from "./types";
 import RecipeImplementation from "./recipeImplementation";
-import { Querier } from "../../querier";
-import SuperTokens from "../../supertokens";
-import { maxVersion } from "../../utils";
 
 export default class Session implements SessionContainerInterface {
     private sessionHandle: string;
@@ -55,17 +52,7 @@ export default class Session implements SessionContainerInterface {
 
     getSessionData = async (): Promise<any> => {
         try {
-            let querier = Querier.getNewInstanceOrThrowError(SuperTokens.getInstanceOrThrowError().isInServerlessEnv);
-            let apiVersion = await querier.getAPIVersion();
-
-            // Call older function for < 2.8
-            if (maxVersion(apiVersion, "2.7") === "2.7") {
-                return await SessionFunctions.getSessionData(this.recipeImplementation, this.sessionHandle);
-            } else {
-                return await (
-                    await SessionFunctions.getSessionInformation(this.recipeImplementation, this.sessionHandle)
-                ).userDataInDatabase;
-            }
+            return await SessionFunctions.getSessionData(this.recipeImplementation, this.sessionHandle);
         } catch (err) {
             if (err.type === STError.UNAUTHORISED) {
                 clearSessionFromCookie(this.recipeImplementation.config, this.res);
@@ -137,13 +124,6 @@ export default class Session implements SessionContainerInterface {
 
     getTimeCreated = async (): Promise<number> => {
         try {
-            let querier = Querier.getNewInstanceOrThrowError(SuperTokens.getInstanceOrThrowError().isInServerlessEnv);
-            let apiVersion = await querier.getAPIVersion();
-
-            if (maxVersion(apiVersion, "2.7") === "2.7") {
-                throw new Error("Please use core version >= 3.5 to call this function.");
-            }
-
             return (await SessionFunctions.getSessionInformation(this.recipeImplementation, this.sessionHandle))
                 .timeCreated;
         } catch (err) {
@@ -156,13 +136,6 @@ export default class Session implements SessionContainerInterface {
 
     getExpiry = async (): Promise<number> => {
         try {
-            let querier = Querier.getNewInstanceOrThrowError(SuperTokens.getInstanceOrThrowError().isInServerlessEnv);
-            let apiVersion = await querier.getAPIVersion();
-
-            if (maxVersion(apiVersion, "2.7") === "2.7") {
-                throw new Error("Please use core version >= 3.5 to call this function.");
-            }
-
             return (await SessionFunctions.getSessionInformation(this.recipeImplementation, this.sessionHandle)).expiry;
         } catch (err) {
             if (err.type === STError.UNAUTHORISED) {
--- a/lib/ts/recipe/session/sessionClass.ts
+++ b/lib/ts/recipe/session/sessionClass.ts
@@ -19,6 +19,9 @@ import STError from "./error";
 import NormalisedURLPath from "../../normalisedURLPath";
 import { SessionContainerInterface } from "./types";
 import RecipeImplementation from "./recipeImplementation";
+import { Querier } from "../../querier";
+import SuperTokens from "../../supertokens";
+import { maxVersion } from "../../utils";
 
 export default class Session implements SessionContainerInterface {
     private sessionHandle: string;
@@ -52,7 +55,17 @@ export default class Session implements SessionContainerInterface {
 
     getSessionData = async (): Promise<any> => {
         try {
-            return await SessionFunctions.getSessionData(this.recipeImplementation, this.sessionHandle);
+            let querier = Querier.getNewInstanceOrThrowError(SuperTokens.getInstanceOrThrowError().isInServerlessEnv);
+            let apiVersion = await querier.getAPIVersion();
+
+            // Call older function for < 2.8
+            if (maxVersion(apiVersion, "2.7") === "2.7") {
+                return await SessionFunctions.getSessionData(this.recipeImplementation, this.sessionHandle);
+            } else {
+                return await (
+                    await SessionFunctions.getSessionInformation(this.recipeImplementation, this.sessionHandle)
+                ).userDataInDatabase;
+            }
         } catch (err) {
             if (err.type === STError.UNAUTHORISED) {
                 clearSessionFromCookie(this.recipeImplementation.config, this.res);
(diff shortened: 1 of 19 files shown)


lib/ts/recipe/session/types.ts
```diff
@@ -225,3 +234,12 @@ export interface APIInterface {
         options: APIOptions;
     }): Promise<void>;
 }
+
+export type SessionInformation = {
+    sessionHandle: string;
+    userId: string;
+    userDataInDatabase: Object;
```
rishabhpoddar, line 241: Use `any` instead of `object`

lib/ts/recipe/session/types.ts
```diff
@@ -225,3 +234,12 @@ export interface APIInterface {
...
+    expiry: number;
+    userDataInJWT: Object;
```
rishabhpoddar, line 243: Use `any` instead of `object`

rishabhpoddar:
- Make modifications to changelog.md file
- Do you need to make the `getSessionData` and `getJWTPayload` functions in the `index.ts` file as deprectaed? You need to make sure that when a user is calling these functions, it shows a strikethrough to them. You can test this by editing the `/test/with-typescript/index.ts` file.
- Should we deprecate the `getSession` function in `SessionClass.ts` as well? And add a `getSessionInformation` function instead? If yes, we don't want to deprecate the `getJWTPayload` function as that doesn't to any db query so it is useful. But then the API to the user becomes weird, in that calling `getJWTPayload` returns the payload without a db query, and calling `getSessionInformation` returns the same payload (and more), but with a db query.. Any ideas on how we can streamline this interface to be better?

lib/ts/recipe/session/types.ts
nkshah2: Done

lib/ts/recipe/session/types.ts
nkshah2: Done

nkshah2:
> Make modifications to changelog.md file

Done

---

> Do you need to make the getSessionData and getJWTPayload functions in the index.ts file as deprectaed? You need to make sure that when a user is calling these functions, it shows a strikethrough to them. You can test this by editing the /test/with-typescript/index.ts file.

Marking the interface declarations as deprecated should suffice I think since the index.ts class simply inherits from it. Taking a look at the code in the tests directory, the method usage warns for deprecation without needing any additional changes.
Screenshot 2021-07-18 at 11 20 38 PM

---

> Should we deprecate the getSession function in SessionClass.ts as well? And add a getSessionInformation function instead? If yes, we don't want to deprecate the getJWTPayload function as that doesn't to any db query so it is useful. But then the API to the user becomes weird, in that calling getJWTPayload returns the payload without a db query, and calling getSessionInformation returns the same payload (and more), but with a db query.. Any ideas on how we can streamline this interface to be better?

The class in `sessionClass.ts` is being used as an object to hold session information and is initialised in specific cases, technically an instance of that class would be used to return session information from elsewhere. In this PR adding a type for returning from the `getSessionInformation` function was enough, I dont think modifying that class is required.

---

> Also write down the list of changes needed for the docs for this PR (as a checklist in this PR).

On it

nkshah2:
- [ ] Adding a deprecation notice for the [getJWTPayload](https://supertokens.io/docs/nodejs/session/getjwtpayload) function 
- [ ] Adding a deprecation notice for the [getSessionData](https://supertokens.io/docs/nodejs/session/getsessiondata) function
- [ ] Adding a documentation page for the new `getSessionInformation` function and adding a link to the deprecation warnings of the other functions which redirects to this one. Possible also add text explaining that this function replaces the other two (So users are aware of the change no matter the order they visit the function pages in)

rishabhpoddar (approved):

rishabhpoddar:
> In this PR adding a type for returning from the getSessionInformation function was enough, I dont think modifying that class is required.

I think we should.. cause we can also get the expiration time etc.. now. So we should either add those extra functions or add `session.getInformation()` (which I don't like..). Maybe we should add `session.getExpiry()` and other functions per field returned by the new API.

>  (docs changes): Deprecation notices for the same functions under the SessionContainer section (getJWTPayload, getSessionData) &  Adding a documentation page for getSessionInformation under the SessionContainer section

We don't need this right? Cause you said we are not going to change anything in the session container object.

nkshah2:
> We don't need this right? Cause you said we are not going to change anything in the session container object.

You're right, this wont be required

---

> I think we should.. cause we can also get the expiration time etc.. now. So we should either add those extra functions or add session.getInformation() (which I don't like..). Maybe we should add session.getExpiry() and other functions per field returned by the new API.

So this would require making changes to every place that initialises this class. I can add expiry, user id etc to that class and make them optional in the constructor so the exiting usages of that class remain unaffected but I wasn't sure if that was a good idea. If you think it is ill make the changes

rishabhpoddar:
> So this would require making changes to every place that initialises this class. I can add expiry, user id etc to that class and make them optional in the constructor so the exiting usages of that class remain unaffected but I wasn't sure if that was a good idea. If you think it is ill make the changes

So you don't need to change the constructor.. you can just create new functions in the class that query the core to get this info. I would also like you to "rethink" parts of this.. maybe we are not coming up with a good enough interface. Please think about how best to expose this new functionality and how and what to deprecate.. 

If after you have thought about it, you still think what we have and propose is the best, go for it. Whilst thinking about API, imagine how and where it would be communicated in the docs, and make the API that leads to the simplest doc explanation.

nkshah2:
Just to clarify my understanding, from the user's point of view the primary class/structure that you interact with for sessions is SessionWrapper correct?

If yes, then I think the following changes would be required:
- Marking methods in SessionWrapper as deprecated as well, marking them in the interface will help to give warnings within the codebase of the package but `SessionWrapper` deprecation will be required to give warnings to the end user. (Not sure if marking the exports as deprecated works, if it does then i will do that as well)
- Add methods in sessionClass (via the `SessionContainerInterface`) for `expiry` and `timeCreated`, these will query the database using the SessionFunctions class
- Change the `getSessionData` in sessionClass to use getSessionInformation (This can be left as is but i think we should use the new method since we are already making a DB call)

rishabhpoddar:
> Just to clarify my understanding, from the user's point of view the primary class/structure that you interact with for sessions is SessionWrapper correct?

There are two ways to interact with sessions:
1) Via direct import statements. Here the user can use the `static` functions from the `SessionWrapper` class. These functions work independently to a request. i.e., even if a request doesn't have any session tokens, or it's a backend cronjob, you can use these functions to manipulate sessions.
2) Via the session object. This object is an instance of `sessionClass.ts`. It's returned as a result of session verification or refreshing of a session. As a result, it's tied to that request / response and can manipulate cookies in them.

As a result, there are operaitons that can be done via both the methods, but the results are different. For example: 
- `revokeSession` via the (1) way would not clear the cookies on the frontend for that user (since it's not tied to a request / response). Whereas, the same function in the `sessionClass.ts` would clear the cookies as well.
- getting JWT payload from (1) would require a call to the core (since there is no access token to refer to, since there is no request / response). However, doing the same thing via `sessionClass.ts` would not call the core because it has access to the request object, which contains the access token, which can be used to get the JWT payload.

nkshah2:
Ok that makes sense, with that in mind I think the 3 changes I mentioned above should suffice (in addition to the changes already in the PR that is). Does that make sense?

rishabhpoddar:
Yes. Those changes make sense. However:

> Change the getSessionData in sessionClass to use getSessionInformation (This can be left as is but i think we should use the new method since we are already making a DB call)

This should continue to use the older API since if we make it use the new API, it will be a breaking change (cause then users of this SDK will need to upgrade their core so that this function can still work).

> Add methods in sessionClass (via the SessionContainerInterface) for expiry and timeCreated, these will query the database using the SessionFunctions class

These functions should check if the current CDI version being used (via the api version) matches what is required.. else it should throw an appropriate error.

---------
Also, don't make docs changes at the moment as we are migrating to v2. Post that, you can make changes to that..

nkshah2:
Makes sense, should the getSessionData function be marked as deprecated?

rishabhpoddar:
We are not adding a `getSessionInformation` function to that class right? If we aren't, then it should not be marked as deprecated.

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -52,7 +55,17 @@ export default class Session implements SessionContainerInterface {
 
     getSessionData = async (): Promise<any> => {
         try {
-            return await SessionFunctions.getSessionData(this.recipeImplementation, this.sessionHandle);
+            let querier = Querier.getNewInstanceOrThrowError(SuperTokens.getInstanceOrThrowError().isInServerlessEnv);
```
rishabhpoddar, line 58: You should change `SessionFunctions.getSessionData` function to do this logic that you have made, and continue to use `SessionFunctions.getSessionData` inside this function here.

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -121,4 +134,41 @@ export default class Session implements SessionContainerInterface {
...
+    getTimeCreated = async (): Promise<number> => {
+        try {
+            let querier = Querier.getNewInstanceOrThrowError(SuperTokens.getInstanceOrThrowError().isInServerlessEnv);
+            let apiVersion = await querier.getAPIVersion();
+
+            if (maxVersion(apiVersion, "2.7") === "2.7") {
+                throw new Error("Please use core version >= 3.5 to call this function.");
+            }
+
+            return (await SessionFunctions.getSessionInformation(this.recipeImplementation, this.sessionHandle))
```
rishabhpoddar, line 147: Checking of `apiVersion` (or throwing error) should happen inside `SessionFunctions.getSessionInformation`.. so that this code is not repeated whenever `SessionFunctions.getSessionInformation` is used.

rishabhpoddar (changes requested):

rishabhpoddar:
@nkshah2 please also update the checklist in one of the comments above.

lib/ts/recipe/session/sessionClass.ts
nkshah2: Done

lib/ts/recipe/session/sessionClass.ts
nkshah2: Done

nkshah2:
Made the changes and edited the documentation checklist

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -244,6 +247,13 @@ export async function getSessionInformation(
     recipeImplementation: RecipeImplementation,
     sessionHandle: string
 ): Promise<SessionInformation> {
+    let querier = Querier.getNewInstanceOrThrowError(SuperTokens.getInstanceOrThrowError().isInServerlessEnv);
```
rishabhpoddar, line 250: Please see how other functions in this file gets the `querier` -> `await recipeImplementation.querier`

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -381,7 +391,15 @@ export async function getSessionData(recipeImplementation: RecipeImplementation,
         sessionHandle,
     });
     if (response.status === "OK") {
-        return response.userDataInDatabase;
```
rishabhpoddar, line 361: - You should first check for version, and then check call the getSession or getSessionInformation API.. 
- Please see how other functions in this file gets the `querier` -> `await recipeImplementation.querier`

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionFunctions.ts
nkshah2: Done

lib/ts/recipe/session/sessionFunctions.ts
nkshah2: Makes sense, done

rishabhpoddar (approved):

rishabhpoddar:
@nkshah2 this looks good now. Feel free to merge.

PR #155, opened 2021-07-21
change in oauth scopes

### LINKED ISSUE #148: Change in OAuth scopes
- Need to recheck if we are actually asking for the minimum scope or not
- If the user providers their own scope array, should we not use ours at all? In case they don't ask for the minimum scope, then we through an error since we can't get the profile info.
- This is a breaking change.

bhumilsarvaiya:
closes #148

## TODO:
- [x] Merge to `7.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     -   `ThirdParty.getUserCount()`, `ThirdParty.getUsersNewestFirst()`, `ThirdParty.getUsersOldestFirst`
     -   `EmailPassword.getUserCount()`, `EmailPassword.getUsersNewestFirst()`, `EmailPassword.getUsersOldestFirst`
     -   `ThirdPartyEmailPassword.getUserCount()`, `ThirdPartyEmailPassword.getUsersNewestFirst()`, `ThirdPartyEmailPassword.getUsersOldestFirst`
+-   Deprecates (instead use `Session.getSessionInformation()`)
+    -   `Session.getSessionData()`, `Session.getJWTPayload()`
 
 ## [6.0.3] - 2021-07-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.2.0] - 2021-07-21
+
+## Changes
+
+-   Updated minimum Oauth scopes required for various thirdparty providers.
+-   For thirdparty providers, if the scope array is passed by the user, the default scopes will not be used.
+
 ## [6.1.0] - 2021-06-24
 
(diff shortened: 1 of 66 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.2.0] - 2021-07-21
+
+## Changes
+
+-   Updated minimum Oauth scopes required for various thirdparty providers.
+-   For thirdparty providers, if the scope array is passed by the user, the default scopes will not be used.
```
rishabhpoddar, line 13: These are breaking changes.. So please change the sub title to `## Breaking Changes` and this should be released in version `7.0.0` then?

lib/ts/recipe/thirdparty/providers/apple.ts
```diff
@@ -132,11 +132,14 @@ export default function Apple(config: TypeThirdPartyProviderAppleConfig): TypePr
...
-        let scopes = ["name", "email"];
+        let scopes = ["email"];
         if (config.scope !== undefined) {
-            scopes.push(...config.scope);
+            scopes = config.scope;
             scopes = Array.from(new Set(scopes));
         }
+        if (!scopes.includes("email")) {
+            throw Error("Scope email is required. Please add it to the scope array");
+        }
```
rishabhpoddar, line 142: Why are you adding this check if you are allowing the user to pass their own `scope`? If you are doing this, you might as well add this to their scope too.. I think you should remove this `if` statement completely (applies for all providers).

lib/ts/recipe/thirdparty/providers/apple.ts
```diff
@@ -132,11 +132,14 @@ export default function Apple(config: TypeThirdPartyProviderAppleConfig): TypePr
             accessTokenAPIParams.redirect_uri = redirectURI;
         }
         let authorisationRedirectURL = "https://appleid.apple.com/auth/authorize";
-        let scopes = ["name", "email"];
```
rishabhpoddar, line 135: I think we should keep `name`? Let's discuss..

lib/ts/recipe/thirdparty/providers/google.ts
```diff
@@ -80,14 +80,16 @@ export default function Google(config: TypeThirdPartyProviderGoogleConfig): Type
             accessTokenAPIParams.redirect_uri = redirectURI;
         }
         let authorisationRedirectURL = "https://accounts.google.com/o/oauth2/v2/auth";
-        let scopes = [
-            "https://www.googleapis.com/auth/userinfo.profile",
-            "https://www.googleapis.com/auth/userinfo.email",
-        ];
+        let scopes = ["https://www.googleapis.com/auth/userinfo.email"];
```
rishabhpoddar, line 83: Maybe we should keep userinfo.profile? Let's discuss..

rishabhpoddar (changes requested):

rishabhpoddar:
@bhumilsarvaiya please feel free to merge this into 7.0, and you can move the corresponding issue into the documentation column on zenhub.

PR #159, opened 2021-08-01
fixes #156

### LINKED ISSUE #156: Issue when frontend is using `sendBeacon` and session verification is optional
## Issue 1:
If a session exists, but the access token has expired, using `sendBeacon` will only cause the `sIdRefreshToken` to be sent to the backend (no access token, no `rid` header, no `anti-csrf` token).

Even if the session is marked as optional, it will still throw a try refresh token exception since `sIdRefreshToken` exists. However, the frontend will not be able to refresh the session since there are no interceptors added to `sendBeacon`.

## Issue 2:
If the access token is there too, then the logic will try and do an anti-csrf check. This will fail, since the `rid` header is missing. The user can solve this by setting `antiCsrf: false`.

This is OK, but the user will need to set `option: true` and  `antiCsrf: false` for it to work fully. Is this a good API?

bhumilsarvaiya:
## Summary of change
- updated `getSession` function to throw TRY_REFRESH_TOKEN error only if session required is true in verification options or rid is set in headers (idRefreshToken is passed but accessToken is not passed).

## Related issues
- #156

## Test Plan
added a test in session.test.js

- [x] Changelog has been updated
- [x] Changes to the version if needed
   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`
- [x] Had run `npm run build-pretty`
- [x] Had installed and ran the pre-commit hook
- [x] Issue this PR against the latest non released version branch.


rishabhpoddar:
Have you considered what happens when we use getSession in server side rendering? In there, we won't have the rid in the header either.

rishabhpoddar:
@bhumilsarvaiya do you think this is a breaking change? if yes, how we can make it not a breaking change? If not, then you can make the target branch `6.0`?

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.1.0] - 2021-07-31
+
+## Fixes
+
+-   https://github.com/supertokens/supertokens-node/issues/156
```
rishabhpoddar, line 12: Needs to go in 7.0.0

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #174, opened 2021-09-09
Added support for multiple access token signing keys

porcellus:
## Summary of change

Added support for multiple access token signing keys. See supertokens/supertokens-core#307

## Related issues

-   supertokens/supertokens-core#305

## Test Plan

Only manual tested yet, TODO

## Documentation changes

TODO

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Documentation
-   [ ] Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Support for multiple access token signing keys: https://github.com/supertokens/supertokens-core/issues/305
 -   Supporting CDI 2.9
 
+### Removed
+
+-   Unnecessary serverless optimisation
+
 ## [7.0.1] - 2021-09-10
 
 ### Breaking Change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.1.0]
+
+### Added
+
+-   Support for multiple access token signing keys: https://github.com/supertokens/supertokens-core/issues/305
+-   Supporting CDI 2.9
+
 ## [7.0.1] - 2021-09-10
 
 ### Breaking Change
@@ -23,7 +30,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   BaseRequest and BaseResponse interface added which will be used inside recipe instead of previously used express.Request and express.Response
 -   `framework` config option. Default value is `express`.
 -   basic tests for all frameworks.
--   Support for multiple access token signing keys
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.0.1] - 2021-09-10
+
+### Breaking Change
+
+-   Removed redwood support
+
+### Fixes
+
+-   fixes issue #175 regarding making error handlers async, session error callabacks async and making VerifySessionOptions optional for verifySession middleware.
+
 ## [7.0.0] - 2021-07-31
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Unnecessary serverless optimisation
 
+# [7.0.2] - 2021-09-16
+
+### Fixes
+
+-   fixes issue #180 where user was required to have all frameworks as depedencies or else use `compilerOptions.skipLibCheck` in their tsconfig.json
+-   `verifySession` middleware ts issue fix for Hapi framework
+-   ts import for `verifySession` fix for koa
+
 ## [7.0.1] - 2021-09-10
 
 ### Breaking Change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   BaseRequest and BaseResponse interface added which will be used inside recipe instead of previously used express.Request and express.Response
 -   `framework` config option. Default value is `express`.
 -   basic tests for all frameworks.
+-   Support for multiple access token signing keys
 
 ### Changed
 
--- a/lib/ts/recipe/session/recipeImplementation.ts
+++ b/lib/ts/recipe/session/recipeImplementation.ts
@@ -276,9 +276,7 @@ export default class RecipeImplementation implements RecipeInterface {
             let response = await this.querier.sendPostRequest(new NormalisedURLPath("/recipe/handshake"), {});
 
             this.handshakeInfo = {
-                signingKeyLastUpdated: this.handshakeInfo !== undefined ? this.handshakeInfo.signingKeyLastUpdated : 0,
-                jwtSigningPublicKeyList:
-                    this.handshakeInfo !== undefined ? this.handshakeInfo.jwtSigningPublicKeyList : [],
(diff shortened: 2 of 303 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   BaseRequest and BaseResponse interface added which will be used inside recipe instead of previously used express.Request and express.Response
 -   `framework` config option. Default value is `express`.
 -   basic tests for all frameworks.
+-   Support for multiple access token signing keys
```
rishabhpoddar, line 26: Incorrect placement of this. Should be in 7.1.0. Likewise, change base branch to 7.1 for this PR

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -267,41 +274,36 @@ export default class RecipeImplementation implements RecipeInterface {
...
             if (this.isInServerlessEnv) {
                 storeIntoTempFolderForServerlessCache(SERVERLESS_CACHE_HANDSHAKE_INFO_FILE_PATH, this.handshakeInfo);
             }
         }
         return this.handshakeInfo;
     };
 
-    updateJwtSigningPublicKeyInfo = (newKey: string, newExpiry: number) => {
+    updateJwtSigningPublicKeyInfo = (keyList: KeyInfo[] | undefined, publicKey: string, expiryTime: number) => {
+        if (keyList === undefined) {
```
rishabhpoddar, line 300: or the length of it is 0? Cause when will it ever be `undefined`?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -267,41 +274,36 @@ export default class RecipeImplementation implements RecipeInterface {
...
+
+            this.updateJwtSigningPublicKeyInfo(
+                response.jwtSigningPublicKeyList,
+                response.jwtSigningPublicKey,
+                response.jwtSigningPublicKeyExpiryTime
+            );
+
             if (this.isInServerlessEnv) {
                 storeIntoTempFolderForServerlessCache(SERVERLESS_CACHE_HANDSHAKE_INFO_FILE_PATH, this.handshakeInfo);
             }
```
rishabhpoddar, line 294: This reminds me that the cached data for serverless can be stored in the older format (without `jwtSigningPublicKeyList`.. need to discuss the implications of this.

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -83,64 +85,85 @@ export async function getSession(
...
-             * if access token creation time is before the signing key was last
-             * updated, we just return TRY_REFRESH_TOKEN to the client
-             */
-            let payload;
+
+    // If we have no key old enough to verify this access token we should reject it without calling the core
+    let foundOlderKey = false;
+    for (const key of handShakeInfo.jwtSigningPublicKeyList) {
+        // If the key is already expired we shouldn't use it for verification
+        if (key.expiryTime > Date.now()) {
```
rishabhpoddar, line 93: - We would expect the core to give only those tokens have not expired?
- Also, this check is not done in the core [here](https://github.com/supertokens/supertokens-core/pull/307/files#diff-5621d130c254985cb161a1060a401fcc48c5aadaf882e0cad1d52746eecb9d48R56). Is that ok in the core?

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -83,64 +85,85 @@ export async function getSession(
...
+                // of the oldest key instead of the last update time
+                if (timeCreated >= key.createdAt) {
+                    foundOlderKey = true;
+                    break;
+                }
             }
         }
     }
+
+    if (accessTokenInfo === undefined && !foundOlderKey) {
```
rishabhpoddar, line 160: Why do we need both the checks? Just `if (!foundOlderKey)` should be enough?

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -83,64 +85,85 @@ export async function getSession(
...
+                if (timeCreated === undefined) {
+                    throw err;
+                }
+
+                // This is kind of like the old signingKeyLastUpdated logic using the creation time
+                // of the oldest key instead of the last update time
+                if (timeCreated >= key.createdAt) {
+                    foundOlderKey = true;
+                    break;
+                }
```
rishabhpoddar, line 155: This would come in this if statement if the access token is a new token, but its signing key is not yet known to the core? In this case, we want to call the core for session verificaiton and we expect an OK reply along with the new signing key array (which should contain the new key?)

Would it be possible for this if statement to be true whilst we might have the actual key (used to sign the input access token) in the `jwtSigningPublicKeyList` array? Why yes / no?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
porcellus: Before 2.9, it's always undefined. With 2.9, it always contains at least one key.

lib/ts/recipe/session/sessionFunctions.ts
porcellus: It's not necessary; I'll remove it.

lib/ts/recipe/session/sessionFunctions.ts
porcellus: We are going from newest signing key to oldest, and the token should only ever be signed with the first key older than the token itself. So the only key we need to check is the first one older than the token itself. 
Selecting this key could be done without the loop, but in most cases, we only have one key anyway and we'd need to parse the token to do it before verification.

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: Alright. Please add a comment for this.

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: TODO: Bhumil is testing if we can remove the serverless stuff entirely...

lib/ts/recipe/session/sessionFunctions.ts
rishabhpoddar: Never access `handShakeInfo.jwtSigningPublicKeyList` directly and instead call a function which filters array to remove expired keys, and use that function everywhere. Therefore, you won't need to check `if (key.expiryTime > Date.now()) {`

lib/ts/recipe/session/sessionFunctions.ts
rishabhpoddar: - Consider renaming `foundOlderKey` to something like `foundASigningKeyThatIsOlderThanTheAccessToken`
- Add a comment in the code explaining the logic behind the if statement.

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: @porcellus perhaps you missed this comment.

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -87,77 +87,75 @@ export async function getSession(
...
-                // of the oldest key instead of the last update time
-                if (timeCreated >= key.createdAt) {
-                    foundOlderKey = true;
-                    break;
-                }
+            // This is kind of like the old signingKeyLastUpdated logic using the creation time
+            // of the oldest key instead of the last update time
+            if (timeCreated >= key.createdAt) {
+                foundASigningKeyThatIsOlderThanTheAccessToken = true;
+                break;
```
rishabhpoddar, line 153: Add a comment here explaining the logic behind the if statement.

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
porcellus: I added the comment right above the function, I think GitHub is just displaying it weirdly.

rishabhpoddar: Ah fair enough.. my bad. Will resolve this comment

lib/ts/recipe/session/sessionFunctions.ts
porcellus: There was already a comment above this, but I expanded it a bit anyway.

rishabhpoddar (approved):

PR #176, opened 2021-09-10
fixes #175

### LINKED ISSUE #175: Fix issues for 7.0
- [x] Error Handler in supertokens.ts needs to be async and all errorHandler functions in recipes must also be async and awaited for.
- [x] Session error callbacks need to be async / await as well (should change utils.ts normalising for this too)
- [x] Add a test for session callback error handler where you do something like:
   ```
   onUnauthorised: async (...) => {
      await new Promise(r => setTimeout(r), 1) // simulate async operation
      throw new Error("custom error") 
   }
   ```
   And then in your error handler, you check if you get the error with "custom error"

- [x] Need to change `verifySession` function signature to have optional options and not make it `| undefined`: https://github.com/supertokens/supertokens-node/blob/master/lib/ts/recipe/session/framework/express.ts#L21
- [x] Add example of using `verifySession()` in with-typescript test folder.
- [ ] Release as 7.0.1

bhumilsarvaiya:
## Summary of change

- Error Handler in supertokens.ts and all errorHandler functions in recipes changed to async and awaited for
- Session error callbacks are now async
- Changed `verifySession` function signature to have optional options

## Related issues

- #175 

## Test Plan
- added test `test session error handler overriding` in `sessionExpress.test.js`
- Added example of using verifySession() in with-typescript test folder

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/framework/awsLambda/framework.ts
+++ b/lib/ts/framework/awsLambda/framework.ts
@@ -286,7 +286,7 @@ export const middleware = (handler?: Handler): Handler => {
             }
             return response.sendResponse({});
         } catch (err) {
-            supertokens.errorHandler(err, request, response);
+            await supertokens.errorHandler(err, request, response);
             if (response.responseSet) {
                 return response.sendResponse({});
             }
--- a/lib/ts/framework/fastify/framework.ts
+++ b/lib/ts/framework/fastify/framework.ts
@@ -167,7 +167,7 @@ export const errorHandler = () => {
         let supertokens = SuperTokens.getInstanceOrThrowError();
         let request = new FastifyRequest(req);
         let response = new FastifyResponse(res);
-        supertokens.errorHandler(err, request, response);
+        await supertokens.errorHandler(err, request, response);
     };
 };
 
--- a/lib/ts/framework/koa/framework.ts
+++ b/lib/ts/framework/koa/framework.ts
@@ -148,7 +148,7 @@ export const middleware = () => {
                 return await next();
             }
         } catch (err) {
-            return supertokens.errorHandler(err, request, response);
+            return await supertokens.errorHandler(err, request, response);
         }
     };
 };
--- a/lib/ts/framework/loopback/framework.ts
+++ b/lib/ts/framework/loopback/framework.ts
@@ -133,7 +133,7 @@ export const middleware: Middleware = async (ctx: MiddlewareContext, next: Next)
         }
         return;
     } catch (err) {
-        return supertokens.errorHandler(err, request, response);
+        return await supertokens.errorHandler(err, request, response);
     }
 };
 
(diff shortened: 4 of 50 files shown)


lib/ts/framework/fastify/framework.ts
```diff
@@ -167,7 +167,7 @@ export const errorHandler = () => {
         let supertokens = SuperTokens.getInstanceOrThrowError();
         let request = new FastifyRequest(req);
         let response = new FastifyResponse(res);
-        supertokens.errorHandler(err, request, response);
+        await supertokens.errorHandler(err, request, response);
```
rishabhpoddar, line 170: If this throws an error, what will happen?

lib/ts/framework/koa/framework.ts
```diff
@@ -148,7 +148,7 @@ export const middleware = () => {
                 return await next();
             }
         } catch (err) {
-            return supertokens.errorHandler(err, request, response);
+            return await supertokens.errorHandler(err, request, response);
```
rishabhpoddar, line 151: If this throws an error, what will happen?

lib/ts/framework/awsLambda/framework.ts
```diff
@@ -286,7 +286,7 @@ export const middleware = (handler?: Handler): Handler => {
             }
             return response.sendResponse({});
         } catch (err) {
-            supertokens.errorHandler(err, request, response);
+            await supertokens.errorHandler(err, request, response);
```
rishabhpoddar, line 289: If this throws an error, what will happen?

lib/ts/framework/loopback/framework.ts
```diff
@@ -133,7 +133,7 @@ export const middleware: Middleware = async (ctx: MiddlewareContext, next: Next)
         }
         return;
     } catch (err) {
-        return supertokens.errorHandler(err, request, response);
+        return await supertokens.errorHandler(err, request, response);
```
rishabhpoddar, line 136: If this throws an error, what will happen?

rishabhpoddar (changes requested):

lib/ts/framework/fastify/framework.ts
bhumilsarvaiya: if the middleware is inside a try catch block, the error will go to that catch block. Else the lambda will throw the error and the API gateway return a 500 response

lib/ts/framework/awsLambda/framework.ts
bhumilsarvaiya: if the middleware is inside a try catch block, the error will go to that catch block. Else the lambda will throw the error and the API gateway return a 500 response

PR #189, opened 2021-10-13
Draft: feat: added possibility to add route options to supertokens route (Hapi framework)

ealionel:
Hello, this is my first PR in an open-source project :)

## Summary of change

It adds the possibility to add options to supertokens route when registering the Hapi plugin.

**Use case description :** 
I needed this in a project because I wanted to implement session verification as Hapi auth scheme ([see here](https://hapi.dev/tutorials/auth?lang=en_US#authenticate)) and not as a lifecycle method (`verifySession()`) like it is implemented right now.
However, by implementing the scheme into the project, auth is enabled by default on all routes, even on the routes added by supertokens. So calling when calling `/auth/session/refresh`, the method authenticating the call threw and it didn't refresh.

Now, I can just disable authentication on supertokens route this way :
```typescript
server.register(supertokensPlugin, {
  options: {
    routeOptions: { auth: false },
  }
})
```
## Documentation changes

- Maybe add documentation for this new option

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 .
 
+## [7.4.0] - 2021-10-11
+
+### Added
+
+-   Hapi framework: Adds possibility to pass `RouteOptions` when registering the plugin. These route options will be applied to all supported routes.
+
 ## [7.3.0] - 2021-10-11
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 .
 
-## [7.4.0] - 2021-10-11
+## [7.4.0] - 2021-10-13
 
 ### Added
 
--- a/lib/build/framework/hapi/framework.d.ts
+++ b/lib/build/framework/hapi/framework.d.ts
@@ -1,5 +1,5 @@
 // @ts-nocheck
-import type { Request, ResponseToolkit, Plugin, ResponseObject } from "@hapi/hapi";
+import type { Request, ResponseToolkit, Plugin, ResponseObject, RouteOptions } from "@hapi/hapi";
 import type { HTTPMethod } from "../../types";
 import { BaseRequest } from "../request";
 import { BaseResponse } from "../response";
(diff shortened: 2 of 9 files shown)


rishabhpoddar:
Thank you for this PR @ealionel. @bhumilsarvaiya will review the changes.

I have a question: In our `verifySession` function, we provide an option using which you can make session optional (See [this](https://supertokens.io/docs/thirdpartyemailpassword/common-customizations/verify-session#optional-session-verification)). If you used this in your auth scheme, would that also solve your issue? Or have I misunderstood something?

Thanks :))

rishabhpoddar:
I understand. A few more thoughts:

> I would have to check inside the authenticate the request URL and add the argument on the verifySession or not

If you add the optional flag without checking the routes, then the behaviour will be as follows:
- If a session doesn't exist, the session object passed to your API will be `undefined`, and if your API requires a session, you can then return a `401` status code to the frontend.
- If a session exists, then the session object will be defined and passed to your API for consumption.

Do you still think that adding an extra option is needed? (We will still merge this PR regard. Just curious about your thoughts on if my thinking is correct or not)

> make an auth scheme directly inside

Would this require us to use any function provided by Hapi? I ask cause then our lib will have to depend on Hapi to work, which would not be a good idea, since users who don't use Hapi would end up getting that in their node_modules. If a "non-type" import is not required though, we can do this (would be very helpful if you could make a PR for that 😄)

ealionel:
> Do you still think that adding an extra option is needed? 

After more thoughts, my solution is disabling the authentication for all supertokens route, but I think a better solution would be to implement directly into the code which supertokens route needs authentication or not.

> Would this require us to use any function provided by Hapi? I ask cause then our lib will have to depend on Hapi to work, which would not be a good idea, since users who don't use Hapi would end up getting that in their node_modules. If a "non-type" import is not required though, we can do this (would be very helpful if you could make a PR for that 😄)

The methods to implement only depends on interfaces (like your implementation for `verifySession` in Hapi framework).
I can try to propose a PR for that, don't merge this PR yet because after more thoughts I don't think this is the best solution for my use case !

rishabhpoddar:
>  my solution is disabling the authentication for all supertokens route

Even if you disable it for all routes provided by SuperTokens' middleware, that doesn't have any effect for those routes since we verify the session in those routes as well (for example [here](https://github.com/supertokens/supertokens-node/blob/master/lib/ts/recipe/emailverification/api/implementation.ts#L23) - this is the API to know if a user's email is verified or not). 

Here are the options I see possible:
1) Add a global middleware that is applied to all routes. This middleware is provided the `sessionRequired: false` option. In here, if a session doesn't exist, the target route will still run with the `session`object being `undefined`.
2) Add a global middleware without `sessionRequired: false`, but exclude it from running for the `/auth/*` path (all routes exposed by supertokens). In this case, for all your APIs (not SuperTokens APIs), a session will be required. If no session is given, the middleware will return a `401`.
3) Use the `verifySession` on a per API basis - only for the APIs that need session verification.

rishabhpoddar:
@ealionel any updates on this PR please? Which method do you think is the best? Thanks.

rishabhpoddar:
Closing this PR due to inactivity. Please feel free to repoen it.

PR #234, opened 2021-12-20
fixes #230

### LINKED ISSUE #230: NextJS - Type error when calling init functions
Hi.
I'm following the NextJS docs (https://supertokens.io/docs/emailpassword/nextjs/init), and on step 6 - "Call the init functions" we need to add the following function to the _app.ts file:

```ts
async function initNode() {
  const supertokensNode = await import('supertokens-node')
  const { backendConfig } = await import('../config/backendConfig')
  supertokensNode.init(backendConfig())
}
```

But I'm getting the following error on `supertokensNode.init(backendConfig())`:

```
Argument of type '{ framework: string; supertokens: { connectionURI: string; apiKey: string; }; appInfo: { appName: string; apiDomain: string; websiteDomain: string; apiBasePath: string; }; recipeList: RecipeListFunction[]; isInServerlessEnv: boolean; }' is not assignable to parameter of type 'TypeInput'.
  Types of property 'framework' are incompatible.
    Type 'string' is not assignable to type 'TypeFramework | undefined'.ts(2345)
```

Here's the backendConfig file that was auto-generated by the docs:

```ts
import EmailPasswordNode from "supertokens-node/recipe/emailpassword";
import SessionNode from "supertokens-node/recipe/session";
import { appInfo } from "./appInfo";

export const backendConfig = () => {
  return {
    framework: "express",
    supertokens: {
      // These are the connection details of the app you created on supertokens.io
      connectionURI:
        "<removed>",
      apiKey: "<removed>",
    },
    appInfo,
    recipeList: [EmailPasswordNode.init(), SessionNode.init()],
    isInServerlessEnv: true,
  };
};
```

I'm using supertokens-node@8.2.1

Thanks.

bhumilsarvaiya:
## Summary of change

Exposes types files

## Related issues

-   #230

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

## TODO:
- [ ] Edit docs for nextjs to use TypeInput in backendConfig function


rishabhpoddar:
This should be issued against branch 8.3

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 .
 
+## [8.2.2] - 2021-12-20
```
rishabhpoddar, line 9: Against branch 8.3

types/index.d.ts
```diff
@@ -0,0 +1,10 @@
+export * from "../lib/build/types";
```
rishabhpoddar, line 1: Also, edit playground app in the test folder (it's called with-typescript) to make use of these types for all recipes and check that it's working as expected.

rishabhpoddar (changes requested):

types/index.d.ts
bhumilsarvaiya: done

CHANGELOG.md
bhumilsarvaiya: done

rishabhpoddar:
As discussed, types should be the end of the path, and not in the middle.

PR #242, opened 2022-01-14
Reset password resp

porcellus:
## Summary of change

Added userId as an optional property to the response of `recipe/user/password/reset`. It's marked optional to be compatible with Cores not supporting 2.11.

## Related issues

-   supertokens/supertokens-core#327

## Test Plan

Type change

## Documentation changes

Not necessary

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Adds compatibility with FDI 1.12 and CDI 2.11
 -   Adds passwordless tests
 
+### Changed
+
+-   Added userId as an optional property to the response of `recipe/user/password/reset`.
```
rishabhpoddar, line 21: Should be in its own version here.

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -200,7 +200,17 @@ export type RecipeInterface = {
     resetPasswordUsingToken(input: {
         token: string;
         newPassword: string;
-    }): Promise<{ status: "OK" | "RESET_PASSWORD_INVALID_TOKEN_ERROR" }>;
+    }): Promise<
```
rishabhpoddar, line 203: Need to also change `passwordResetPOST`?

lib/ts/recipe/thirdpartyemailpassword/types.ts
```diff
@@ -205,7 +205,17 @@ export type RecipeInterface = {
     resetPasswordUsingToken(input: {
         token: string;
         newPassword: string;
-    }): Promise<{ status: "OK" | "RESET_PASSWORD_INVALID_TOKEN_ERROR" }>;
+    }): Promise<
```
rishabhpoddar, line 208: Need to also change `passwordResetPOST`?

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/types.ts
porcellus: If I remember correctly when we discussed this we did not want to add the `userId` to the FDI.

rishabhpoddar: Yes. We don't want to add it to the FDI. But we still want to add it as a result of this function since it can be useful for overrides. But we won't send it to the frontend (hence no change to FDI)

rishabhpoddar: I have made this change already.

lib/ts/recipe/thirdpartyemailpassword/types.ts
rishabhpoddar: I have made this change already.

rishabhpoddar (approved):

PR #276, opened 2022-03-11
feat(debugger): adds logger for debugging

jscyo:
## Summary of change

Adds debug logging functions

## Related issues

-   https://github.com/supertokens/for-zenhub/issues/111

## Test Plan

Manually tested. Added an info log to the SignIn API and an error log to EmailExistsAPI(if email exists) and called signin, then signed out and tried to sign up with the same email and got an email exists error.
logger_test

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   none


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [9.1.0] - 2022-03-11
```
rishabhpoddar, line 8: Put this in unreleased section.

lib/ts/constants.ts
```diff
@@ -15,3 +15,9 @@
 
 export const HEADER_RID = "rid";
 export const HEADER_FDI = "fdi-version";
+
+const SUPERTOKENS_NAMESPACE = "com.supertokens:";
+const ERROR_ID = "error";
```
rishabhpoddar, line 20: No error ones for now. Make this into debug

lib/ts/utils.ts
```diff
@@ -122,3 +124,36 @@ export function validateTheStructureOfUserInput(config: any, inputSchema: any, c
...
+
+export function infoLogger(message: string) {
+    debug(`${SUPERTOKENS_NAMESPACE}${INFO_ID}`)(`{t: "${new Date().toISOString()}", msg: "${message}"}`);
+}
+
+export let errorLoggerWithCode = {
+    1: (item: string) => {
+        errLoggerHelper(ERROR_CODE_1, "Example error with param: " + item);
+    },
+};
```
rishabhpoddar, line 157: Move this all into its own file at ts/logger.ts

lib/ts/utils.ts
```diff
@@ -122,3 +124,36 @@ export function validateTheStructureOfUserInput(config: any, inputSchema: any, c
...
+    infoLogger("some info")
+
+Output: (with DEBUG=com.supertokens:*)
+  com.supertokens:error {t: "2022-03-09T10:47:57.379Z", errCode: 1, msg: "Example error with param Test", sdkVer: "9.0.0"} +0ms
+  com.supertokens:info {t: "2022-03-09T10:47:57.380Z", msg: "Some info", sdkVer: "9.0.0"} +0ms
+
+*/
+
+// ERROR_CODE values should be the same as the key in the errorLoggerWithCode function
+const ERROR_CODE_1 = 1;
```
rishabhpoddar, line 141: This is not needed.. overkill

lib/ts/utils.ts
```diff
@@ -122,3 +124,36 @@ export function validateTheStructureOfUserInput(config: any, inputSchema: any, c
...
+
+let errLoggerHelper = (errorCode: number, message: string) => {
+    debug(`${SUPERTOKENS_NAMESPACE}${ERROR_ID}`)(
+        `{t: "${new Date().toISOString()}", errCode: ${errorCode}, msg: "${message}", sdkVer: "${version}"}`
```
rishabhpoddar, line 145: Make timestamp in milliseconds

lib/ts/utils.ts
```diff
@@ -122,3 +124,36 @@ export function validateTheStructureOfUserInput(config: any, inputSchema: any, c
...
+    );
+};
+
+export function infoLogger(message: string) {
+    debug(`${SUPERTOKENS_NAMESPACE}${INFO_ID}`)(`{t: "${new Date().toISOString()}", msg: "${message}"}`);
+}
+
+export let errorLoggerWithCode = {
+    1: (item: string) => {
+        errLoggerHelper(ERROR_CODE_1, "Example error with param: " + item);
```
rishabhpoddar, line 155: Don't have an example error code in the SDK. Actually use it in one place in the code

lib/ts/constants.ts
rishabhpoddar: Also, no point in having this constant if it is used in just one place.

lib/ts/constants.ts
```diff
@@ -15,3 +15,9 @@
...
+const INFO_ID = "info";
```
rishabhpoddar, line 21: Also, no point in having this constant if it is used in just one place.

lib/ts/constants.ts
```diff
@@ -15,3 +15,9 @@
 
 export const HEADER_RID = "rid";
 export const HEADER_FDI = "fdi-version";
+
+const SUPERTOKENS_NAMESPACE = "com.supertokens:";
```
rishabhpoddar, line 19: Rename to `SUPERTOKENS_LOGGER_NAMESPACE`

lib/ts/utils.ts
```diff
@@ -122,3 +124,36 @@ export function validateTheStructureOfUserInput(config: any, inputSchema: any, c
...
+const ERROR_CODE_1 = 1;
+
+let errLoggerHelper = (errorCode: number, message: string) => {
+    debug(`${SUPERTOKENS_NAMESPACE}${ERROR_ID}`)(
+        `{t: "${new Date().toISOString()}", errCode: ${errorCode}, msg: "${message}", sdkVer: "${version}"}`
+    );
+};
+
+export function infoLogger(message: string) {
+    debug(`${SUPERTOKENS_NAMESPACE}${INFO_ID}`)(`{t: "${new Date().toISOString()}", msg: "${message}"}`);
```
rishabhpoddar, line 150: Missing sdkVersion??

rishabhpoddar (changes requested):

lib/ts/logger.ts
```diff
@@ -0,0 +1,36 @@
...
+Output: (with DEBUG=com.supertokens:*)
+  com.supertokens:info {t: "1647263677740", msg: "Calling SignInAPI", sdkVer: "9.1.0"} +0ms
+  com.supertokens:debug {t: "1647263677952", msg: "API replied with status: OK", debugCode: 1, sdkVer: "9.1.0"} +0ms
+  com.supertokens:info {t: "1647263690053", msg: "Calling SignInAPI", sdkVer: "9.1.0"} +0ms
+  com.supertokens:debug {t: "1647263690199", msg: "API replied with status: WRONG_CREDENTIALS_ERROR", debugCode: 1, sdkVer: "9.1.0"} +0ms
+
+*/
+
+let logMessage = (namespaceId: string, message: string) => {
+    debug(`com.supertokens:${namespaceId}`)(`{t: "${Date.now()}", msg: ${message}, sdkVer: "${version}"}`);
```
rishabhpoddar, line 21: Some indication of which file (ts file) the message is being called from.

lib/ts/logger.ts
```diff
@@ -0,0 +1,36 @@
...
+let debugLoggerHelper = (debugCode: number, message: string) => {
+    logMessage("debug", `"${message}", debugCode: ${debugCode}`);
+};
+
+export function infoLogger(message: string) {
+    logMessage("info", `"${message}"`);
+}
+
+export let debugLoggerWithCode = {
+    1: (item: string) => {
```
rishabhpoddar, line 33: Change to `1: (apiName: string, status: string)`

lib/ts/recipe/emailpassword/api/signin.ts
```diff
@@ -16,10 +16,11 @@
 import { send200Response } from "../../../utils";
 import { validateFormFieldsOrThrowError } from "./utils";
 import { APIInterface, APIOptions } from "../";
+import { debugLoggerWithCode, infoLogger } from "../../../logger";
 
 export default async function signInAPI(apiImplementation: APIInterface, options: APIOptions): Promise<boolean> {
     // Logic as per https://github.com/supertokens/supertokens-node/issues/20#issuecomment-710346362
-
+    infoLogger("Calling SignInAPI");
```
rishabhpoddar, line 23: You can have codes for info logger as well.

lib/ts/logger.ts
```diff
@@ -0,0 +1,36 @@
...
+  com.supertokens:debug {t: "1647263690199", msg: "API replied with status: WRONG_CREDENTIALS_ERROR", debugCode: 1, sdkVer: "9.1.0"} +0ms
+
+*/
+
+let logMessage = (namespaceId: string, message: string) => {
+    debug(`com.supertokens:${namespaceId}`)(`{t: "${Date.now()}", msg: ${message}, sdkVer: "${version}"}`);
+};
+
+let debugLoggerHelper = (debugCode: number, message: string) => {
+    logMessage("debug", `"${message}", debugCode: ${debugCode}`);
```
rishabhpoddar, line 25: Changing message to have debugCode is misleading.

rishabhpoddar (changes requested):

lib/ts/logger.ts
```diff
@@ -0,0 +1,79 @@
...
+    let messageWithOptionalCode = `msg: \"${message}\"`;
+    if (code != undefined) {
+        messageWithOptionalCode = `${messageWithOptionalCode}, code: ${code}`;
+    }
+    debug(`com.supertokens:${namespaceId}`)(
+        `{t: "${Date.now()}", ${messageWithOptionalCode}, file: \"${fileNameAndLineNumber}\" sdkVer: "${version}"}`
+    );
+};
+
+export let infoLoggerWithCode = {
```
porcellus, line 50: Using log levels as namespaces is a bit weird for me, because if I enable debug logs (`DEBUG=com.supertokens:debug`) then I'd not get the info logs at all which is not what I'd expect. Maybe we could use a lib that supports log levels, e.g.: debug-level.
The namespace could also include "API", maybe even the recipe name.

lib/ts/logger.ts
```diff
@@ -0,0 +1,79 @@
...
+    [loggerCodes.API_CALLED]: (apiName: string) => {
+        logMessage("info", `${apiName} is called`, loggerCodes.API_CALLED);
+    },
+};
+
+export let debugLoggerWithCode = {
+    [loggerCodes.API_RESPONSE]: (apiName: string, status: string) => {
```
porcellus, line 57: I think the call log should be debug-level and the response info since the response log includes more information and it's more likely to be useful on its own.

lib/ts/recipe/emailpassword/api/signin.ts
```diff
@@ -16,10 +16,11 @@
 import { send200Response } from "../../../utils";
 import { validateFormFieldsOrThrowError } from "./utils";
 import { APIInterface, APIOptions } from "../";
+import { debugLoggerWithCode, infoLoggerWithCode, loggerCodes } from "../../../logger";
 
 export default async function signInAPI(apiImplementation: APIInterface, options: APIOptions): Promise<boolean> {
     // Logic as per https://github.com/supertokens/supertokens-node/issues/20#issuecomment-710346362
-
+    infoLoggerWithCode[loggerCodes.API_CALLED]("SignInAPI");
```
porcellus, line 23: This call signature is fairly ugly; I think it'd be better if we exported these log functions as named exports and used them like `logAPICalled("...")` instead. This has the added benefit that if we later want to tweak the level of a message type, we'd only have to update it in one place. I think the API implementations should avoid knowing/caring about log levels.

lib/ts/recipe/emailpassword/api/signin.ts
```diff
@@ -36,11 +37,13 @@ export default async function signInAPI(apiImplementation: APIInterface, options
     let result = await apiImplementation.signInPOST({ formFields, options, userContext: {} });
 
     if (result.status === "OK") {
+        debugLoggerWithCode[loggerCodes.API_RESPONSE]("SignInAPI", result.status);
```
porcellus, line 40: I'd prefer if both of these kinds of logs were done in a single place (`supertokens.ts` or `recipeModule.ts`). A few more things:
- We should log errors (e.g., 500) if we log normal responses.
- How does this interact with user overrides?
- If the API is disabled, what do we log? Maybe add a separate log for that.

porcellus (changes requested):

lib/ts/logger.ts
jscyo: >  Using log levels as namespaces is a bit weird for me, because if I enable debug logs (DEBUG=com.supertokens:debug) then I'd not get the info logs at all which is not what I'd expect. Maybe we could use a lib that supports log levels, e.g.: debug-level.

Makes sense, when we were initially discussing this we only had info and error loggers so the choice to enable them through the namespace didn't seem too strange, but yeah, using a lib that supports log levels makes more sense now.

> The namespace could also include "API", maybe even the recipe name.
Having recipes as namespaces was discussed initially, I think we decided to make it as simple as possible in the beginning so we went with log levels as the namespace

lib/ts/logger.ts
jscyo: Agreed

lib/ts/logger.ts
```diff
@@ -0,0 +1,79 @@
...
+    SignIn with with valid credentials;
+    SignIn with invalid credentials
+Output: (with DEBUG=com.supertokens:*)
+  com.supertokens:info {t: "1647340361090", msg: "SignInAPI is called", code: 2, file: "/home/supertokens-node/lib/build/recipe/emailpassword/api/signin.js:54:69", sdkVer: "9.1.0"} +0ms
+  com.supertokens:debug {t: "1647340361295", msg: "SignInAPI replied with status: OK", code: 1, file: "/home/supertokens-node/lib/build/recipe/emailpassword/api/signin.js:65:76", sdkVer: "9.1.0"} +0ms
+  com.supertokens:info {t: "1647340563421", msg: "SignInAPI is called", code: 2, file: "/home/supertokens-node/lib/build/recipe/emailpassword/api/signin.js:54:69", sdkVer: "9.1.0"} +0ms
+  com.supertokens:debug {t: "1647340563565", msg: "SignInAPI replied with status: WRONG_CREDENTIALS_ERROR", code: 1, file: "/home/supertokens-node/lib/build/recipe/emailpassword/api/signin.js:71:76", sdkVer: "9.1.0"} +0ms
+*/
+
+export let loggerCodes = {
```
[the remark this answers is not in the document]
jscyo: > What are these codes used for?

The initial thought behind codes was that if there were common issues between users and we found common patterns in the debug logs we could have a section in our docs for troubleshooting like "Hey if you got this `sequence of codes` this is how you fix it". 

> It should be a const or even better, an enum. In general, you should prefer const to let.

Yup will make the change

lib/ts/recipe/emailpassword/api/signin.ts
jscyo: Agreed

lib/ts/recipe/emailpassword/api/signin.ts
jscyo: > I'd prefer if both of these kinds of logs were done in a single place (supertokens.ts or recipeModule.ts)

Sure

> - We should log errors (e.g., 500) if we log normal responses.
> - If the API is disabled, what do we log? Maybe add a separate log for that.

Sure, the logs that have been added are only example logs. The idea was that over time we would add more logs

lib/ts/logger.ts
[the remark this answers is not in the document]
porcellus: A "sequence of code" sounds a bit hard to search. I thought they were for looking up more info about each log item (that's how I've seen these used, or localized messages). That may come in handy if/when we have some harder-to-understand logs where users need more explanation.
Anyway, it doesn't complicate things too much, and it may be helpful; I'm just curious. I'll resolve this after the second point is addressed in the code.

jscyo:
Hey @porcellus , so in a discussion with Rishabh he mentioned that we should keep this logger as simple as possible. These were his requirements:
- User can enable our logging through an env variable
- Just one single log function, no need for different namespaces or debug levels for now

lib/ts/logger.ts
```diff
@@ -0,0 +1,49 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+import debug from "debug";
+import { version } from "./version";
+
+/*
+ The debug logger below can be used to log debug messages in the following format
+    com.supertokens {t: "1647508320340", message: Your message, file: "/home/supertokens-node/lib/build/supertokens.js:231:18" sdkVer: "9.1.0"} +0m
```
porcellus, line 21: Please update the format in the comment

porcellus (changes requested):

lib/ts/logger.ts
jscyo: I've updated the comment

porcellus (approved):
LGTM

lib/ts/logger.ts
```diff
@@ -0,0 +1,49 @@
...
+    com.supertokens {t: "2022-03-18T11:15:24.608Z", message: Your message, file: "/home/supertokens-node/lib/build/supertokens.js:231:18" sdkVer: "9.2.0"} +0m
+*/
+
+function logDebugMessage(message: string) {
+    debug(`com.supertokens`)(
+        `{t: "${new Date().toISOString()}", message: \"${message}\", file: \"${getFileLocation()}\" sdkVer: "${version}"}`
+    );
+}
+
+let getFileLocation = () => {
```
rishabhpoddar, line 30: You need to make sure that this function is not called if the user has not enabled logging. Cause creating errors can be expensive.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

lib/ts/logger.ts
```diff
@@ -0,0 +1,54 @@
...
+
+function logDebugMessage(message: string) {
+    debug(SUPERTOKENS_DEBUG_NAMESPACE)(
+        `{t: "${new Date().toISOString()}", message: \"${message}\", file: \"${getFileLocation()}\" sdkVer: "${version}"}`
+    );
+}
+
+let getFileLocation = () => {
+    if (!debug.enabled(SUPERTOKENS_DEBUG_NAMESPACE)) {
+        return;
```
rishabhpoddar, line 33: You are returning `undefined` here. But you should return `"N/A"`?

rishabhpoddar (changes requested):

PR #277, opened 2022-03-11
feat(usermetadata): add support for usermetadata w/ tests

porcellus:
## Summary of change

Adds support for usermetadata

## Related issues

-   supertokens/supertokens-core#381

## Test Plan

Added basic tests for all functions

## Documentation changes

TODO

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Documentation


lib/ts/recipe/usermetadata/types.ts
```diff
@@ -0,0 +1,75 @@
...
+export type JSONPrimitive = string | number | boolean | null;
+export type JSONArray = Array<JSONPrimitive>;
+export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
+export interface JSONObject {
+    [ind: string]: JSONValue;
+}
+
+export type TypeInput = {
+    override?: {
+        functions?: (
```
porcellus, line 27: I don't think an `apis` prop would be good here: it can only be empty and this way it's even more clear that this recipe provides no APIs. On the other hand, all other recipes have it...

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [9.1.0] - Unreleased
```
rishabhpoddar, line 8: Unreleased should not have a version associated with it. It's OK for now, but just FYI

lib/ts/recipe/usermetadata/types.ts
rishabhpoddar: We should have it, and it can be empty.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
...
+
+# Adds
+
+-   Adds UserMetadata recipe
```
rishabhpoddar, line 12: Need to discuss:
- By default, should we tie this recipe to other recipes? For example, other auth recipes do an init of user metadata recipe in them, and when fetching user info, they also fetch the user's metadata info and return it to the user.
- In emailpassword recipe, we allow custom form fields, should that be saved in metadata automatically by default?
- How will this be tied to Session grants? And should we do something about this now?

rishabhpoddar (changes requested):

CHANGELOG.md
porcellus: Ok, let's discuss this tomorrow, but my feelings so far:
- I don't think so, I think it should be a separate recipe. 
  - If we want to tie everything into the auth recipes their config will bloat and we'd also have to do it on the frontend which would increase the bundle size.
  - I don't think we should automatically fetch this info. We shouldn't control how/when they are refreshed and if (or how) they are cached.
- We've discussed this in the progressive profiling part of the session grants docs, but I don't think we need to do anything for now. When session grants are implemented we export a grant in this recipe for requiring/validating fields in the metadata but I think that's about it.

CHANGELOG.md
porcellus: Yeah, I just wanted to signify that this will be a separate major version. Next time I have extra time, I should look into changie. It'd simplify this a fair bit, I think.

CHANGELOG.md
rishabhpoddar: I agree with everything, except for adding a new config to the ep recipe. So resolving this

CHANGELOG.md
rishabhpoddar: > Yeah, I just wanted to signify that this will be a separate major version

Major or minor version?

rishabhpoddar:
@porcellus we should keep this as a recipe only as this is something that is user facing. If we want to make it an ingredient, then we will have to expose it via all the other recipes (as we don't want to introduce another terminology for users). 

So no change required from this point of view.

rishabhpoddar (approved):

CHANGELOG.md
porcellus: Minor, all changes are backward compatible

lib/ts/recipe/usermetadata/types.ts
```diff
@@ -0,0 +1,79 @@
...
+     *   - result: `{ "preferences": { "theme":"dark" }, "notifications": { "sms": true } }`
+     */
+    updateUserMetadata: (input: {
+        userId: string;
+        metadataUpdate: JSONObject;
+        userContext: any;
+    }) => Promise<{
+        status: "OK";
+        metadata: JSONObject;
+    }>;
```
rishabhpoddar, line 71: In other languages, there is no value for `null` and `undefined`. So it's not easily possible to express if a field needs to be deleted or not updated in those languages.

To make the SDKs consistent, we should not allow passing of `null` against fields in the `metadataUpdate` object. Instead, we should have different functions to delete a specific field:
- Either we change `clearUserMetadata` to -> `clearUserMetadata({userId: string, fieldName?: string, userContext: any})` in which if a fieldName is passed, only that is cleared, otherwise all of the fields are cleared
- Create a new function like `clearUserMetadataField({userId: string, fieldName: string, userContext: any})`.

rishabhpoddar (changes requested):

lib/ts/recipe/usermetadata/types.ts
rishabhpoddar: And if someone passes `null` for any of the fields in `metadataUpdate`, we can throw an error from the function telling them to use `clearUserMetadataField` for that field instead.

PR #288, opened 2022-04-05
sms delivery skeleton added

bhumilsarvaiya:
## Summary of change

- smsDelivery Ingredient added

> Twilio service will be added in the next PR

## Related PR

-   #274

### DIFF
--- a/lib/ts/ingredients/smsdelivery/types.ts
+++ b/lib/ts/ingredients/smsdelivery/types.ts
@@ -1,4 +1,3 @@
-"use strict";
 /* Copyright (c) 2021, VRAI Labs and/or its affiliates. All rights reserved.
  *
  * This software is licensed under the Apache License, Version 2.0 (the
@@ -13,4 +12,19 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-Object.defineProperty(exports, "__esModule", { value: true });
+import OverrideableBuilder from "supertokens-js-override";
+
+export type SmsDeliveryInterface<T> = {
+    sendSms: (input: T & { userContext: any }) => Promise<void>;
+};
+
+/**
+ * config class parameter when parent Recipe create a new SmsDeliveryRecipe object via constructor
+ */
+export interface TypeInput<T> {
+    service: SmsDeliveryInterface<T>;
+    override?: (
+        originalImplementation: SmsDeliveryInterface<T>,
+        builder: OverrideableBuilder<SmsDeliveryInterface<T>>
+    ) => SmsDeliveryInterface<T>;
+}
--- a/lib/ts/ingredients/smsdelivery/types.ts
+++ b/lib/ts/ingredients/smsdelivery/types.ts
@@ -22,6 +22,14 @@ export type SmsDeliveryInterface<T> = {
  * config class parameter when parent Recipe create a new SmsDeliveryRecipe object via constructor
  */
 export interface TypeInput<T> {
+    service?: SmsDeliveryInterface<T>;
+    override?: (
+        originalImplementation: SmsDeliveryInterface<T>,
+        builder: OverrideableBuilder<SmsDeliveryInterface<T>>
+    ) => SmsDeliveryInterface<T>;
+}
+
+export interface TypeInputWithService<T> {
     service: SmsDeliveryInterface<T>;
     override?: (
         originalImplementation: SmsDeliveryInterface<T>,
--- a/lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
+++ b/lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
@@ -32,7 +32,7 @@ function defaultCreateAndSendCustomSms(_: NormalisedAppinfo) {
         if (process.env.TEST_MODE === "testing") {
             return;
         }
-        // TODO
+        // TODO: Skip in this PR
     };
 }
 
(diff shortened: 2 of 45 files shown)


lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
```diff
@@ -0,0 +1,90 @@
...
+            // Full url that the end-user can click to finish sign in/up
+            urlWithLinkCode?: string;
+            codeLifetime: number;
+        },
+        _: any
+    ): Promise<void> => {
+        if (process.env.TEST_MODE === "testing") {
+            return;
+        }
+        // TODO
```
rishabhpoddar, line 35: TODO

lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
```diff
@@ -0,0 +1,90 @@
...
+                // Full url that the end-user can click to finish sign in/up
+                urlWithLinkCode?: string;
+                codeLifetime: number;
+                // Unlikely, but someone could display this (or a derived thing) to identify the device
+                preAuthSessionId: string;
+            },
+            userContext: any
+        ) => Promise<void>
+    ) {
+        this.appInfo = appInfo;
```
rishabhpoddar, line 73: We don't need this as an instance variable since it's not being used outside of the constructor anyway

lib/ts/ingredients/smsdelivery/types.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type SmsDeliveryInterface<T> = {
+    sendSms: (input: T & { userContext: any }) => Promise<void>;
+};
+
+/**
+ * config class parameter when parent Recipe create a new SmsDeliveryRecipe object via constructor
+ */
+export interface TypeInput<T> {
+    service: SmsDeliveryInterface<T>;
```
rishabhpoddar, line 25: Shouldn't `service` be optional too? (Same goes for email sending service as well). The reason why it should be optional here is that:
- The user may want to override fully the way to send SMS / email and then this service param makes no sense anymore.
- You are anyway handling the case when the server param is `undefined` in the `getSmsDeliveryConfig()` function.

Finally, the return type of `getSmsDeliveryConfig` should not be same as `TypeInput` since there, we want `service` to always be defined. So you will require two types:
```ts
export interface TypeInput<T> {
    service?: SmsDeliveryInterface<T>;
    override?: (
        originalImplementation: SmsDeliveryInterface<T>,
        builder: OverrideableBuilder<SmsDeliveryInterface<T>>
    ) => SmsDeliveryInterface<T>;
}
```
and
```ts
export interface TypeInputWithService<T> {
    service: SmsDeliveryInterface<T>;
    override?: (
        originalImplementation: SmsDeliveryInterface<T>,
        builder: OverrideableBuilder<SmsDeliveryInterface<T>>
    ) => SmsDeliveryInterface<T>;
}
```

The input to the `constructor` in `/ingredients/smsdelivery/index.ts` should be `TypeInputWithService`

lib/ts/recipe/passwordless/utils.ts
```diff
@@ -79,15 +79,49 @@ export function validateAndNormaliseUserInput(
...
+                       *    service: smsService,
+                       *    ...config.smsDelivery,
+                       * };
+                       *
+                       * and if the user has passed service as undefined,
+                       * it it again get set to undefined, so we
+                       * set service at the end
+                       */
+                      service: smsService,
+                  };
```
rishabhpoddar, line 117:
```suggestion
            {
                      ...config.smsDelivery,
                      /**
                       * if we do
                       * let smsDelivery = {
                       *    service: smsService,
                       *    ...config.smsDelivery,
                       * };
                       *
                       * and if the user has passed service as undefined,
                       * it it again get set to undefined, so we
                       * set service at the end
                       */
                      service: smsService,
                  };
```

rishabhpoddar: The reason for not doing `config.contactMethod === "EMAIL" ?..` is that `smsService` is on the root level now. Same goes for emailsending service.

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
bhumilsarvaiya: Skip for this PR

lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
bhumilsarvaiya: done

lib/ts/ingredients/smsdelivery/types.ts
bhumilsarvaiya: made the changes for smsDelivery. changes for emailDelivery would be done in other PR

lib/ts/recipe/passwordless/utils.ts
bhumilsarvaiya: done

lib/ts/ingredients/smsdelivery/types.ts
rishabhpoddar: The changes for SMS delivery for this are fine. Need to make same for email delivery. Please do that in this PR only

rishabhpoddar (approved):

Issue #313, opened 2022-05-24
JSON parsing in middleware not working correctly


rishabhpoddar:
The code incorrectly assumes that `jsonParser` function from `body-parser` is always sync. This causes an issue that the API is called before the JSON parser is finished.

PR #314, opened 2022-05-24
fix: assertThatBodyParserHasBeenUsedForExpressLikeRequest

bhumilsarvaiya:
## Summary of change

- added `fromNextJS` property for next js request

## Related issues

-   #313

### Test

- [x] disable nextjs parser and verify that the json parser deosn't get stuck and also parses the request body correctly
- [x] disable nextjs praser, modify the code to disable json parser. Call a POST API with correct request body and the request should fail
- [x] send empty json as request body and the request should not get stuck during the json parsing part
- [x] send no request body and the request should not get stuck during the json parsing part
- [x] send correct request body and the API should yield correct response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [9.2.2] - 2022-05-24
+
+### Fixes:
+
+-   Calling the setImmediate function inside assertThatBodyParserHasBeenUsedForExpressLikeRequest only if the function is getting executed in NextJS env
+
 ## [9.2.1] - 2022-05-19
 
 ### Fixes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixes:
 
--   Calling the setImmediate function inside assertThatBodyParserHasBeenUsedForExpressLikeRequest only if the function is getting executed in NextJS env
+-   Calling the setImmediate function inside assertThatBodyParserHasBeenUsedForExpressLikeRequest only if the function is getting executed in NextJS env. Fixes #313
 
 ## [9.2.1] - 2022-05-19
 
--- a/lib/ts/framework/utils.ts
+++ b/lib/ts/framework/utils.ts
@@ -110,7 +110,7 @@ function JSONCookies(obj: any) {
 
 export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(
     method: HTTPMethod,
-    request: Request | NextApiRequest
+    request: (Request | NextApiRequest) & { fromNextJS?: true }
 ) {
     // according to https://github.com/supertokens/supertokens-node/issues/33
     if (method === "post" || method === "put") {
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [9.2.2] - 2022-05-24
+
+### Fixes:
+
+-   Calling the setImmediate function inside assertThatBodyParserHasBeenUsedForExpressLikeRequest only if the function is getting executed in NextJS env
```
rishabhpoddar, line 14: link to github issue

lib/ts/framework/utils.ts
```diff
@@ -136,17 +136,19 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(
...
+                     * where the json parser would not resolve and thus the request
+                     * just hangs forever. Next.JS does json parsing on its own.
+                     */
+                    setImmediate(() => {
+                        if (!resolvedCalled) {
+                            resolvedCalled = true;
+                            resolve(undefined);
+                        }
+                    });
+                }
```
rishabhpoddar, line 151: If it's next js, we don't need to parse anything right? So you can remove the setImmediate and the jsonParser if this if statmeent is true??

lib/ts/nextjs.ts
```diff
@@ -39,6 +39,7 @@ export default class NextJS {
         response: any
     ): Promise<T> {
         return new Promise<T>(async (resolve: any, reject: any) => {
+            request.fromNextJS = true;
```
rishabhpoddar, line 42: I prefer a name like `request.__supertokensFromNextJS` just so it's clear this is a var from us

rishabhpoddar, line 42: You should add a test that if this wrapper is used, then the request has this variable.

rishabhpoddar (changes requested):

lib/ts/framework/utils.ts
bhumilsarvaiya: nextjs do allow users to disable the default parser. In that case, we should be parsing the request body

lib/ts/framework/utils.ts
```diff
@@ -137,16 +137,22 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(
             let err = await new Promise((resolve) => {
                 let resolvedCalled = false;
                 /**
-                 * the setImmediate here is to counter the next.js issue
-                 * where the json parser would not resolve and thus the request
-                 * just hangs forever. Next.JS does json parsing on its own.
+                 * Nextjs allow users to disable the default parser.
+                 * To handle that scenario, we are still parsing the request body
                  */
```
rishabhpoddar, line 142: Need to test that if user has disabled bodyParser in NextJS, our parser runs and it doesn't get "stuck"

test/nextjs.test.js
```diff
@@ -404,6 +404,100 @@ describe(`NextJS Middleware Test: ${printPath("[test/nextjs.test.js]")}`, functi
...
+                return done();
+            });
+
+            superTokensNextWrapper(
+                async (next) => {
+                    return middleware()(request, response, next);
+                },
+                request,
+                response
+            );
```
rishabhpoddar, line 497: how is this test passing if the middleware is added after the request is being made?

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #324, opened 2022-06-07
feat: adds User Roles Recipe

jscyo:
## Summary of change

Adds user roles recipe

## Related issues
-  https://github.com/supertokens/for-zenhub/issues/124

## Test Plan

- TODO

## Documentation changes
- TODO

## Checklist for important updates

-   [ ] Changelog has been updated
-   [ ] `coreDriverInterfaceSupported.json` file has been updated (if needed)
    -   Along with the associated array in `lib/ts/version.ts`
-   [ ] `frontendDriverInterfaceSupported.json` file has been updated (if needed)
-   [ ] Changes to the version if needed
    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [ ] Had run `npm run build-pretty`
-   [ ] Had installed and ran the pre-commit hook
-   [ ] Issue this PR against the latest non released version branch.
    -   To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.
    -   If no such branch exists, then create one from the latest released branch.
-   [ ] If have added a new web framework, update the `add-ts-no-check.js` file to include that
-   [ ] If added a new recipe / api interface, then make sure that the implementation of it uses NON arrow functions only (like `someFunc: function () {..}`).
-   [ ] If added a new recipe, then make sure to expose it inside the recipe folder present in the root of this repo. We also need to expose its types.

## Remaining TODOs for this PR

-   [ ] Item1
-   [ ] Item2

PR #330, opened 2022-06-16
test: Adds userrole tests

jscyo:
## Summary of change

Adds recipe function tests

## Related issues
https://github.com/supertokens/supertokens-node/pull/324

## Test Plan
- addRoleToUserTests
  - add a role to a user and check that the user has the role
  - add a the same role to a user twice
  -  add an unknown role to the user
- Config test
- createNewRoleOrAddPermissions test
   - create a new role and check the output
   - create the same role twice
   - create a role with permissions
   - create a role with permissions, add new permissions to the role
   - create a role with duplicate permissions  
- deleteRole
  - create roles, add them to a user, delete one of the roles
  -  delete a role that does not exist
- getPermissionsForRole:
  - create a role with permissions, and check that the role has those permissions
  - get permissions for an unknown role
- getRolesForUser
  - create roles, add them to a user check that the user has the roles
- getRolesThatHavePermission
   - create multiple roles with the same permission. call the function and check for the output
   - get roles for unknown permission
- getUsersForRole
  - create a role, add it, multiple users, and retrieve the users who have the role
  - get users for an unknown role
- removePermissionsFromRole
  - create a role with permissions,  remove one permission from the role, check that it is removed
  - remove permissions from an unknown role
- removeUserRole 
  - create a role, add it to a user, remove it from the user and check that it is removed
  - remove a role the user does not have
  - remove an unknown role from a user 

## Documentation changes

-   ~~[ ] Changelog has been updated~~
-   ~~[ ] `frontendDriverInterfaceSupported.json` file has been updated (if needed)~~
    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   ~~[ ] If have added a new web framework, update the `add-ts-no-check.js` file to include that~~

-

### DIFF
--- a/.github/workflows/pre-commit-hook-run.yml
+++ b/.github/workflows/pre-commit-hook-run.yml
@@ -20,5 +20,5 @@ jobs:
               with:
                   node-version: "16"
             - run: git init && git add --all && git -c user.name='test' -c user.email='test@example.com' commit -m 'init for pr action'
-            - run: npm i
+            - run: npm i --force
             - run: ./hooks/pre-commit.sh
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
                   node-version: "16"
             - name: Make git use https instead of ssh
               run: git config --global url."https://github.com/".insteadOf ssh://git@github.com/
-            - run: npm i
+            - run: npm i --force
             - name: Cloning supertokens-root
               run: cd ../ && git clone https://github.com/supertokens/supertokens-root.git
             - name: Update Java 1
--- a/coreDriverInterfaceSupported.json
+++ b/coreDriverInterfaceSupported.json
@@ -1,4 +1,4 @@
 {
     "_comment": "contains a list of core-driver interfaces branch names that this core supports",
-    "versions": ["2.8", "2.9", "2.10", "2.11", "2.12", "2.13"]
+    "versions": ["2.8", "2.9", "2.10", "2.11", "2.12", "2.13", "2.14"]
 }
--- a/recipe/userroles/index.js
+++ b/recipe/userroles/index.js
@@ -1,0 +1,6 @@
+"use strict";
+function __export(m) {
+    for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
+}
+exports.__esModule = true;
+__export(require("../../lib/build/recipe/userroles"));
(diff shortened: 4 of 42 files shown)


.github/workflows/pre-commit-hook-run.yml
```diff
@@ -20,5 +20,5 @@ jobs:
               with:
                   node-version: "16"
             - run: git init && git add --all && git -c user.name='test' -c user.email='test@example.com' commit -m 'init for pr action'
-            - run: npm i
+            - run: npm i --force
```
rishabhpoddar, line 23: this change is already in branch 9.2. Also, since this is nothing to do with this PR, it should not be here in the first place.

.github/workflows/tests.yml
```diff
@@ -40,7 +40,7 @@ jobs:
                   node-version: "16"
             - name: Make git use https instead of ssh
               run: git config --global url."https://github.com/".insteadOf ssh://git@github.com/
-            - run: npm i
+            - run: npm i --force
```
rishabhpoddar, line 43: this change is already in branch 9.2. Also, since this is nothing to do with this PR, it should not be here in the first place.

coreDriverInterfaceSupported.json
```diff
@@ -1,4 +1,4 @@
 {
     "_comment": "contains a list of core-driver interfaces branch names that this core supports",
-    "versions": ["2.8", "2.9", "2.10", "2.11", "2.12", "2.13"]
+    "versions": ["2.8", "2.9", "2.10", "2.11", "2.12", "2.13", "2.14"]
```
rishabhpoddar, line 3: should mention this in changelog

test/utils.js
```diff
@@ -489,3 +489,20 @@ module.exports.generateRandomCode = function (size) {
...
     await new Promise((r) => setTimeout(r, time * 1000));
 };
+
+module.exports.areArraysEqual = function (arr1, arr2) {
+    if (arr1.length !== arr2.length) {
+        return false;
+    }
+
+    arr1.sort();
+    arr2.sort();
```
rishabhpoddar, line 499: I think this would only work for arrays that have primitive types. So rename this function to something like `arePrimitiveTypeArraysEqual`

rishabhpoddar (changes requested):

recipe/userroles/index.js
```diff
@@ -0,0 +1,6 @@
+"use strict";
```
rishabhpoddar, line 1: Need to change typescript playground as well

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #340, opened 2022-06-28
fix!: makes changes to the session recipe functions

rishabhpoddar:
## Summary of change

Makes changes to the session recipe for not throwing unauthorised error from functions that accept a session handle / access token and may not be working with the current session.

## Related issues

-   https://github.com/supertokens/backend/issues/83

## Test Plan

TODO

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Tests

### DIFF
--- a/lib/ts/recipe/session/sessionClass.ts
+++ b/lib/ts/recipe/session/sessionClass.ts
@@ -54,45 +54,33 @@ export default class Session implements SessionContainerInterface {
     };
 
     getSessionData = async (userContext?: any): Promise<any> => {
-        try {
-            let sessionInfo = await this.helpers.sessionRecipeImpl.getSessionInformation({
-                sessionHandle: this.sessionHandle,
-                userContext: userContext === undefined ? {} : userContext,
+        let sessionInfo = await this.helpers.sessionRecipeImpl.getSessionInformation({
+            sessionHandle: this.sessionHandle,
+            userContext: userContext === undefined ? {} : userContext,
+        });
+        if (sessionInfo === undefined) {
+            clearSessionFromCookie(this.helpers.config, this.res);
+            throw new STError({
+                message: "Session does not exist anymore",
+                type: STError.UNAUTHORISED,
             });
-            if (sessionInfo === undefined) {
-                throw new STError({
-                    message: "Session information does not exist.",
-                    type: STError.UNAUTHORISED,
-                });
-            }
-            return sessionInfo.sessionData;
-        } catch (err) {
-            if (err.type === STError.UNAUTHORISED) {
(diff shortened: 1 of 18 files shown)


lib/ts/recipe/session/sessionClass.ts
```diff
@@ -71,11 +76,18 @@ export default class Session implements SessionContainerInterface {
...
-            });
+            if (
+                !(await this.helpers.sessionRecipeImpl.updateSessionData({
+                    sessionHandle: this.sessionHandle,
+                    newSessionData,
+                    userContext: userContext === undefined ? {} : userContext,
+                }))
+            ) {
+                throw new STError({
+                    message: "Session does not exist anymore.",
```
porcellus, line 87: This error message is different from the rest. Also, I think this is a better error message, since the lifetime of the "session information" is the same as the lifetime of the entire session, right?

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -55,12 +55,17 @@ export default class Session implements SessionContainerInterface {
...
-                    sessionHandle: this.sessionHandle,
-                    userContext: userContext === undefined ? {} : userContext,
-                })
-            ).sessionData;
+            let sessionInfo = await this.helpers.sessionRecipeImpl.getSessionInformation({
+                sessionHandle: this.sessionHandle,
+                userContext: userContext === undefined ? {} : userContext,
+            });
+            if (sessionInfo === undefined) {
+                throw new STError({
```
porcellus, line 63: This is caught a few lines below, I'd prefer if we just did the cleanup here without a try-catch + rethrow. This also applies to: `updateSessionData`, `updateAccessTokenPayload`, `getTimeCreated`, and `getExpiry`.

porcellus (changes requested):

porcellus:
This isn't really an issue, but I don't think this fits the "fix" in the PR title it's more of an interface change than a bugfix.

lib/ts/recipe/session/sessionClass.ts
rishabhpoddar: Fair point. Done.

rishabhpoddar:
> This isn't really an issue, but I don't think this fits the "fix" in the PR title it's more of an interface change than a bugfix.

Well, actually this issue was discovered cause an issue which I would call a bug in design -> when looping through all sessions of a user, and whilst looping, if one of them expires (edge case, but it happened!), and then if the `updateAccessTokenPayload` was called on that session handle, it would throw an unauthorised error, logging out the current session.

porcellus (approved):

PR #278, opened 2022-03-19
feat(sessiongrants): add support for session grants

porcellus:
## Summary of change

Add support for session grants.

## Related issues

- supertokens/core-driver-interface#36
- supertokens/supertokens-core#390

## Test Plan

TODO

## Documentation changes

TODO

## Checklist for important updates

-   [ ] Changelog has been updated
-   [ ] `coreDriverInterfaceSupported.json` file has been updated (if needed)
    -   Along with the associated array in `lib/ts/version.ts`
-   [ ] `frontendDriverInterfaceSupported.json` file has been updated (if needed)
-   [ ] Changes to the version if needed
    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [ ] Had run `npm run build-pretty`
-   [ ] Had installed and ran the pre-commit hook
-   [ ] Issue this PR against the latest non released version branch.
    -   To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.
    -   If no such branch exists, then create one from the latest released branch.
-   [ ] If have added a new web framework, update the `add-ts-no-check.js` file to include that
-   [ ] If added a new recipe / api interface, then make sure that the implementation of it uses NON arrow functions only (like `someFunc: function () {..}`).
-   [ ] If added a new recipe, then make sure to expose it inside the recipe folder present in the root of this repo. We also need to expose its types.

## Remaining TODOs for this PR

- [ ] Finalize name for grant
- [ ] Check inteface

PR #350, opened 2022-07-09
feat: review feedback

porcellus:
## Summary of change

Fixing review feedback on linked PR.

## Related issues

-   https://github.com/supertokens/supertokens-node/pull/278

## Test Plan

Added a few tests

## Documentation changes

Done separately

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/session/claimBaseClasses/primitiveClaim.ts
```diff
@@ -27,6 +27,15 @@ export class PrimitiveClaim<T extends JSONValue> extends SessionClaim<T> {
...
     }
 
+    removeFromPayload(payload: any, _userContext?: any): any {
+        const res = {
+            ...payload,
+        };
+        delete res[this.key];
+
+        return res;
+    }
```
rishabhpoddar, line 37: if this is not being called anywhere, or if you don't think this is useful for the user either, maybe we should remove this function?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -463,12 +486,28 @@ export default function getRecipeInterface(querier: Querier, config: TypeNormali
...
+            this: RecipeInterface,
+            input: { sessionHandle: string; claim: SessionClaim<T>; userContext?: any }
+        ) {
+            const sessionInfo = await this.getSessionInformation({
+                sessionHandle: input.sessionHandle,
+                userContext: input.userContext,
+            });
+
+            if (sessionInfo === undefined) {
+                throw new Error("Session does not exist");
```
rishabhpoddar, line 499: Maybe we should return `undefined` here?

rishabhpoddar (changes requested):

PR #386, opened 2022-09-14
fix: backwards compatibility

porcellus:
## Summary of change

- updated auth-react test server
- added a fallback mechanism if email verification is called with the wrong `rid`

## Test Plan

Added a testcase 

## Documentation changes

N/A

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


porcellus:
We've decided to remove support for 1.14

PR #389, opened 2022-09-14
feat: clearing session for deleted users

porcellus:
## Summary of change

- Email verification endpoint now clears the session if called with a deleted/unknown user
- Sign out now always clears cookies - even if there is no valid session attached

## Related issues

-   supertokens/supertokens-auth-react/issues/580

## Test Plan

Added tests

## Documentation changes

N/A

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/emailverification/api/implementation.ts
```diff
@@ -17,7 +18,17 @@ export default function getAPIInterface(): APIInterface {
             const res = await options.recipeImplementation.verifyEmailUsingToken({ token, userContext });
 
             if (res.status === "OK" && session !== undefined) {
-                await session.fetchAndSetClaim(EmailVerificationClaim, userContext);
+                try {
+                    await session.fetchAndSetClaim(EmailVerificationClaim, userContext);
+                } catch (err) {
+                    // This should never happen, since we've just set the status above.
```
rishabhpoddar, line 24: what if a session does not exist? Then what affect will throwing this error have?

lib/ts/recipe/session/api/implementation.ts
```diff
@@ -82,6 +84,9 @@ export default function getAPIInterface(): APIInterface {
             if (session !== undefined) {
                 await session.revokeSession(userContext);
             }
+
+            clearSessionFromCookie(options.config, options.res);
```
rishabhpoddar, line 88: is this needed?

rishabhpoddar (changes requested):

lib/ts/recipe/emailverification/api/implementation.ts
porcellus: Then we can't have a `UNKOWN_USER_ID` thrown since that's only from `fetchAndSetClaim`.
In the case of a deleted user, it'll return `EMAIL_VERIFICATION_INVALID_TOKEN_ERROR` if the token has already been revoked, otherwise, it'll succeed.

lib/ts/recipe/session/api/implementation.ts
porcellus: Not strictly necessary, but since a user has been complaining about `signOutPost` not revoking their session in some cases (that I couldn't replicate), I thought it made sense just to go ahead and always clear the cookies even if we couldn't make sense of them.

test/sessionExpress.test.js
```diff
@@ -174,59 +174,6 @@ describe(`sessionExpress: ${printPath("[test/sessionExpress.test.js]")}`, functi
         assert(res2.status === 404);
     });
 
-    it("test that the sign out API clears cookies without session", async function () {
```
rishabhpoddar: this test is still useful to have.

rishabhpoddar (changes requested):

test/sessionExpress.test.js
porcellus: It'd fail. Should I re-add it with testing that we do not clear cookies?

PR #401, opened 2022-10-07
feat(sessionrework): remove id refresh token

porcellus:
## Summary of change

- Removed id-refresh-token
- Setting access token with unlimited cookie lifetime
- Sending "remove" for front-token to signal we are clearing the session 

## Related issues

-   

## Test Plan

We have extensive tests to make sure sessions work.

## Documentation changes

N/A internal change

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/recipe/session/cookieAndHeaders.ts
+++ b/lib/ts/recipe/session/cookieAndHeaders.ts
@@ -137,7 +137,6 @@ export function setToken(
     }
 
     if (transferMethod === "cookie") {
-        // We intentionally use accessTokenPath for idRefresh tokens
         setCookie(
             config,
             res,
--- a/lib/ts/recipe/session/cookieAndHeaders.ts
+++ b/lib/ts/recipe/session/cookieAndHeaders.ts
@@ -14,11 +14,11 @@
  */
 import { HEADER_RID } from "../../constants";
 import { BaseRequest, BaseResponse } from "../../framework";
-import { TokenType, TypeNormalisedInput } from "./types";
+import { TokenTransferMethod, TokenType, TypeNormalisedInput } from "./types";
 
 const authorizationHeaderKey = "authorization";
 const accessTokenCookieKey = "sAccessToken";
-const accessTokenheaderKey = "st-access-token";
+const accessTokenHeaderKey = "st-access-token";
 const refreshTokenCookieKey = "sRefreshToken";
 const refreshTokenHeaderKey = "st-refresh-token";
 
@@ -26,22 +26,18 @@ const antiCsrfHeaderKey = "anti-csrf";
 
 const frontTokenHeaderKey = "front-token";
 
+const authModeHeaderKey = "st-auth-mode";
+
 /**
  * @description clears all the auth cookies from the response
  */
-export function clearSession(config: TypeNormalisedInput, req: BaseRequest, res: BaseResponse, userContext: any) {
-    const transferMethod = config.getTokenTransferMethod({ req, userContext });
-
+export function clearSession(config: TypeNormalisedInput, res: BaseResponse, transferMethod: TokenTransferMethod) {
+    // If we can tell it's a cookie based session we are not clearing using headers
     const tokenTypes: TokenType[] = ["access", "refresh"];
     for (const token of tokenTypes) {
-        setToken(config, req, res, token, "", 0, userContext, transferMethod);
-
-        // This is to ensure we clear the cookies as well if the user has migrated to headers,
-        // because this can't be done on the client side
-        if (transferMethod === "header" && isTokenInCookies(req, token)) {
-            setToken(config, req, res, token, "", 0, userContext, "cookie");
-        }
+        setToken(config, res, token, "", 0, transferMethod);
     }
+
     res.setHeader(frontTokenHeaderKey, "remove", false);
     res.setHeader("Access-Control-Expose-Headers", frontTokenHeaderKey, true);
 }
@@ -66,7 +62,7 @@ export function setFrontTokenInHeaders(res: BaseResponse, userId: string, atExpi
 }
 
 export function getCORSAllowedHeaders(): string[] {
-    return [antiCsrfHeaderKey, HEADER_RID, authorizationHeaderKey, refreshTokenHeaderKey];
+    return [antiCsrfHeaderKey, HEADER_RID, authorizationHeaderKey, authModeHeaderKey];
 }
 
 function getCookieNameFromTokenType(tokenType: TokenType) {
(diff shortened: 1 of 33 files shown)


lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -167,26 +167,26 @@ export default function getRecipeInterface(
             let doAntiCsrfCheck = options !== undefined ? options.antiCsrfCheck : undefined;
             const transferMethod = config.getTokenTransferMethod({ req, userContext });
```
rishabhpoddar, line 168: getSession should not check token transfer method - as per our last discussion.

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -262,7 +261,11 @@ export default function getRecipeInterface(
                         res,
                         "access",
                         response.accessToken.token,
-                        response.accessToken.expiry,
+                        // We set the expiration to 10 years, because we can't really access the expiration of the refresh token everywhere we are setting it.
+                        // This should be safe to do, since this is only the validity of the cookie (set here or on the frontend) but we check the expiration of the JWT anyway.
+                        // Even if the token is expired the presence of the token indicates that the user could have a valid refresh
+                        // Setting them to infinity would require special case handling on the frontend and just adding 10 years seems enough.
+                        Date.now() + 315360000000,
```
rishabhpoddar, line 268: We should set this to the same as the refresh token expiry time instead? Just more clean

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -219,7 +219,11 @@ export default class Session implements SessionContainerInterface {
                 this.res,
                 "access",
                 response.accessToken.token,
-                response.accessToken.expiry,
+                // We set the expiration to 10 years, because we can't really access the expiration of the refresh token everywhere we are setting it.
+                // This should be safe to do, since this is only the validity of the cookie (set here or on the frontend) but we check the expiration of the JWT anyway.
+                // Even if the token is expired the presence of the token indicates that the user could have a valid refresh
+                // Setting them to infinity would require special case handling on the frontend and just adding 10 years seems enough.
+                Date.now() + 315360000000,
```
rishabhpoddar, line 226: We should set this to the refresh token's expiry time?

lib/ts/recipe/session/utils.ts
```diff
@@ -266,11 +266,21 @@ export function attachCreateOrRefreshSessionResponseToExpressRes(
...
+        config,
+        req,
+        res,
+        "access",
+        accessToken.token,
+        // We set the expiration to 10 years, because we can't really access the expiration of the refresh token everywhere we are setting it.
+        // This should be safe to do, since this is only the validity of the cookie (set here or on the frontend) but we check the expiration of the JWT anyway.
+        // Even if the token is expired the presence of the token indicates that the user could have a valid refresh
+        // Setting them to infinity would require special case handling on the frontend and just adding 10 years seems enough.
+        Date.now() + 315360000000,
```
rishabhpoddar, line 280: set it to the same as the refresh token time?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: @porcellus missed this?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -262,7 +261,11 @@ export default function getRecipeInterface(
                         res,
                         "access",
                         response.accessToken.token,
-                        response.accessToken.expiry,
+                        // We set the expiration to 100 years, because we can't really access the expiration of the refresh token everywhere we are setting it.
+                        // This should be safe to do, since this is only the validity of the cookie (set here or on the frontend) but we check the expiration of the JWT anyway.
+                        // Even if the token is expired the presence of the token indicates that the user could have a valid refresh
+                        // Setting them to infinity would require special case handling on the frontend and just adding 10 years seems enough.
+                        Date.now() + 3153600000000,
```
rishabhpoddar, line 268: in the decision log, we wrote to change it to 100 years, but we didn't write why. The comment above says that we don't have access to the refresh token's lifetime all the time - but that's not true since the session information has the expiry time.

lib/ts/utils.ts
```diff
@@ -5,6 +5,7 @@ import NormalisedURLDomain from "./normalisedURLDomain";
 import NormalisedURLPath from "./normalisedURLPath";
 import type { BaseRequest, BaseResponse } from "./framework";
 import { logDebugMessage } from "./logger";
+import { URL } from "url";
```
rishabhpoddar, line 8: unnecessary import?

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -152,17 +146,8 @@ export function setToken(
             expires,
             tokenType === "refresh" ? "refreshTokenPath" : "accessTokenPath"
         );
-
```
rishabhpoddar, line 155: there is a comment above "We intentionally use accessTokenPath for idRefresh tokens" -> remove that

lib/ts/recipe/session/types.ts
```diff
@@ -75,7 +75,7 @@ export interface ErrorHandlers {
     onInvalidClaim?: InvalidClaimErrorHandlerMiddleware;
 }
 
-export type TokenType = "access" | "refresh" | "idRefresh";
+export type TokenType = "access" | "refresh";
```
rishabhpoddar, line 78: there is a type below called `CreateOrRefreshAPIResponse` which has an IdrefreshToken section - remove that?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -167,26 +167,26 @@ export default function getRecipeInterface(
             let doAntiCsrfCheck = options !== undefined ? options.antiCsrfCheck : undefined;
             const transferMethod = config.getTokenTransferMethod({ req, userContext });
 
-            let idRefreshToken = getToken(config, req, "idRefresh", userContext, transferMethod);
```
rishabhpoddar, line 170: do we not need to check if the idrefreshtoken is there in the input during session refreshing to know if the old method was used or not? - is there nothing special that we need to do for migrating from older sessions to newer ones?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: We want to avoid querying the core, and that's why it's not available all the time.

rishabhpoddar (approved):

Issue #423, opened 2022-11-16
JWT payload doesn't get updated when calling mergeIntoAccessTokenPayload


rishabhpoddar:

rishabhpoddar:
This has been fixed in node sdk version >= 12.0.6

PR #426, opened 2022-11-16
fix: make mergeIntoAccessTokenPayload properly update the jwt payload

porcellus:
## Summary of change

- make mergeIntoAccessTokenPayload properly update the jwt payload
- use class methods instead of arrow functions defined on the instance

## Related issues

-   #423 

## Test Plan

Added test cases:
- `SessionContainer.mergeIntoAccessTokenPayload ` should update JWT
- `SessionContainer.fetchAndSetClaim` should update JWT when setting claim value
- `SessionContainer.setClaimValue` should update JWT when setting claim value
- `SessionContainer.removeClaim` should update JWT when removing claim
- `Session.mergeIntoAccessTokenPayload` updates JWT (passed originally)

## Documentation changes

N/A, fix change

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [12.0.6] - 2022-11-16
+
+### Fixed:
+
+-   Fixed mergeIntoAccessTokenPayload not updating the JWT payload
+
 ## [12.0.5] - 2022-10-17
 
 -   Updated google token endpoint.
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [12.0.7] - 2022-11-16
```
rishabhpoddar, line 10: 12.0.6 has not yet been released

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #434, opened 2022-11-20
feat(session rework): fallback auth method

porcellus:
## Summary of change

Implement fallback when auth method is not defined in the headers

## Related issues

- https://supertokens.com/docs/contribute/decisions/session/0003
- https://supertokens.com/docs/contribute/decisions/session/0007

## Test Plan

Added separately

## Documentation changes

Added separately

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/recipe/session/cookieAndHeaders.ts
+++ b/lib/ts/recipe/session/cookieAndHeaders.ts
@@ -18,7 +18,7 @@ import { TokenType, TypeNormalisedInput } from "./types";
 
 const authorizationHeaderKey = "authorization";
 const accessTokenCookieKey = "sAccessToken";
-const accessTokenheaderKey = "st-access-token";
+const accessTokenHeaderKey = "st-access-token";
 const refreshTokenCookieKey = "sRefreshToken";
 const refreshTokenHeaderKey = "st-refresh-token";
 
@@ -30,16 +30,21 @@ const frontTokenHeaderKey = "front-token";
  * @description clears all the auth cookies from the response
  */
 export function clearSession(config: TypeNormalisedInput, req: BaseRequest, res: BaseResponse, userContext: any) {
-    const transferMethod = config.getTokenTransferMethod({ req, userContext });
+    // If we can tell it's a cookie based session we are not clearing using headers
+    let transferMethod = config.getTokenTransferMethod({ req, userContext });
+
+    if (transferMethod === "MISSING_AUTH_HEADER") {
+        transferMethod = "header";
+    }
 
     const tokenTypes: TokenType[] = ["access", "refresh"];
     for (const token of tokenTypes) {
-        setToken(config, req, res, token, "", 0, userContext, transferMethod);
+        setToken(config, res, token, "", 0, transferMethod);
 
         // This is to ensure we clear the cookies as well if the user has migrated to headers,
         // because this can't be done on the client side
         if (transferMethod === "header" && isTokenInCookies(req, token)) {
-            setToken(config, req, res, token, "", 0, userContext, "cookie");
+            setToken(config, res, token, "", 0, "cookie");
         }
     }
     res.setHeader(frontTokenHeaderKey, "remove", false);
@@ -83,7 +88,9 @@ function getCookieNameFromTokenType(tokenType: TokenType) {
 function getHeaderNameFromTokenType(tokenType: TokenType) {
     switch (tokenType) {
         case "access":
-            return accessTokenheaderKey;
+            // We are getting the access token from the authorization header during verification.
+            // This case is handled in the getToken fn below.
+            return accessTokenHeaderKey;
         case "refresh":
             return refreshTokenHeaderKey;
         default:
@@ -95,17 +102,7 @@ export function isTokenInCookies(req: BaseRequest, tokenType: TokenType) {
     return req.getCookieValue(getCookieNameFromTokenType(tokenType)) !== undefined;
 }
 
-export function getToken(
-    config: TypeNormalisedInput,
-    req: BaseRequest,
-    tokenType: TokenType,
-    userContext: any,
-    transferMethod?: "cookie" | "header"
-) {
-    if (transferMethod === undefined) {
-        transferMethod = config.getTokenTransferMethod({ req, userContext });
-    }
-
+export function getToken(req: BaseRequest, tokenType: TokenType, transferMethod: "cookie" | "header") {
     if (transferMethod === "cookie") {
         return req.getCookieValue(getCookieNameFromTokenType(tokenType));
     } else if (transferMethod === "header") {
@@ -124,18 +121,12 @@ export function getToken(
 
 export function setToken(
     config: TypeNormalisedInput,
-    req: BaseRequest,
     res: BaseResponse,
     tokenType: TokenType,
     value: string,
     expires: number,
-    userContext: any,
-    transferMethod?: "cookie" | "header"
+    transferMethod: "cookie" | "header"
 ) {
-    if (transferMethod === undefined) {
-        transferMethod = config.getTokenTransferMethod({ req, userContext });
-    }
-
     if (transferMethod === "cookie") {
         // We intentionally use accessTokenPath for idRefresh tokens
         setCookie(
--- a/lib/ts/recipe/session/cookieAndHeaders.ts
+++ b/lib/ts/recipe/session/cookieAndHeaders.ts
@@ -14,7 +14,7 @@
  */
 import { HEADER_RID } from "../../constants";
 import { BaseRequest, BaseResponse } from "../../framework";
-import { TokenType, TypeNormalisedInput } from "./types";
+import { TokenTransferMethod, TokenType, TypeNormalisedInput } from "./types";
 
 const authorizationHeaderKey = "authorization";
 const accessTokenCookieKey = "sAccessToken";
@@ -26,24 +26,34 @@ const antiCsrfHeaderKey = "anti-csrf";
 
 const frontTokenHeaderKey = "front-token";
 
+const authModeHeaderKey = "st-auth-mode";
+
+export const availableTokenTransferMethods: TokenTransferMethod[] = ["cookie", "header"];
+
 /**
  * @description clears all the auth cookies from the response
  */
-export function clearSession(config: TypeNormalisedInput, req: BaseRequest, res: BaseResponse, userContext: any) {
+export function clearSession(
+    config: TypeNormalisedInput,
+    req: BaseRequest,
+    res: BaseResponse,
+    userContext: any,
+    transferMethod?: TokenTransferMethod | "missing_auth_header"
+) {
     // If we can tell it's a cookie based session we are not clearing using headers
-    let transferMethod = config.getTokenTransferMethod({ req, userContext });
+    let outputTransferMethod = transferMethod || config.getTokenTransferMethod({ req, userContext });
 
-    if (transferMethod === "MISSING_AUTH_HEADER") {
-        transferMethod = "header";
+    if (outputTransferMethod === "missing_auth_header") {
+        outputTransferMethod = "header";
     }
 
     const tokenTypes: TokenType[] = ["access", "refresh"];
     for (const token of tokenTypes) {
-        setToken(config, res, token, "", 0, transferMethod);
+        setToken(config, res, token, "", 0, outputTransferMethod);
 
         // This is to ensure we clear the cookies as well if the user has migrated to headers,
         // because this can't be done on the client side
-        if (transferMethod === "header" && isTokenInCookies(req, token)) {
+        if (outputTransferMethod === "header" && isTokenInCookies(req, token)) {
             setToken(config, res, token, "", 0, "cookie");
         }
     }
@@ -71,7 +81,7 @@ export function setFrontTokenInHeaders(res: BaseResponse, userId: string, atExpi
 }
 
 export function getCORSAllowedHeaders(): string[] {
-    return [antiCsrfHeaderKey, HEADER_RID, authorizationHeaderKey, refreshTokenHeaderKey];
+    return [antiCsrfHeaderKey, HEADER_RID, authorizationHeaderKey, refreshTokenHeaderKey, authModeHeaderKey];
 }
 
 function getCookieNameFromTokenType(tokenType: TokenType) {
@@ -85,7 +95,7 @@ function getCookieNameFromTokenType(tokenType: TokenType) {
     }
 }
 
-function getHeaderNameFromTokenType(tokenType: TokenType) {
+function getResponseHeaderNameForTokenType(tokenType: TokenType) {
     switch (tokenType) {
         case "access":
             // We are getting the access token from the authorization header during verification.
@@ -102,18 +112,16 @@ export function isTokenInCookies(req: BaseRequest, tokenType: TokenType) {
     return req.getCookieValue(getCookieNameFromTokenType(tokenType)) !== undefined;
 }
 
-export function getToken(req: BaseRequest, tokenType: TokenType, transferMethod: "cookie" | "header") {
+export function getToken(req: BaseRequest, tokenType: TokenType, transferMethod: TokenTransferMethod) {
     if (transferMethod === "cookie") {
         return req.getCookieValue(getCookieNameFromTokenType(tokenType));
     } else if (transferMethod === "header") {
-        if (tokenType === "access") {
-            const value = req.getHeaderValue(authorizationHeaderKey);
-            if (value === undefined || !value.startsWith("Bearer ")) {
-                return undefined;
-            }
-            return value.replace(/^Bearer /, "");
+        const value = req.getHeaderValue(authorizationHeaderKey);
+        if (value === undefined || !value.startsWith("Bearer ")) {
+            return undefined;
         }
-        return req.getHeaderValue(getHeaderNameFromTokenType(tokenType));
+
+        return value.replace(/^Bearer /, "");
     } else {
         throw new Error("Should never happen: Unknown transferMethod: " + transferMethod);
     }
@@ -125,7 +133,7 @@ export function setToken(
     tokenType: TokenType,
     value: string,
     expires: number,
-    transferMethod: "cookie" | "header"
+    transferMethod: TokenTransferMethod
 ) {
     if (transferMethod === "cookie") {
         setCookie(
@@ -137,7 +145,7 @@ export function setToken(
             tokenType === "refresh" ? "refreshTokenPath" : "accessTokenPath"
         );
     } else if (transferMethod === "header") {
-        setHeader(res, getHeaderNameFromTokenType(tokenType), value, expires);
+        setHeader(res, getResponseHeaderNameForTokenType(tokenType), value, expires);
     }
 }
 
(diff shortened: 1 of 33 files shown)


lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -30,16 +30,21 @@ const frontTokenHeaderKey = "front-token";
...
     for (const token of tokenTypes) {
-        setToken(config, req, res, token, "", 0, userContext, transferMethod);
+        setToken(config, res, token, "", 0, transferMethod);
 
         // This is to ensure we clear the cookies as well if the user has migrated to headers,
         // because this can't be done on the client side
         if (transferMethod === "header" && isTokenInCookies(req, token)) {
-            setToken(config, req, res, token, "", 0, userContext, "cookie");
+            setToken(config, res, token, "", 0, "cookie");
         }
```
rishabhpoddar, line 48: Can we not mix migration related stuff with normal code flow? I would ideally like migration stuff to be very clearly isolated in code with a very clear code path of:
if (is using older session) {
  do migration stuff
}

Let's discuss this on call

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -83,7 +88,9 @@ function getCookieNameFromTokenType(tokenType: TokenType) {
 function getHeaderNameFromTokenType(tokenType: TokenType) {
```
rishabhpoddar, line 88: should we call this function getResponseHeaderNameForTokenType?

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -95,17 +102,7 @@ export function isTokenInCookies(req: BaseRequest, tokenType: TokenType) {
...
-    transferMethod?: "cookie" | "header"
-) {
-    if (transferMethod === undefined) {
-        transferMethod = config.getTokenTransferMethod({ req, userContext });
-    }
-
+export function getToken(req: BaseRequest, tokenType: TokenType, transferMethod: "cookie" | "header") {
     if (transferMethod === "cookie") {
         return req.getCookieValue(getCookieNameFromTokenType(tokenType));
     } else if (transferMethod === "header") {
```
rishabhpoddar, line 108: won't refresh token also be via authorization header?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -162,38 +171,59 @@ export default function getRecipeInterface(
             userContext: any;
         }): Promise<Session | undefined> {
             logDebugMessage("getSession: Started");
-            logDebugMessage("getSession: rid in header: " + frontendHasInterceptor(req));
-            logDebugMessage("getSession: request method: " + req.getMethod());
+
+            const sessionOptional =
+                options !== undefined && typeof options !== "boolean" && options.sessionRequired === false;
```
rishabhpoddar, line 176: typeof options !== "boolean" will always be false?

lib/ts/recipe/session/utils.ts
```diff
@@ -326,6 +324,13 @@ export async function validateClaimsInPayload(
     return validationErrors;
 }
 
-function defaultGetTokenTransferMethod({ req }: { req: BaseRequest }): "cookie" | "header" {
-    return getAuthModeFromHeader(req) === "header" ? "header" : "cookie";
+function defaultGetTokenTransferMethod({ req }: { req: BaseRequest }): "cookie" | "header" | "MISSING_AUTH_HEADER" {
+    switch (getAuthModeFromHeader(req)) {
```
rishabhpoddar, line 328: why is getAuthModeFromHeader not defined in the session recipe? Is it used in ts/supertokens.ts file?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -162,38 +171,59 @@ export default function getRecipeInterface(
...
+                            type: STError.TRY_REFRESH_TOKEN,
+                        });
+                    }
+                }
+            }
+
+            if (accessToken === undefined) {
+                // This is done here to ensure smooth migration of sessions started before removing the id-refresh-token
+                // This token isn't handled by getToken to limit the scope of this legacy/migration code
+                if (req.getCookieValue("sIRTFrontend") !== undefined) {
```
rishabhpoddar, line 217: sIRTFrontend is http cookie attached to the websiteDomain and may not be sent in API domain.

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -162,38 +171,59 @@ export default function getRecipeInterface(
...
+            let transferMethod = preferredTransferMethod === "MISSING_AUTH_HEADER" ? "header" : preferredTransferMethod;
 
-            let accessToken = getToken(config, req, "access", userContext, transferMethod);
+            let accessToken = getToken(req, "access", transferMethod);
             if (accessToken === undefined) {
+                const fallbackMethod = transferMethod === "cookie" ? "header" : "cookie";
                 // We are checking if we could've gone ahead with validation if the transferMethod was different
                 // However, we don't want to do the fallback here, but force a call to refresh
                 // This is done to ensure that the browsers update the transfermethod in a timely manner (basically on the next API call)
                 // instead of waiting for the session to expire
```
rishabhpoddar, line 194: this comment is misleading. You are actually allowing fallback method in case of MISSING_AUTH_HEADER, but the comment says that "However, we don't want to do the fallback here, but force a call to refresh"

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -162,38 +171,59 @@ export default function getRecipeInterface(
...
+                            `getSession: falling back to use ${fallbackMethod} as an authentication-method`
+                        );
+                    } else {
+                        logDebugMessage(
+                            `getSession: returning TRY_REFRESH_TOKEN because preferred auth-mode doesn't allow fallback and access token was sent with the wrong method`
+                        );
+                        throw new STError({
+                            message: "access token sent with the wrong method. Please call the refresh API",
+                            type: STError.TRY_REFRESH_TOKEN,
+                        });
```
rishabhpoddar, line 209: Shouldn't we actually send this is the input access token is valid? What if that access token has expired, or it's an access token for another system etc.. we want to try refresh only if it's our access token otherwise we want to throw unauthorised (or return undefined in case of optional session verification).

This is useful in case of an API can be called by two types of clients - one that uses supertokens session and one that doesn't. In this case, the API would need to do getSession with optional true and if the auth mode is cookie based, but the input is an authorisation token, containing the other system's access token, then we want to let it pass.

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -202,36 +232,13 @@ export default function getRecipeInterface(
...
-                        message: "Access token has expired. Please call the refresh API",
-                        type: STError.TRY_REFRESH_TOKEN,
-                    });
-                }
-                return undefined;
-            }
+
             try {
                 let antiCsrfToken = getAntiCsrfTokenFromHeaders(req);
                 const disableAntiCSRF = transferMethod === "header";
```
rishabhpoddar, line 244: i know what `disableAntiCSRF` and `doAntiCsrfCheck` mean in the context of this code and that you are doing `!disableAntiCSRF && doAntiCsrfCheck,` below. 

But as a new reader of this code, it can be very confusing as to why you have two booleans that mean the same thing (semantically), but can have opposing values at the same time.

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -278,11 +284,16 @@ export default function getRecipeInterface(
...
-                    req
+                    req,
+                    transferMethod
                 );
 
                 return session;
             } catch (err) {
+                if (err.type === STError.TRY_REFRESH_TOKEN && sessionOptional && !gotRID && method !== "get") {
+                    return undefined;
+                }
```
rishabhpoddar, line 295: I don't quite understand this part. Session optional is a boolean that has effect when we get UNAUTHORISED error and not try refresh token.

rishabhpoddar: Also i don't understand why && !gotRID && method !== "get" matters.

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar, line 217: Since this block is related to migration, i will have a look at it later. I prefer if for now, all code is written without migration in mind and then we can add the migration code in isolation after - leads to easier code reviews.

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -382,24 +393,34 @@ export default function getRecipeInterface(
             logDebugMessage("refreshSession: Started");
             // We use a fallback mechanism here, to ensure there is a smooth upgrade path when switching transfer methods
```
rishabhpoddar, line 394: to make code review easier for this function, you should add a comment explaining all the possible inputs and the expected output like:
```
/*
Case 1:
input:
   - auth-method header: cookie
   - cookie: refreshtoken
   - authorisationheader: undefined
output:
   - new tokens in cookies

Case 2:
input:
   - auth-method header: header
   - cookie: refreshtoken
   - authorisationheader: undefined
output:
   - new tokens in headers 

Case 3:
input:
   - auth-method header: undefined
   - cookie: refreshtoken
   - authorisationheader: undefined
output:
   - new tokens in cookies

and so on.. (including what happens if both cookies and auth headers are present)
*/
```

lib/ts/recipe/session/types.ts
```diff
@@ -134,7 +129,10 @@ export type TypeNormalisedInput = {
     errorHandlers: NormalisedErrorHandlers;
     antiCsrf: "VIA_TOKEN" | "VIA_CUSTOM_HEADER" | "NONE";
 
-    getTokenTransferMethod: (input: { req: BaseRequest; userContext: any }) => "cookie" | "header";
+    getTokenTransferMethod: (input: {
+        req: BaseRequest;
+        userContext: any;
+    }) => "cookie" | "header" | "MISSING_AUTH_HEADER";
```
rishabhpoddar, line 135: should we make it all caps or all smallcase? Having smallcase and caps is strange - but you can keep it this if you want.

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -278,11 +284,16 @@ export default function getRecipeInterface(
...
 
                 return session;
             } catch (err) {
+                if (err.type === STError.TRY_REFRESH_TOKEN && sessionOptional && !gotRID && method !== "get") {
+                    return undefined;
+                }
+
                 if (err.type === STError.UNAUTHORISED) {
                     logDebugMessage("getSession: Clearing cookies because of UNAUTHORISED response");
                     clearSession(config, req, res, userContext);
```
rishabhpoddar, line 299: this might use a different method than what was used to verify

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar, line 295: remove this. Also add this to the ADR that we are removing this.

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar, line 217: move this up so in this API so that this is the first check that happens (easy code readability)

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -382,24 +393,34 @@ export default function getRecipeInterface(
             logDebugMessage("refreshSession: Started");
```
rishabhpoddar, line 393: you need to check for older session and clear the cookies from there if relevant

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: To check by @porcellus

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -124,23 +130,39 @@ export default function getRecipeInterface(
...
                 userId,
                 disableAntiCSRF,
                 accessTokenPayload,
                 sessionData
             );
-            attachCreateOrRefreshSessionResponseToExpressRes(config, req, res, response, userContext);
+
+            for (const transferMethod of availableTokenTransferMethods) {
+                if (transferMethod !== outputTransferMethod) {
+                    clearSession(config, req, res, userContext, transferMethod);
```
rishabhpoddar, line 152: this will add remove to the front-token - is that OK? I know that below we are calling `attachCreateOrRefreshSessionResponseToExpressRes` which should overwrite that `remove` value, but isn't this a bit strange of a code flow?

Im ok with this though - just thought to comment

rishabhpoddar, line 152: What if the existing session being cleared is the legacy session?

rishabhpoddar, line 152: I don't know if clearing sessions in createNewSession is something written in the ADR - maybe i forgot?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -162,38 +187,52 @@ export default function getRecipeInterface(
...
+                    } else {
+                        logDebugMessage(
+                            `getSession: returning TRY_REFRESH_TOKEN because preferred auth-mode doesn't allow fallback and access token was sent with the wrong method`
+                        );
+                        throw new STError({
+                            message: "access token sent with the wrong method. Please call the refresh API",
+                            type: STError.TRY_REFRESH_TOKEN,
+                        });
+                    }
                 }
```
rishabhpoddar, line 228: The code flow for this block is a little strange:
- If the `preferredTransferMethod !== missing_auth_header`, then we should actually keep accessToken as undefined so that the code below runs and the user either gets undefined or unauthorised error based on sessionoptional.
- If the `preferredTransferMethod !== missing_auth_header`, then we shouldn't even bother with querying `getToken(req, "access", fallbackMethod);`? So ideally the `if` condition should be `if (accessToken === undefined && preferredTransferMethod === "missing_auth_header") {`?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -278,7 +298,8 @@ export default function getRecipeInterface(
                     response.session.userId,
                     response.session.userDataInJWT,
                     res,
-                    req
+                    req,
+                    transferMethod
```
rishabhpoddar, line 302: there is a clearSession in the catch below - should that also take in the transferMethod? In fact, why not make clearSession always taken in a transferMethod? This way, errors like this can be avoided.

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -375,43 +396,69 @@ export default function getRecipeInterface(
...
-                        message:
-                            "Refresh token not found. Are you sending the refresh token in the request as a cookie?",
-                        type: STError.UNAUTHORISED,
-                    });
+            const refreshTokens: {
+                [key in TokenTransferMethod]?: string;
+            } = {};
+
+            // We check all token transfer methods for available refresh tokens
+            // We do this so that we can later clear all we are not overwrite
```
rishabhpoddar, line 419: "We do this so that we can later clear all we are not overwrite" -> what does this sentence mean?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -286,27 +303,27 @@ export default function getRecipeInterface(
...
-                    transferMethod
+                    requestTransferMethod
                 );
 
                 return session;
             } catch (err) {
                 if (err.type === STError.UNAUTHORISED) {
-                    logDebugMessage("getSession: Clearing cookies because of UNAUTHORISED response");
-                    clearSession(config, req, res, userContext);
+                    logDebugMessage("getSession: Clearing cookies because of UNAUTHORISED response from getSession");
```
rishabhpoddar, line 325: Change message

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
porcellus: I'll check and add one today, I'll be making an ADR PR anyway.

porcellus: 1. In most cases a refresh call is made before login, which clears/migrates the legacy session:
- doesSessionExist & getAccessTokenPayload securely is called by `SessionAuth`
- even in custom UIs, they are likely to be called (check the logged-in state and or/get a user id) before showing a sign-in screen
- if any call is made to an endpoint that uses `verifySession` it'll trigger a refresh.
2. Even if it's not, the first call to a verified endpoint will trigger a refresh that cleans the cookie of the old session
3. The old cookie is actually safe to ignore

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -124,23 +130,39 @@ export default function getRecipeInterface(
             sessionData?: any;
             userContext: any;
         }): Promise<Session> {
-            const disableAntiCSRF = config.getTokenTransferMethod({ req, userContext }) === "header";
```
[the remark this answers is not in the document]
porcellus: Since we are clearing the old sessions when creating a new one this case is also covered.

rishabhpoddar (approved):

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -146,7 +120,7 @@ export function setToken(
             tokenType === "refresh" ? "refreshTokenPath" : "accessTokenPath"
         );
     } else if (transferMethod === "header") {
-        setHeader(res, getHeaderNameFromTokenType(tokenType), value, expires);
+        setHeader(res, getResponseHeaderNameForTokenType(tokenType), value, expires);
```
rishabhpoddar, line 123: right below, the setHeader function has a `if (value === "remove") {` -> do we need that if block still?

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -66,7 +62,7 @@ export function setFrontTokenInHeaders(res: BaseResponse, userId: string, atExpi
 }
 
 export function getCORSAllowedHeaders(): string[] {
-    return [antiCsrfHeaderKey, HEADER_RID, authorizationHeaderKey, refreshTokenHeaderKey];
+    return [antiCsrfHeaderKey, HEADER_RID, authorizationHeaderKey, refreshTokenHeaderKey, authModeHeaderKey];
```
rishabhpoddar, line 65:
```suggestion
    return [antiCsrfHeaderKey, HEADER_RID, authorizationHeaderKey, authModeHeaderKey];
```

rishabhpoddar, line 65: Do we also need to add "Authorization" to this?

rishabhpoddar (changes requested):

lib/ts/recipe/session/cookieAndHeaders.ts
porcellus: `Access-Control-Allow-Credentials` covers that.

lib/ts/recipe/session/cookieAndHeaders.ts
porcellus: I guess we can remove it since we are not using that helper for the front-token

rishabhpoddar (approved):

PR #439, opened 2022-11-23
Recipe interface changes regarding account-linking

bhumilsarvaiya:
## Summary of change
- EP recipe interface changes regarding account-linking

### DIFF
--- a/lib/ts/recipe/emailpassword/index.ts
+++ b/lib/ts/recipe/emailpassword/index.ts
@@ -52,9 +52,10 @@ export default class Wrapper {
         });
     }
 
-    static createResetPasswordToken(userId: string, userContext?: any) {
+    static createResetPasswordToken(userId: string, email: string, userContext?: any) {
         return Recipe.getInstanceOrThrowError().recipeInterfaceImpl.createResetPasswordToken({
             userId,
+            email,
             userContext: userContext === undefined ? {} : userContext,
         });
     }
--- a/lib/ts/recipe/emailpassword/index.ts
+++ b/lib/ts/recipe/emailpassword/index.ts
@@ -52,6 +52,17 @@ export default class Wrapper {
         });
     }
 
+    /**
+     * We do not make email optional here cause we want to
+     * allow passing in primaryUserId. If we make email optional,
+     * and if the user provides a primaryUserId, then it may result in two problems:
+     *  - there is no recipeUserId = input primaryUserId, in this case,
+     *    this function will throw an error
+     *  - There is a recipe userId = input primaryUserId, but that recipe has no email,
+     *    or has wrong email compared to what the user wanted to generate a reset token for.
+     *
+     * And we want to allow primaryUserId being passed in.
+     */
     static createResetPasswordToken(userId: string, email: string, userContext?: any) {
         return Recipe.getInstanceOrThrowError().recipeInterfaceImpl.createResetPasswordToken({
             userId,
--- a/lib/ts/recipe/emailpassword/recipeImplementation.ts
+++ b/lib/ts/recipe/emailpassword/recipeImplementation.ts
@@ -99,11 +99,8 @@ export default function getRecipeInterface(querier: Querier): RecipeInterface {
         }): Promise<
             | {
                   status: "OK";
-                  /**
-                   * The id of the user whose password was reset.
-                   * Defined for Core versions 3.9 or later
-                   */
-                  userId?: string;
+                  userId: string;
+                  recipeUserId?: string;
               }
             | { status: "RESET_PASSWORD_INVALID_TOKEN_ERROR" }
         > {
--- a/lib/ts/recipe/emailpassword/recipeImplementation.ts
+++ b/lib/ts/recipe/emailpassword/recipeImplementation.ts
@@ -100,7 +100,7 @@ export default function getRecipeInterface(querier: Querier): RecipeInterface {
             | {
                   status: "OK";
                   userId: string;
-                  recipeUserId?: string;
+                  email: string;
               }
             | { status: "RESET_PASSWORD_INVALID_TOKEN_ERROR" }
         > {
--- a/lib/ts/recipe/emailpassword/types.ts
+++ b/lib/ts/recipe/emailpassword/types.ts
@@ -117,12 +117,9 @@ export type RecipeInterface = {
 
     createResetPasswordToken(input: {
         userId: string; // the id can be either recipeUserId or primaryUserId
+        email: string;
         userContext: any;
-    }): Promise<
-        | { status: "OK"; token: string }
-        | { status: "UNKNOWN_USER_ID_ERROR" }
-        | { status: "PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR" }
-    >;
+    }): Promise<{ status: "OK"; token: string } | { status: "UNKNOWN_USER_ID_ERROR" }>;
 
     resetPasswordUsingToken(input: {
         token: string;
@@ -131,6 +128,7 @@ export type RecipeInterface = {
     }): Promise<
         | {
               status: "OK";
+              email: string;
               userId: string;
           }
         | { status: "RESET_PASSWORD_INVALID_TOKEN_ERROR" }
@@ -142,11 +140,7 @@ export type RecipeInterface = {
         password?: string;
         userContext: any;
     }): Promise<{
-        status:
-            | "OK"
-            | "UNKNOWN_USER_ID_ERROR"
-            | "EMAIL_ALREADY_EXISTS_ERROR"
-            | "PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR";
+        status: "OK" | "UNKNOWN_USER_ID_ERROR" | "EMAIL_ALREADY_EXISTS_ERROR";
     }>;
 };
 
@@ -190,9 +184,9 @@ export type APIInterface = {
                     status: "OK";
                 }
               | {
-                    {status: "PASSWORD_RESET_NOT_ALLOWED", reason: string};
+                    status: "PASSWORD_RESET_NOT_ALLOWED";
+                    reason: string;
                 }
-              | { status: "PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR" }
               | GeneralErrorResponse
           >);
 
@@ -209,6 +203,7 @@ export type APIInterface = {
           }) => Promise<
               | {
                     status: "OK";
+                    email: string;
                     userId: string;
                 }
               | {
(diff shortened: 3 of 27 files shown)


lib/ts/recipe/emailpassword/types.ts
```diff
@@ -116,9 +116,13 @@ export type RecipeInterface = {
...
 
     createResetPasswordToken(input: {
-        userId: string;
+        userId: string; // the id can be either recipeUserId or primaryUserId
         userContext: any;
-    }): Promise<{ status: "OK"; token: string } | { status: "UNKNOWN_USER_ID_ERROR" }>;
+    }): Promise<
+        | { status: "OK"; token: string }
+        | { status: "UNKNOWN_USER_ID_ERROR" }
+        | { status: "PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR" }
```
rishabhpoddar, line 124: PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR -> makes no sense to me. Whats this for?

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -127,21 +131,23 @@ export type RecipeInterface = {
...
         email?: string;
         password?: string;
         userContext: any;
-    }): Promise<{ status: "OK" | "UNKNOWN_USER_ID_ERROR" | "EMAIL_ALREADY_EXISTS_ERROR" }>;
+    }): Promise<{
+        status:
+            | "OK"
+            | "UNKNOWN_USER_ID_ERROR"
+            | "EMAIL_ALREADY_EXISTS_ERROR"
+            | "PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR";
```
rishabhpoddar, line 149: PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR -> makes no sense to me

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -183,6 +189,10 @@ export type APIInterface = {
               | {
                     status: "OK";
                 }
+              | {
+                    status: "PASSWORD_RESET_NOT_ALLOWED_CONTACT_SUPPORT";
+                }
+              | { status: "PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR" }
```
rishabhpoddar, line 195: PROVIDE_RECIPE_USER_ID_AS_USER_ID_ERROR -> makes no sense to me

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -183,6 +189,10 @@ export type APIInterface = {
               | {
                     status: "OK";
                 }
+              | {
+                    status: "PASSWORD_RESET_NOT_ALLOWED_CONTACT_SUPPORT";
```
rishabhpoddar, line 193:
```suggestion
                    {status: "PASSWORD_RESET_NOT_ALLOWED", reason: string};
```

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -246,6 +256,34 @@ export type APIInterface = {
               | {
                     status: "EMAIL_ALREADY_EXISTS_ERROR";
                 }
+              | {
+                    status: "SIGNUP_NOT_ALLOWED";
```
rishabhpoddar, line 260:
```suggestion
                    status: "SIGNUP_NOT_ALLOWED", reason: string;
```

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -246,6 +256,34 @@ export type APIInterface = {
...
+                }
+              | GeneralErrorResponse
+          >);
+
+    postSignInAccountLinkingPOST:
```
rishabhpoddar, line 265:
```suggestion
    linkNewAccountToExistingAccountPOST:
```

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -246,6 +256,34 @@ export type APIInterface = {
...
+          }) => Promise<
+              | {
+                    status: "OK";
+                    user: User;
+                    session: SessionContainerInterface;
+                }
+              | {
+                    status: "ACCOUNT_LINK_FAILURE";
+                    reason: string;
+                    contactSupport: boolean;
```
rishabhpoddar, line 284: - contactSupport should not be a boolean.
- We should propagate all the reasons for account link failure to this function? Cause functions like linkAccounts return all of those status types?

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -254,6 +292,7 @@ export type TypeEmailPasswordPasswordResetEmailDeliveryInput = {
     type: "PASSWORD_RESET";
```
rishabhpoddar, line 292: the email delivery implementations should use the recipeUserId instead of user.id in its implementation

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -199,7 +209,7 @@ export type APIInterface = {
           }) => Promise<
               | {
                     status: "OK";
-                    userId?: string;
+                    userId: string;
```
rishabhpoddar, line 212: should this also return a recipeUserId?

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -127,21 +131,23 @@ export type RecipeInterface = {
     }): Promise<
         | {
               status: "OK";
-              /**
-               * The id of the user whose password was reset.
-               * Defined for Core versions 3.9 or later
-               */
-              userId?: string;
+              userId: string;
```
rishabhpoddar, line 134: should this also return a recipeUserId?

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -246,6 +256,34 @@ export type APIInterface = {
...
+              | {
+                    status: "OK";
+                    user: User;
+                    session: SessionContainerInterface;
+                }
+              | {
+                    status: "ACCOUNT_LINK_FAILURE";
+                    reason: string;
+                    contactSupport: boolean;
+                    recipeUserCreated: boolean;
```
rishabhpoddar, line 285: is this boolean needed?

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -246,6 +256,34 @@ export type APIInterface = {
               | {
                     status: "EMAIL_ALREADY_EXISTS_ERROR";
                 }
+              | {
```
rishabhpoddar, line 259: add a boolean in status: OK called `createdNewUser`. See https://github.com/supertokens/supertokens-core/issues/447#issuecomment-1298433805

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -116,9 +116,13 @@ export type RecipeInterface = {
     getUserByEmail(input: { email: string; userContext: any }): Promise<User | undefined>;
 
     createResetPasswordToken(input: {
-        userId: string;
+        userId: string; // the id can be either recipeUserId or primaryUserId
         userContext: any;
```
rishabhpoddar, line 120: need to take an email as well

lib/ts/recipe/emailpassword/types.ts
rishabhpoddar, line 134: also return an email

rishabhpoddar: TODO -> think about this.

lib/ts/recipe/emailpassword/recipeImplementation.ts
```diff
@@ -99,11 +99,8 @@ export default function getRecipeInterface(querier: Querier): RecipeInterface {
         }): Promise<
             | {
                   status: "OK";
-                  /**
-                   * The id of the user whose password was reset.
-                   * Defined for Core versions 3.9 or later
-                   */
-                  userId?: string;
+                  userId: string;
+                  recipeUserId?: string;
```
rishabhpoddar, line 103: make this return type consistent with the types file

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/types.ts
rishabhpoddar: what if we don’t change the user object on recipe level at all. So it has no recipeId and the id in that is always recipeId just like before

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -251,18 +246,21 @@ export type APIInterface = {
               | {
                     status: "OK";
                     user: User;
+                    createdNewUser: boolean;
+                    createdNewRecipeUser: boolean;
```
rishabhpoddar, line 250: this only needs createdNewUser since createdNewRecipeUser will always be true

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -241,11 +246,54 @@ export type APIInterface = {
...
+              }[];
+              session: SessionContainerInterface;
+              options: APIOptions;
+              userContext: any;
+          }) => Promise<
+              | {
+                    status: "OK";
+                    user: User;
+                    createdNewRecipeUser: boolean;
+                    session: SessionContainerInterface;
```
rishabhpoddar, line 277:
```suggestion
                    session: SessionContainerInterface;
                    wereAccountsAlreadyLinked: boolean
```

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -280,22 +279,20 @@ export type APIInterface = {
...
-              | {
-                    status: "EXISTING_ACCOUNT_NEEDS_TO_BE_VERIFIED_ERROR";
+                    description: string;
                 }
               | {
                     status: "ACCOUNT_LINKING_NOT_ALLOWED_ERROR";
-                }
-              | {
-                    status: "CANNOT_CREATE_PRIMARY_USER_FOR_EXISTING_ACCOUNT_ERROR";
+                    description: string;
```
rishabhpoddar, line 291: add description: string to all account linking related statuses

lib/ts/recipe/thirdpartyemailpassword/types.ts
```diff
@@ -251,7 +251,7 @@ export type APIInterface = {
                 }
           >);
 
-    emailPasswordLinkNewAccountToExistingAccountPOST:
+    emailPasswordLinkAccountToExistingAccountPOST:
```
rishabhpoddar, line 254:
```suggestion
    linkEmailPasswordAccountToExistingAccountPOST:
```

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/index.ts
```diff
@@ -52,9 +52,10 @@ export default class Wrapper {
         });
     }
 
-    static createResetPasswordToken(userId: string, userContext?: any) {
+    static createResetPasswordToken(userId: string, email: string, userContext?: any) {
```
rishabhpoddar, line 55: Add a comment here explaining that we do not make email optional here cause we want to allow passing in pf primary user ID. If we make email optional, and if the user provides a primary user ID, then it may result in two problems:
- there is no recipeUserId = input primary user ID, in this case, this function will throw an error
- There is a recipe userId = input primary user ID, but that recipe has no email, or has wrong email compared to what the user wanted to generate a reset token for.

And we want to allow primary user ID being passed in.

Also make an ADR for this.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #440, opened 2022-11-25
feat: add support for header based auth

porcellus:
## Summary of change

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -28,12 +28,30 @@ import { logDebugMessage } from "../../logger";
...
+                isAnIpAddress(helpers.appInfo.topLevelWebsiteDomain))
+        )
+    ) {
+        // We can allow insecure cookie when both website & API domain are localhost or an IP
+        // When either of them is a different domain, API domain needs to have https and a secure cookie to work
+        throw new Error(
+            "Since your API and website domain are different, for sessions to work, please use https on your apiDomain and dont set cookieSecure to false."
+        );
+    }
+
```
rishabhpoddar, line 54: the placement of this is strange. I think it should be in the recipeImplementation's createNewSession, and it should be called if transfer method is cookie. Right now, you are determining if this check should run based on the value of  `disableAntiCsrf` which is very indirect.

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -323,48 +419,111 @@ export default function getRecipeInterface(
...
-                // race condition mentioned here: https://github.com/supertokens/supertokens-node/issues/17
 
-                throw new STError({
-                    message: "Session does not exist. Are you sending the session tokens in the request as cookies?",
-                    type: STError.UNAUTHORISED,
-                });
+            // This token isn't handled by getToken/setToken to limit the scope of this legacy/migration code
+            if (req.getCookieValue(LEGACY_ID_REFRESH_TOKEN_COOKIE_NAME) !== undefined) {
+                setCookie(config, res, LEGACY_ID_REFRESH_TOKEN_COOKIE_NAME, "", 0, "accessTokenPath");
+            }
```
rishabhpoddar, line 435: should this only happen if we were successfully able to refresh the incoming session?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -139,104 +173,166 @@ export default function getRecipeInterface(
...
+            } catch (err) {
+                // We can clear the session from all transfer methods here, since we received a valid, non-expired token in the current method
+                if (err.type === STError.UNAUTHORISED) {
+                    logDebugMessage(
+                        `getSession: Clearing ${requestTransferMethod} because of UNAUTHORISED response from getSession`
+                    );
+                    clearSession(config, res, requestTransferMethod);
+                }
+                throw err;
+            }
```
rishabhpoddar, line 335: - why do we need to explicitly do this here? Won't the errorHandler just catch the `UNAUTHORISED` and clear the session as well?
- The comment says "We can clear the session from all transfer methods here". But when you call `clearSession`, you pass in the specific `requestTransferMethod`.
- Should't you check for if clearTokens is true in the error?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -323,48 +419,112 @@ export default function getRecipeInterface(
...
+            } catch (err) {
+                if (
+                    (err.type === STError.UNAUTHORISED && err.payload.clearTokens) ||
+                    err.type === STError.TOKEN_THEFT_DETECTED
+                ) {
+                    logDebugMessage(
+                        "refreshSession: Clearing tokens because of UNAUTHORISED or TOKEN_THEFT_DETECTED response"
+                    );
+                    clearSession(config, res, requestTransferMethod);
+                }
```
rishabhpoddar, line 525: won't the errorhandler itself clear the cookies here from the session by catching these error types? Why do we need this check here as well?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipeImplementation.ts
porcellus: I haven't updated these during the merge (that introduced the errors clearing everything). I'll just remove these sections.

lib/ts/recipe/session/recipeImplementation.ts
porcellus: I haven't updated these during the merge (that introduced the errors clearing everything). I'll just remove these sections.

lib/ts/recipe/session/recipeImplementation.ts
porcellus: There are 3 cases (involving the id refresh token):
- We get the id refresh token but no refresh token for some reason: we can remove it because future refreshes will fail  anyway
- We get the id refresh token and a refresh token, but the refresh fails for some reason:
  - anti-csrf check fails: we can keep it
  - core refresh call fails: we clear all tokens, so we should remove this as well
- We get the id refresh token, and the refresh token and the refresh succeeds: we should remove it because it was upgraded

I suppose I can move this around to avoid clearing if only the anti-csrf check fails.

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -12,94 +12,55 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
+import { HEADER_RID } from "../../constants";
 import { BaseRequest, BaseResponse } from "../../framework";
-import { TypeNormalisedInput } from "./types";
+import { availableTokenTransferMethods } from "./constants";
+import { TokenTransferMethod, TokenType, TypeNormalisedInput } from "./types";
 
+const authorizationHeaderKey = "authorization";
```
rishabhpoddar, line 20: Shouldn't this be capital "A"?

rishabhpoddar (changes requested):

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -111,7 +72,71 @@ export function setFrontTokenInHeaders(res: BaseResponse, userId: string, atExpi
...
+export function getToken(req: BaseRequest, tokenType: TokenType, transferMethod: TokenTransferMethod) {
+    if (transferMethod === "cookie") {
+        return req.getCookieValue(getCookieNameFromTokenType(tokenType));
+    } else if (transferMethod === "header") {
+        const value = req.getHeaderValue(authorizationHeaderKey);
+        if (value === undefined || !value.startsWith("Bearer ")) {
+            return undefined;
+        }
+
+        return value.replace(/^Bearer /, "");
```
rishabhpoddar, line 109: what if the user adds "Bearer  <token>" -> that is two spaces between "Bearer" and the token. We should also do trim to the result.

lib/ts/recipe/session/cookieAndHeaders.ts
porcellus: Header names are not case-sensitive according to the spec. 
Maybe we could add some processing internally to make sure, but for example, Axios always lowercases them.

test/utils.js
```diff
@@ -125,25 +126,33 @@ module.exports.extractInfoFromResponse = function (res) {
...
+        };
+    }
+    let accessTokenFromHeader = undefined;
+    if (res.headers["st-access-token"]) {
+        const [value, expiry] = res.headers["st-access-token"].split(";");
+        accessTokenFromHeader = {
+            value,
+            expiry: Number.parseInt(expiry),
+        };
+    }
```
KShivendu, line 147: I think this needs to be updated.

porcellus: You're right, I'll update it. Although the parsing still works since this will just parse it as NaN

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -8,28 +8,31 @@ import {
     SessionClaimValidator,
     SessionClaim,
```
rishabhpoddar, line 9: Please update change log in detail

rishabhpoddar (changes requested):

test/auth-modes.test.js
```diff
@@ -0,0 +1,1139 @@
...
+                                    getTokenTransferMethod: () => conf.getTokenTransferMethodRes,
+                                    antiCsrf: "VIA_TOKEN",
+                                }),
+                            ],
+                        });
+
+                        const app = getTestApp();
+
+                        const createInfo = await createSession(app, "cookie");
+                        await delay(3);
```
KShivendu, line 422: Should be avoided as it will keep on slowing tests.

porcellus: Sure, but AFAIK, we do not have a better way of making a token expire.

KShivendu: How about updating token content to make it expired.

PR #463, opened 2022-12-26
feat: Account linking interface update

bhumilsarvaiya:
## Summary of change
- interface updates regarding account linking

### DIFF
--- a/lib/ts/recipe/accountlinking/types.ts
+++ b/lib/ts/recipe/accountlinking/types.ts
@@ -72,6 +72,30 @@ export type TypeNormalisedInput = {
 export type APIInterface = {};
 
 export type RecipeInterface = {
+    getRecipeUserIdsForPrimaryUserIds: (input: {
+        primaryUserIds: string[];
+    }) => Promise<{
+        [primaryUserId: string]: string[]; // recipeUserIds
+    }>;
+    getPrimaryUserIdsforRecipeUserIds: (input: {
+        recipeUserIds: string[];
+    }) => Promise<{
+        [recipeUserId: string]: string | null;
+    }>;
+    addNewRecipeUserIdWithoutPrimaryUserId: (input: {
+        recipeUserId: string;
+        recipeId: string;
+        timeJoined: number;
+    }) => Promise<void>;
+    getUsers: (input: {
+        timeJoinedOrder: "ASC" | "DESC";
+        limit?: number;
+        paginationToken?: string;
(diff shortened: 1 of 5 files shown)


lib/ts/recipe/accountlinking/types.ts
```diff
@@ -65,13 +64,38 @@ export type TypeNormalisedInput = {
...
 };
 
-export type APIInterface = {};
-
 export type RecipeInterface = {
+    getRecipeUserIdsForPrimaryUserIds: (input: {
+        primaryUserIds: string[];
+        userContext: any;
+    }) => Promise<{
+        [primaryUserId: string]: string[]; // recipeUserIds
```
rishabhpoddar, line 75:
```suggestion
        [primaryUserId: string]: string[]; // recipeUserIds. If input primary user ID doesn't exists, those ids will not be part of the output set.
```

lib/ts/recipe/accountlinking/types.ts
```diff
@@ -65,13 +64,38 @@ export type TypeNormalisedInput = {
...
+    }>;
+    getPrimaryUserIdsforRecipeUserIds: (input: {
+        recipeUserIds: string[];
+        userContext: any;
+    }) => Promise<{
+        [recipeUserId: string]: string | null;
```
rishabhpoddar, line 81:
```suggestion
        [recipeUserId: string]: string | null; // if recipeUserId doesn't have a primaryUserId, then it will be mapped to `null`. If the input recipeUserId doesn't exist, then it won't be a part of the map
```

lib/ts/recipe/accountlinking/types.ts
```diff
@@ -65,13 +64,38 @@ export type TypeNormalisedInput = {
...
+    }>;
+    addNewRecipeUserIdWithoutPrimaryUserId: (input: {
+        recipeUserId: string;
+        recipeId: string;
+        timeJoined: number;
+        userContext: any;
+    }) => Promise<void>;
```
rishabhpoddar, line 88:
```suggestion
    }) => Promise<{
       status: "OK",
       createdNewEntry: boolean
    } | {
       status: "RECIPE_USER_ID_ALREADY_LINKED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR"
    }>;
```

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #471, opened 2023-01-04
feat: thirdparty re-work

sattvikc:
## Summary of change

## Test Plan
- [ ] Test `tenantId` with different valid and invalid values to ensure router does not break
- [ ] Test dashboard with `tenantId`
- [ ] Test dashboard API matches when using `tenantId`
- [ ] test for coreConfig in tenants-crud.
- [ ] add test for counting the calls of `makeDefaultUserContext` to ensure it's called from frameworks only.
- [ ] test that overriding getTenantId works
- [ ] test allowed domains claim
- [ ] test user roles claims that it contains roles / permissions according to the tenant
- [ ] test all email password APIs with tenant id
- [ ] test passwordless recipe functions with phoneNumber for multitenancy
- [ ] test passwordless recipe APIs for multitenancy
- [ ] test passwordless code magic link contains tenantId
- [ ] test resend code from different tenants

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

## Progress tracking
- [ ] FDI changes for multitenancy
- [ ] Router changes to accept optional `tenantId` - DONE (node)
- [ ] Multitenancy recipe implementation - DONE (node)
- [ ] create `userContext` in the frameworks and remove from API implementations - DONE (node)
- [ ] call `MultitenancyRecipe.getTenantId` in the router - DONE (node)
- [ ] update `fetchValue` in claims to accept tenantId & update allowedDomainsClaim to use it
- [ ] Update querier to take tenantId
- [ ] Update API interfaces and implementations for all recipes
  - [ ] dashboard
  - [ ] emailpassword
  - [ ] emailverification
  - [ ] jwt
  - [ ] multitenancy
  - [ ] openid
  - [ ] passwordless
  - [ ] session
  - [ ] thirdparty
  - [ ] thirdpartyemailpassword
  - [ ] thirdpartypasswordless
  - [ ] usermetadata
  - [ ] userroles
- [ ] tests for new thirdparty interface
- [ ] tests for multitenant
- [ ] tests for each recipe agains multitenancy

- [x] Check if any extra validations are required for gitlab and bitbucket providers (implement in SDK) while updating the SDK
- [x] We need to add tenantId to the reset password link as a query param in the backend SDK
- [x] We need to add tenantId to the email verification link as a query param in the backend SDK
- [x] We need to add tenantId to the password reset link as a query param in the backend SDK
- [x] backend sdk should protected prop check for tenantId in the accesstoken payload
- [x] Pass tenantId into sendEmail and sendSms functions on the backend SDK as well.
- [x] remove multitenancy claim validator in the backend SDKs
- [ ] Update `listThirdPartyConfigsForThirdPartyId` response in FDI
- [ ] Update `deleteTenant` response (`didExist`) in FDI
- [x] Fix create or update thirdparty config API in core
- [x] refresh session and few more APIs will not take tenantId from the URL (it should just ignore from the url and use session.getTenantId)
- [x] `getAllowedDomainsForTenantId` should return only string[]
- [x] `getAllowedDomainsForTenantId` takes `userContext`, check why it complained in docs
- [x] fix naming - export const AllowedDomainsClaim = new MultitenancyDomainsClaimClass(); (also rename the file)
- [x] if getAllowedDomainsForTenantId is defined and returns undefined, don't add the claim to the session'
- [x] st-tenant-domains - shorten this and inform Mihaly about the change, and ensure it's done in the frontend
- [x] IMPORTANT - make PR order and notes
- [x] check type of createOrUpdateTenant to ensure epEnabled, tpEnabled etc are optional - in interface and index
- [x] make the input and output type of getTenantId non optional
- [x] search for `_` param in the codebase to see if `tenantId` is being ignored in any of the place where it is required. especially session claims
- [x] user count per tenant needs to be added in the multitenancyStats in core
- [x] mulit tenancy usage stats and feature flag info should be obtained from the base storage and not the app level storage
- [x] Core: In the telemetry ID, we also need to sent the MAU count for that app since last month.
   ```
   path: /st/telemetry
   version: 3
   body: {
    superTokensVersion*: string,
    telemetryId*: string,
    mau*: number (since the last 30 days value),
    appId*: string,
    connectionURIDomain*: string (can be an empty string as well for root CUD)
   }
   ```
- [x] add getTenantId function to session object.
- [x] update email verfy get API in CDI to make it app specific
- [x] rawUserInfo -> fromIdTokenPayload and fromUserInfoAPI should be optional. check in golang if nil is not being expected
- [x] Wherever reading the tenant id from the session, we need to do session.getTenantId() and not `session.getAccessTokenPayload().tId`. Once such instance in in the email verification recipe generateEmailVerifyTokenPOST
- [x] tenantId should not be added to thirdPartyUserId
- [x] handle tenantId in dashboard (assigned empty string as of now)
- [x] Connection URI should allow to have path in it (for app id) - add test for this also. Check if we normalise the domain and that removes any path (connection to core)
- [x] update CDI spec for session APIs (booleans added to fetch / revoke from all tenants)
- [x] Check for edge function compatibility in node sdk (https://github.com/supertokens/supertokens-node/pull/599) - do at the end:
   - [x] no use of axios
   - [x] no use of jsonwebtoken lib
   - [x] no import from "url" lib
- [x] Dashboard related changes:
   - [x] The connection URI added to the dashboard bundle should include the app ID part as well. Need to make sure that on the frontend UI for creating a dashboard user, we show the app id path as well.
   - [x] Need APIs for dashboard recipe:
      - [x]  Getting a list of tenants for the app:
         ```
         path: /api/tenants/list GET
         return: {status: "OK", tenants: {tenantId: string, emailPassword: {enabled: boolean}, thirdParty: {enabled: boolean, providers: [...]}, passwordless:{enabled: boolean}, coreConfig: {..}}[]}
         ```  
   - [x] User object will now have the tenantIds array in it.
   - [x] APIs that will be specific to a tenant:
      - [x] userEmailVerifyTokenPost -> the frontend gives the first tenant (index 0) belonging to the user
      - [x] userEmailVerifyPut -> the frontend gives the first tenant (index 0) belonging to the user.
      - [x] userPasswordPut -> the frontend gives the first tenant which has emailPasswordEnabled: true.
      - [x] usersCountGet needs to be tenant specific (but not in the analytics API)
      - [x] usersGet needs to be tenant specific
      - [x] userPasswordPut
      - [x] userPut - the frontend gives the first tenant which has the correct recipe enabled, based on the type of "Auth Method" we display on the dashboard
   - [x] All requests from the frontend that will be per tenant will be like `{apiDomain}/{apiBasePath}/<tenantId>/api/...`
   - [x] functions like createEmailVerificationToken that are being called in userEmailVerifyPut should give what tenantId?
   - [x] in userSessionsGet and userSessionsPost, we will get for all tenants
   - [x] analyticsPost needs changing when querying the core?
   - [x] Type of Response in usersGet will change to have tenantIds: string[] in the user obj
   - [x] Type of CommonUserInformation needs to change to have tenantIds: string[]
- [x] Do functions like user pagination and getUserCount need to change in the backend SDK as well?
- [x] in sendEmailVerificationEmail, and similar functions in other recipes, take link that contains tenantId and also in the input body, which is a bad interface. we should not take link in the input. also add separate function for creation of link. But we keep the older functions
- [x] functions like `validate` in the formFields config also needs to take in a tenantId by the way (non optional in the func definition).
- [x] core change: when a new app is created, it should have all login methods enabled by default (update CDI spec description as well).
- [x] core change: Disallow setting of ip_whitelist and deny list for saas users
- [x] core change: argon2_parallelism, argon2_memory_kb, argon2_iterations should be config yaml only
- [x] core change: check if bcrypt_log_rounds should be config yaml only (need to decide if tenant level or core level)
- [x] core change: The hello withouth /hello should work for with appId and or tenantId as well.
- [x] All callbacks in all TypeInputs for all recipes should take a tenantId (non optional) unless there is a good reason for it to not take it.
- [x] Verify session and refresh session and all other APIs that take the access / refresh token as an input should return tenantId from the core in the `session` object part (also change CDI spec, but not version).
- [x] getSessionInformation function needs to return tenantId from the core.
- [x] Frontend SDK provider changes:
   - [x] optional thirdpartyid
   - [x] remove clientId from the input
   - [x] confirm that having two items in the privder list with same thirdPartyId throws error
- [x] Fix error message in core for 403 related tenant operation. The current message is `io.supertokens.multitenancy.exception.BadPermissionException: You must use the public tenantId and, public or same appId to add/update an app`
- [x] `io.supertokens.multitenancy.exception.BadPermissionException: Not allowed to modify DB related configs.` is coming when updating the public app config via the public app itself - change message
- [ ] core: license should not be required for all ee features if using inmemory database for better developer experience
- [x] update dashboard spec to include changes to dashboard APIs + new API to get list of all tenants
- [x] change the config.yaml file for removing the DIFFERENT_ACROSS_APPS for those argon2 configs
- [x] GET  /appid-<appId>/<tenantId>/recipe/multitenancy/tenant: is wrong. In the output, it doesn’t have coreConfig
- [x] make tenantid compulsory in index functions and see how it affect scripts in docs
- [x] There is an issue in the core related to user info fetching - it does it based on userID only and not appId, userId.. Specifically, getUsersInfoUsingIdList function in all the auth recipe queries - why does it not take an app ID? Cause otherwise it will not be using any index.. Also, in getUserInfoForRecipeIdFromUserIds query in the core, if a user ID is shared across 2 tenants, then its information will come twice.. Even in getTenantIdsForUserIds_transaction function.
- [x] issue with the core - if a test fails, subsequent tests are unable to start the core cause of address already in use.
- [x] in core, ensure the logs contain the right CUD. for eg. APIKeyUnauthorizedError
- [x] Fix test: testMultipleInstancesAtTheSameTime - the logging test also has the issue that if you run it more than once, it fails the second time, cause the it also picks up the logs from the previous runs.
- [x] getGloablClaimValidator - add tenantId
- [x] remove TenantDoesNotExistErrorHandlerMiddleware & RecipeDisabledForTenantErrorHandlerMiddleware and associated config
- [x] Change routing for Hapi to account for tenant id in the path
- [x] Update all framework tests to test for API paths with and without tenant id
- [x] core: for mysql, check if appId in query makes the deadlock test fail
- [x] core: for postgres, check why deadlock test is not failing, is it using the index?
- [x] make sure there is NO FDI that returns the core config.. whatsoever
- [x] What happens if you query `http://localhost:3567/appid-` to the core?
- [x] ensure there are no createAndSendCustomEmail & createAndSendCustomMessage anywhere - in lib and tests
- [x] update examples and ensure all of them are working
- [x] Update `/apiversion` API behaviour when default CDI version is set to return until that version
- [x] The recipe interface functions for multi tenancy, you need to return undefined in case a tenant does not exist when calling the getTenant function - and not throw a 402 error. Also make sure that wherever we call this in our existing code, it's handled correctly -> you should check for undefined and throw an error where appropriate. Also check this behaviour for other functions in the recipe impl as well.
- [x] https://github.com/supertokens/supertokens-node/issues/580#issuecomment-1635628827 - ensure this doesn't happen on the new interface
- [x] GET /<tenant-id>/recipe/mulltitenancy/tenant in the code with the older cdi and it gave a 404
- [x] Null pointer issue in core for email password user that belongs to no tenant
- [x] check for edge func compatibility

## PR breakups
1. Merge with latest branch - https://github.com/supertokens/supertokens-node/pull/590
2. Update router to handle optional tenantId - https://github.com/supertokens/supertokens-node/pull/592
3. Multitenancy recipe - https://github.com/supertokens/supertokens-node/pull/594, https://github.com/supertokens/supertokens-node/pull/596 & https://github.com/supertokens/supertokens-node/pull/597
4. User context refactor - https://github.com/supertokens/supertokens-node/pull/595
5. User roles recipe - https://github.com/supertokens/supertokens-node/pull/600 & https://github.com/supertokens/supertokens-node/pull/605
6. Email password recipe - https://github.com/supertokens/supertokens-node/pull/602
7. Other auth recipes - https://github.com/supertokens/supertokens-node/pull/606
8. Session recipe - https://github.com/supertokens/supertokens-node/pull/607
9. Dashboard recipe - https://github.com/supertokens/supertokens-node/pull/623
10. Email related - https://github.com/supertokens/supertokens-node/pull/625 & https://github.com/supertokens/supertokens-node/pull/626
11. tenantId in callback functions - https://github.com/supertokens/supertokens-node/pull/627 & https://github.com/supertokens/supertokens-node/pull/629
12. Dashboard router - https://github.com/supertokens/supertokens-node/pull/631
13. Test fixes - https://github.com/supertokens/supertokens-node/pull/630
14. Remove Error handlers in multitenancy recipe - https://github.com/supertokens/supertokens-node/pull/632
15. Add tenantId in `getGlobalClaimValidators` - https://github.com/supertokens/supertokens-node/pull/634
16. Fixes - https://github.com/supertokens/supertokens-node/pull/635 and https://github.com/supertokens/supertokens-node/pull/636
17. Dashboard fixes - https://github.com/supertokens/supertokens-node/pull/635 & https://github.com/supertokens/supertokens-node/pull/636
18. Handle tenant not found error & type fixes for getProvider - https://github.com/supertokens/supertokens-node/pull/639
19. Bitbucket and gitlab impl & other fixes - https://github.com/supertokens/supertokens-node/pull/642


lib/ts/recipe/thirdparty/api/implementation.ts
```diff
@@ -182,42 +84,21 @@ export default function getAPIInterface(): APIInterface {
...
-                "&code=" +
-                code;
-            options.res.sendHTMLResponse(
-                `<html><head><script>window.location.replace("${redirectURL}");</script></head></html>`
-            );
+        appleRedirectHandlerPOST: async function ({ formPostInfoFromProvider, options }): Promise<void> {
+            const queryString = qs.stringify(formPostInfoFromProvider);
+            const state = formPostInfoFromProvider.state;
+            const stateObj = JSON.parse(state);
+            const redirectURI = stateObj.redirectURI;
```
rishabhpoddar, line 96: base 64 decoded missed?

lib/ts/recipe/thirdparty/api/implementation.ts
```diff
@@ -182,42 +84,21 @@ export default function getAPIInterface(): APIInterface {
...
+            const redirectURL = `${redirectURI}?${queryString}`;
+
+            options.res.setHeader("Location", redirectURL, false);
+            options.res.setStatusCode(303);
+            options.res.sendHTMLResponse("");
```
rishabhpoddar, line 101: double check if this is required at all and if the content-type being set in the response is correct or not.

lib/ts/recipe/thirdparty/providers/utils.ts
```diff
@@ -62,19 +123,19 @@ export async function discoverOIDCEndpoints(config: ProviderConfigForClientType)
     if (config.oidcDiscoveryEndpoint !== undefined) {
         const oidcInfo = await getOIDCDiscoveryInfo(config.oidcDiscoveryEndpoint);
 
-        if (oidcInfo.authorisation_endpoint) {
+        if (oidcInfo.authorisation_endpoint && config.authorizationEndpoint === undefined) {
```
rishabhpoddar, line 126:
```suggestion
        if (oidcInfo.authorisation_endpoint !== undefined && config.authorizationEndpoint === undefined) {
```

rishabhpoddar: same applies to below as well. No implicit checks

lib/ts/recipe/thirdparty/types.ts
```diff
@@ -38,39 +38,46 @@ export type UserInfoMap = {
...
+    userInfoEndpointQueryParams?: { [key: string]: string };
+    userInfoEndpointHeaders?: { [key: string]: string };
     jwksURI?: string;
     oidcDiscoveryEndpoint?: string;
     userInfoMap?: UserInfoMap;
     validateIdTokenPayload?: (input: {
-        idTokenPayload: any;
+        idTokenPayload: { [key: string]: any };
         clientConfig: ProviderConfigForClientType;
+        userContext?: any;
```
rishabhpoddar, line 68: should not be optional since the user is never going to call this function. They are only ever going to implement the body of this function

lib/ts/recipe/thirdparty/types.ts
```diff
@@ -38,39 +38,46 @@ export type UserInfoMap = {
...
     }) => Promise<void>;
     requireEmail?: boolean;
-    generateFakeEmail?: (input: { thirdPartyUserId: string; userContext: any }) => string;
-    tenantId?: string;
+    generateFakeEmail?: (input: { thirdPartyUserId: string; userContext?: any }) => Promise<string>;
```
rishabhpoddar, line 71:
```suggestion
    generateFakeEmail?: (input: { thirdPartyUserId: string; userContext: any }) => Promise<string>;
```

lib/ts/recipe/thirdparty/utils.ts
```diff
@@ -34,9 +34,9 @@ export function validateAndNormaliseUserInput(appInfo: NormalisedAppinfo, config
 
 function validateAndNormaliseSignInAndUpConfig(
     _: NormalisedAppinfo,
-    config: TypeInputSignInAndUp
+    config: TypeInputSignInAndUp | undefined
 ): TypeNormalisedInputSignInAndUp {
-    let providers = config.providers;
+    let providers: ProviderInput[] = config?.providers || [];
```
rishabhpoddar, line 39: i do not like implicit checks.. please be explicit - leads to lesser bugs

rishabhpoddar (changes requested):

lib/ts/recipe/thirdparty/providers/configUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+    providers: ProviderInput[],
+    thirdPartyId: string,
+    clientType: string | undefined,
+    userContext: any
+): Promise<TypeProvider> {
+    for (const providerInput of providers) {
+        if (providerInput.config.thirdPartyId === thirdPartyId) {
+            let providerInstance = createProvider(providerInput);
+            await fetchAndSetConfig(providerInstance, clientType, userContext);
+            return providerInstance;
```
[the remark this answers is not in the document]
rishabhpoddar: @sattvikc is this checked?

lib/ts/recipe/thirdparty/providers/configUtils.ts
```diff
@@ -117,8 +97,8 @@ export function mergeConfig(staticConfig: ProviderConfig, coreConfig: ProviderCo
         },
     };
 
-    const mergedClients = [...staticConfig.clients];
-    for (const client of coreConfig.clients) {
+    const mergedClients = [...(staticConfig.clients || [])];
+    for (const client of coreConfig.clients || []) {
```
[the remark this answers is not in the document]
sattvikc: spread operator was not allowed on undefined type, I have made a more implicit assignment here.

lib/ts/recipe/thirdparty/providers/configUtils.ts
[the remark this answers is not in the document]
sattvikc: yes, here is the playground for that - https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gOwGYEsDmUC8UDeAoKQqAQwC4oBnYAJxQTQG4CiAjcq2+pgXzz1EhQAKuAgAFanABuKACYRqWXM0IBjRKjTl4ydEyJQkAVwSqAguQAUASiwA+KBLgBbFBQgAeYghB2efYgoQU0MTVWAUREMIYFUACzMEWQBlGMswSRl5anIRSCcshVtyJ1d3Dyk4OQd8AwzpOQUAOnVdNCaWJQByCGJqYDiuvF48QODVUNMIqNVqXuBxTMbqGygSyTLPPMWG7JqVKFaqKBRnMAAbJVqDIlbNcmubm7IoACI4iHPzuFeAGgOnoQ2G8AO5wajnWR-AFEbj-QGTczkMYhYxTSIIKCrR4IqBzYBGaiYnG4ohNcmnC4tDToGFPXgIhmwvgGYgg4goYDRWIJJKpYCWSnnawsoj4wmYoXDAJBVFhaaY5wchA2dYuNyeSrVZQGI5c+qFRTYNkcrmzeY7Q02fS3RAUODnCBNb5oSwmzlQA3LJpo8w2EUjJV0a14IA

lib/ts/recipe/thirdparty/api/implementation.ts
sattvikc: I tried a basic express app. Browser ignores the content-type and also the content, and redirect works as expected.

lib/ts/recipe/thirdparty/providers/utils.ts
sattvikc: Done

lib/ts/recipe/thirdparty/types.ts
sattvikc: Done

lib/ts/recipe/thirdparty/types.ts
sattvikc: Done

lib/ts/recipe/thirdparty/utils.ts
sattvikc: Done

lib/ts/recipe/thirdparty/providers/activeDirectory.ts
```diff
@@ -38,10 +38,12 @@ export default function ActiveDirectory(input: ProviderInput): TypeProvider {
             const config = await oGetConfig({ clientType, userContext });
 
             if (config.oidcDiscoveryEndpoint === undefined) {
-                config.oidcDiscoveryEndpoint = `https://login.microsoftonline.com/${config.additionalConfig.directoryId}/v2.0/`;
+                config.oidcDiscoveryEndpoint = `https://login.microsoftonline.com/${
+                    config.additionalConfig!.directoryId
```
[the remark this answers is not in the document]
sattvikc: Added checks.

lib/ts/recipe/thirdparty/providers/apple.ts
```diff
@@ -52,16 +52,16 @@ export default function Apple(input: ProviderInput): TypeProvider {
...
 
             if (config.clientSecret === undefined) {
                 config.clientSecret = getClientSecret(
                     config.clientID,
-                    config.additionalConfig.keyId,
-                    config.additionalConfig.teamId,
-                    config.additionalConfig.privateKey
+                    config.additionalConfig!.keyId,
+                    config.additionalConfig!.teamId,
+                    config.additionalConfig!.privateKey
```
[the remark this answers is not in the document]
sattvikc: Added checks.

lib/ts/recipe/thirdparty/providers/custom.ts
```diff
@@ -146,182 +136,193 @@ export default function NewProvider(input: ProviderInput): TypeProvider {
...
-        for (const client of input.config.clients) {
-            if (client.clientType === clientType) {
-                return getProviderConfigForClient(input.config, client);
+                return getProviderConfigForClient(input.config, input.config.clients[0]);
             }
-        }
 
-        throw new Error(`Could not find client config for clientType: ${clientType}`);
-    };
+            for (const client of input.config.clients || []) {
```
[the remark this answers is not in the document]
sattvikc: Done

rishabhpoddar (approved):

PR #473, opened 2023-01-09
fix: clear session fix

porcellus:
## Summary of change

- Updates changelog
- Updates session clearing to remove any session added to current request
- Updates session clearing to remove tokens even if only refresh-token is present

## Related issues

-   https://github.com/supertokens/supertokens-node/issues/476

## Test Plan

Added testcases:
- Throwing UNAUTHORISED during refresh with only refresh token present
- Throwing UNAUTHORISED after createNewSession

## Documentation changes

N/A

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/framework/express/framework.ts
+++ b/lib/ts/framework/express/framework.ts
@@ -29,7 +29,6 @@ import {
 import type { Framework } from "../types";
 import SuperTokens from "../../supertokens";
 import type { SessionContainerInterface } from "../../recipe/session/types";
-import { COOKIE_HEADER } from "../constants";
 
 export class ExpressRequest extends BaseRequest {
     private request: Request;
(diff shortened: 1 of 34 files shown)


lib/ts/framework/express/framework.ts
```diff
@@ -129,6 +134,24 @@ export class ExpressResponse extends BaseResponse {
         setCookieForServerResponse(this.response, key, value, domain, secure, httpOnly, expires, path, sameSite);
     };
 
+    clearCookie = (key: string) => {
```
rishabhpoddar, line 137: im not sure if this is the way to clear cookies.. clearing cookies requires setting the cookie to an expired date. If instead you meant removeSetCookieFromResponse, then you should change the name of this funciton perhaps.

rishabhpoddar (changes requested):

porcellus:
That's what I meant, `removeSetCookieFromResponse` just seemed like a very awkward name. I'll rename it when I start work.

Sent from ProtonMail mobile

-------- Original Message --------

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -134,6 +137,17 @@ export function setToken(
...
 
+function removeTokenUpdatesFromResponse(res: BaseResponse) {
+    res.removeHeader(antiCsrfHeaderKey);
+    res.removeHeader(frontTokenHeaderKey);
+
+    res.removeHeader(getResponseHeaderNameForTokenType("access"));
+    res.removeHeader(getResponseHeaderNameForTokenType("refresh"));
+
+    res.clearCookie(getCookieNameFromTokenType("access"));
+    res.clearCookie(getCookieNameFromTokenType("refresh"));
```
rishabhpoddar, line 148: do we really need to do this? It's a very weird function to have in the framework.. What happens if we don't do this?

rishabhpoddar (changes requested):

lib/ts/recipe/session/cookieAndHeaders.ts
porcellus: The worst case is shown in a test: if the user calls `createNewSession` then throws an UNAUTHORISED it will not clear the new session. This can be a use-case, for example, in a user banning scenario: checking the signed-in user in the `signInPOST` override.

PR #474, opened 2023-01-10
Fix awsLambda header typo

mihaiblaga89:
## Summary of change

Fixed typo. This would cause Content-Type to be plain on Lambda with Middy.

CleanShot 2023-01-10 at 02 08 33@2x


rishabhpoddar (approved):

rishabhpoddar:
Thanks @mihaiblaga89

PR #477, opened 2023-01-16
Add h3 framework

ColinEspinas:
## Summary of change

This PR aims to add compatibility to [h3](https://github.com/unjs/h3) (and by extension [Nitro](https://github.com/unjs/nitro)) to Supertokens back-end.

## Test Plan

So for now I am trying to get it to work with a simple nitro server with an empty "/" route by creating a plugin to init supertokens and adding the required middlewares to h3. This is still WIP and creating a real testing suite like it is done for [other frameworks](https://github.com/supertokens/supertokens-node/tree/master/test/framework) should be the way to go.

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/framework/h3/framework.ts
+++ b/lib/ts/framework/h3/framework.ts
@@ -98,13 +98,6 @@ export class H3Response extends BaseResponse {
 
     sendHTMLResponse = (html: string) => {
         if (this.event.node.res.writable) {
-            /**
-             * response.set method is not available if response
-             * is a nextjs response object. setHeader method
-             * is present on OutgoingMessage which is one of the
-             * bases used to construct response object for express
-             * like response as well as nextjs like response
-             */
             this.event.node.res.setHeader("Content-Type", "text/html");
             this.event.node.res.statusCode = this.statusCode;
             send(this.event, html);
--- a/lib/ts/framework/h3/framework.ts
+++ b/lib/ts/framework/h3/framework.ts
@@ -1,0 +1,212 @@
+/* Copyright (c) 2021, VRAI Labs and/or its affiliates. All rights reserved.
+ *
+ * This software is licensed under the Apache License, Version 2.0 (the
+ * "License") as published by the Apache Software Foundation.
+ *
+ * You may not use this file except in compliance with the License. You may
+ * obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+import {
+    eventHandler,
+    send,
+    setCookie,
+    createError,
+    EventHandler,
+    H3Event,
+    sendError,
+    readBody,
+    getQuery,
+    getCookie,
+    getHeader,
+    getMethod,
+
+} from "h3";
+import type { HTTPMethod } from "../../types";
+import { normaliseHttpMethod } from "../../utils";
+import { BaseRequest } from "../request";
+import { BaseResponse } from "../response";
+import type { Framework } from "../types";
+import SuperTokens from "../../supertokens";
+import type { SessionContainerInterface } from "../../recipe/session/types";
+
+export class H3Request extends BaseRequest {
+    private event: H3Event;
+
+    constructor(event: H3Event) {
+        super();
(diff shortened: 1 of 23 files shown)


lib/ts/framework/h3/framework.ts
```diff
@@ -0,0 +1,201 @@
...
+        session?: SessionContainerInterface;
+    };
+}
+
+export const middleware = eventHandler(async (event) => {
+    const supertokens = SuperTokens.getInstanceOrThrowError();
+    const request = new H3Request(event);
+    const response = new H3Response(event);
+    try {
+        await supertokens.middleware(request, response);
```
rishabhpoddar, line 176: h3 middlewares seem to be compatible with express' middleware. Therefore, I think you can just copy over the logic from the [express framework](https://github.com/supertokens/supertokens-node/blob/master/lib/ts/framework/express/framework.ts#L152) (but of course, use the `H3Request` and `H3Response` wrapper instead of `ExpressRequest` and `ExpressResponse`.

The key point here is that `await supertokens.middleware(request, response)` returns a boolean. If it's `false`, it means that supertokens didn't handle the request and you need to call the `next` function.

lib/ts/framework/h3/framework.ts
```diff
@@ -0,0 +1,201 @@
...
+    } catch (err) {
+        await supertokens.errorHandler(err, request, response);
+    }
+});
+
+export const errorHandler = eventHandler((event: H3Event) => {
+    const error = createError({});
+    sendError(event, error);
```
rishabhpoddar, line 184: I think you can copy over the code from the [express errorHandler](https://github.com/supertokens/supertokens-node/blob/master/lib/ts/framework/express/framework.ts#L176) cause the `errorHandler` too is a middleware that needs to be added to your `h3` app after all your routes.

lib/ts/recipe/session/framework/h3.ts
```diff
@@ -0,0 +1,34 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+import Session from "../recipe";
+import type { VerifySessionOptions } from "..";
+import type { SessionEvent } from "../../../framework/h3/framework";
+import { H3Request, H3Response } from "../../../framework/h3/framework";
+import SuperTokens from "../../../supertokens";
+import { eventHandler } from "h3";
```
rishabhpoddar, line 20:
```suggestion
import type { eventHandler } from "h3";
```

rishabhpoddar: We are only allowed to import `types` from frameworks cause our package.json has `h3` only in dev dependencies.

lib/ts/framework/h3/framework.ts
```diff
@@ -0,0 +1,201 @@
...
+    createError,
+    EventHandler,
+    H3Event,
+    sendError,
+    readBody,
+    getQuery,
+    getCookie,
+    getHeader,
+    getMethod,
+} from "h3";
```
rishabhpoddar, line 29: Our SDK cannot depend on these functions cause we can't add dependency on `h3` in our package.json (but we can in dev dependencies - which you already have). Is there a way in which these functions can be called directly on the H3Event object? For example `event.readBody()`

lib/ts/framework/h3/framework.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+export const middleware = eventHandler(async (event) => {
+    const supertokens = SuperTokens.getInstanceOrThrowError();
+    const request = new H3Request(event);
+    const response = new H3Response(event);
+    try {
+        await supertokens.middleware(request, response);
+    } catch (err) {
+        await supertokens.errorHandler(err, request, response);
```
rishabhpoddar, line 178: In the [discord logs](https://discord.com/channels/603466164219281420/995300294147711066/1064625720741085324) you had sent earlier, it seemed that the code execution came in the error handler (which is expected), and that this function wrote status code `401` to the `response` object, but you still got back a 404. This could possibly be because:
- Setting the status code on the `response` object and sending a reply doesn't actually send it to the client (maybe a bug in the `H3Response` implementation above)
- h3 routing may require you to define paths for all routes that can be called from the frontend, else it returns a `404`. This behaviour is similar to other frameworks like Hapi in which [we push all the routes dynamically as seen here](https://github.com/supertokens/supertokens-node/blob/master/lib/ts/framework/hapi/framework.ts#L232).

rishabhpoddar (changes requested):

PR #501, opened 2023-03-02
feat: Emailpassword accountlinking

bhumilsarvaiya:
## Summary of change
- emailpassword and session claim changes for account linking


lib/ts/recipe/emailpassword/types.ts
```diff
@@ -102,6 +94,7 @@ export type RecipeInterface = {
     signUp(input: {
         email: string;
         password: string;
+        doAccountLinking: boolean;
         userContext: any;
     }): Promise<{ status: "OK"; user: User } | { status: "EMAIL_ALREADY_EXISTS_ERROR" }>;
```
rishabhpoddar, line 99: signUp should also have createNewPrimaryUser boolean if status is OK

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -239,7 +623,26 @@ export default function getAPIImplementation(): APIInterface {
...
+                userContext,
+            });
+
+            if (!isSignUpAllowed) {
+                return {
+                    status: "SIGNUP_NOT_ALLOWED",
+                    reason: "the sign-up info is already associated with another account where it is not verified",
+                };
+            }
+            let response = await options.recipeImplementation.signUp({
```
rishabhpoddar, line 640: call should do account linking before this

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -151,7 +144,7 @@ export type RecipeInterface = {
         password?: string;
         userContext: any;
     }): Promise<{
-        status: "OK" | "UNKNOWN_USER_ID_ERROR" | "EMAIL_ALREADY_EXISTS_ERROR";
+        status: "OK" | "UNKNOWN_USER_ID_ERROR" | "EMAIL_ALREADY_EXISTS_ERROR" | "EMAIL_CHANGE_NOT_ALLOWED";
     }>;
 };
```
rishabhpoddar, line 149: in sign up post API, we need to rename createdNewUser to createdNewPrimaryUser

lib/ts/recipe/emailpassword/recipeImplementation.ts
```diff
@@ -1,21 +1,42 @@
...
 import NormalisedURLPath from "../../normalisedURLPath";
+import { getUser, listUsersByAccountInfo } from "../..";
+import EmailVerification from "../emailverification/recipe";
 
 export default function getRecipeInterface(querier: Querier): RecipeInterface {
     return {
         signUp: async function ({
             email,
             password,
+            doAccountLinking,
```
rishabhpoddar, line 13: i don't think you need doAccountLinking anymore

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -66,13 +66,6 @@ export type TypeNormalisedInputSignIn = {
     formFields: NormalisedFormField[];
 };
 
-export type TypeInputResetPasswordUsingTokenFeature = {
-    /**
-     * @deprecated Please use emailDelivery config instead
-     */
-    createAndSendCustomEmail?: (user: User, passwordResetURLWithToken: string, userContext: any) => Promise<void>;
-};
```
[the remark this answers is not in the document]
bhumilsarvaiya: The user object is changed here. The people who are using this config, will still need to update the logic to account for the updated User Object. Or we will have to keep an old copy of User types and pass it here. Better we just remove it all together

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -102,6 +94,7 @@ export type RecipeInterface = {
     signUp(input: {
         email: string;
         password: string;
+        doAccountLinking: boolean;
         userContext: any;
     }): Promise<{ status: "OK"; user: User } | { status: "EMAIL_ALREADY_EXISTS_ERROR" }>;
```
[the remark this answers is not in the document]
bhumilsarvaiya: Not sure. We can just mark this deprecated ?

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -151,7 +144,7 @@ export type RecipeInterface = {
         password?: string;
         userContext: any;
     }): Promise<{
-        status: "OK" | "UNKNOWN_USER_ID_ERROR" | "EMAIL_ALREADY_EXISTS_ERROR";
```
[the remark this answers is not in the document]
bhumilsarvaiya: the reasoning is very clear if the diagram is referred

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -151,7 +144,7 @@ export type RecipeInterface = {
...
+        status: "OK" | "UNKNOWN_USER_ID_ERROR" | "EMAIL_ALREADY_EXISTS_ERROR" | "EMAIL_CHANGE_NOT_ALLOWED";
```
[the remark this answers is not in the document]
bhumilsarvaiya: not sure. Also, shouldn't that be part of other PR discussion and not account-linking ?

lib/ts/recipe/accountlinking/accountLinkingClaim.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * We include "Class" in the class name, because it makes it easier to import the right thing (the instance) instead of this.
+ * */
+export class AccountLinkingClaimClass extends PrimitiveClaim<string> {
+    constructor() {
+        super({
+            key: "st-acl",
+            fetchValue(_, __, ___) {
+                return undefined;
+            },
```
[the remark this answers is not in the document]
bhumilsarvaiya: No, we just need to know what recipeUserId is set during emailVerification APIs. No need to check the account to be linked table here

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -151,7 +144,7 @@ export type RecipeInterface = {
...
     }>;
```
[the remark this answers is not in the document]
bhumilsarvaiya: Yes

lib/ts/recipe/emailpassword/api/implementation.ts
bhumilsarvaiya: Not really required cause `doPostSignUpAccountLinkingOperations` calls the function

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -90,21 +243,167 @@ export default function getAPIImplementation(): APIInterface {
...
-            if (user === undefined) {
+            /**
+             * check if primaryUserId is linked with this email
+             */
+            let users = await listUsersByAccountInfo({
+                email,
+            });
+            if (users !== undefined) {
+                let primaryUser = users.find((u) => u.isPrimaryUser);
+                if (primaryUser !== undefined) {
```
rishabhpoddar, line 257: why have you flipped the conditions here in comparison to the diagram? In the diagram, we check for if email password used exists for the input email first, and then check if that user is a primary user or not - but here you are doing it the other way around.

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
...
+                            await emailVerificationInstance.recipeInterfaceImpl.verifyEmailUsingToken({
+                                token: tokenResponse.token,
+                                userContext,
+                            });
+                        }
+                    }
+                }
+                let user = await getUser(userId);
+                if (user === undefined) {
+                    throw Error("this error should never be thrown");
```
rishabhpoddar, line 490: Return RESET_PASSWORD_INVALID_TOKEN_ERROR instead?

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
...
+                         * create a new emailpassword recipe user
+                         */
+                        let response = await options.recipeImplementation.signUp({
+                            email,
+                            password: newPassword,
+                            userContext,
+                            doAccountLinking: false,
+                        });
+                        if (response.status !== "OK") {
+                            throw Error("this error should not be thrown. EP user already for email: " + email);
```
rishabhpoddar, line 513: we should account for race conditions properly.

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
...
+
+                        if (tokenResponse.status === "OK") {
+                            await emailVerificationInstance.recipeInterfaceImpl.verifyEmailUsingToken({
+                                token: tokenResponse.token,
+                                userContext,
+                            });
+                        }
+                    }
+                }
+                let user = await getUser(userId);
```
rishabhpoddar, line 488: variable names are bad..

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
...
+                            password: newPassword,
+                            userContext,
+                            doAccountLinking: false,
+                        });
+                        if (response.status !== "OK") {
+                            throw Error("this error should not be thrown. EP user already for email: " + email);
+                        }
+                        let recipeUser = response.user;
+                        await verifyUser(response.user.id);
+                        await linkAccounts(recipeUser.id, user.id, userContext);
```
rishabhpoddar, line 517: you need to take into account responses from these APIs for race conditions?

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
                 userContext,
             });
 
+            if (response.status === "OK") {
```
rishabhpoddar, line 466: the token consumption API called on top will just consume the token and not actually change the password.. see diagram. So the API in the core will need to change to just consume the token and return OK or INVALID_TOKEN. 

Therefore, the recipe interface will also have to change to rename resetPasswordUsingToken to something like `consumePasswordResetToken`

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
...
+                        });
+                        if (response.status !== "OK") {
+                            throw Error("this error should not be thrown. EP user already for email: " + email);
+                        }
+                        let recipeUser = response.user;
+                        await verifyUser(response.user.id);
+                        await linkAccounts(recipeUser.id, user.id, userContext);
+                    } else if (!epUser.verified) {
+                        await verifyUser(epUser.recipeUserId);
+                    }
```
rishabhpoddar, line 520: not as per diagram -> you need to call updatePassword function here + call createPrimaryUserIdOrLinkAccounts?

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
...
+                } else {
+                    /**
```
rishabhpoddar, line 522: you need to check if user has an emailpassword account, if not, we return INVALID_TOKEN_ERROR - see diagram

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -193,11 +584,12 @@ export default function getAPIImplementation(): APIInterface {
             }
             let user = response.user;
 
+            let recipeUser = user.loginMethods.find((u) => u.recipeId === "emailpassword" && u.email === email);
             let session = await Session.createNewSession(
                 options.req,
                 options.res,
                 user.id,
-                user.recipeUserId,
+                recipeUser?.recipeUserId,
```
rishabhpoddar, line 592: recipeUser should never be undefined here right? So instead of doing `?`, we should check if it's undefined and if it is, throw an error like should never come here

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -239,17 +631,37 @@ export default function getAPIImplementation(): APIInterface {
...
+            if (!isSignUpAllowed) {
+                return {
+                    status: "SIGNUP_NOT_ALLOWED",
+                    reason: "the sign-up info is already associated with another account where it is not verified",
+                };
+            }
+            let response = await options.recipeImplementation.signUp({
+                email,
+                password,
+                doAccountLinking: true,
```
rishabhpoddar, line 651: to discuss this - we should call shouldDoAccountLinking instead of assuming that it's true - or remove this boolean if we are calling shouldDoAccountLinking inside this function already

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/recipeImplementation.ts
```diff
@@ -1,22 +1,51 @@
...
+import AccountLinking from "../accountlinking/recipe";
 import { Querier } from "../../querier";
 import NormalisedURLPath from "../../normalisedURLPath";
+import { getUser, listUsersByAccountInfo } from "../..";
+import EmailVerification from "../emailverification/recipe";
+import { User } from "../../types";
 
 export default function getRecipeInterface(querier: Querier): RecipeInterface {
     return {
         signUp: async function ({
```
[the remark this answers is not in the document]
rishabhpoddar: We decided to add a new recipe function called `_createNewUser` which will just take email and password and return {status: "OK", user: {id, timeJoined, email} | {status: "EMAIL_ALREADY_EXISTS_ERROR"}

This means that the password reset and post login account linking APIs should not call the sign up recipe function anymore, but should call this new `_createNewUser` function.

Finally, we can get rid of the `doAccountLinking` input from the sign up recipe function and index.ts file

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -102,8 +87,20 @@ export type RecipeInterface = {
...
+         * of the parameter would be false. So if we are moving the account
+         * linking part to be part of this function, ideally we should keep
+         * this boolean parameter
+         */
+        doAccountLinking: boolean;
         userContext: any;
-    }): Promise<{ status: "OK"; user: User } | { status: "EMAIL_ALREADY_EXISTS_ERROR" }>;
+    }): Promise<{ status: "OK"; createdNewUser: boolean; user: User } | { status: "EMAIL_ALREADY_EXISTS_ERROR" }>;
 
     signIn(input: {
```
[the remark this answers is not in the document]
rishabhpoddar: We decided to remove recipeUserId from the input

lib/ts/recipe/emailpassword/recipe.ts
```diff
@@ -217,12 +224,17 @@ export default class Recipe extends RecipeModule {
...
-        let userInfo = await this.recipeInterfaceImpl.getUserById({ userId, userContext });
-        if (userInfo !== undefined) {
-            return {
-                status: "OK",
-                email: userInfo.email,
-            };
+        let user = await this.recipeInterfaceImpl.getUserById({ userId, userContext });
+        if (user !== undefined) {
+            let recipeLevelUser = user.loginMethods.find(
+                (u) => u.recipeId === "emailpassword" && u.recipeUserId === userId
```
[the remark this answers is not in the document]
rishabhpoddar: We decided that we will rename this function cause we expect only recipeUserId to be passed into it.

lib/ts/recipe/emailpassword/api/implementation.ts
rishabhpoddar: Add a comment here saying that this will be true cause of xyz reason..

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -160,6 +463,94 @@ export default function getAPIImplementation(): APIInterface {
...
+                            session,
+                            userContext,
+                        }
+                    );
+                    if (result.createNewSession) {
+                        await createNewSession(
+                            options.req,
+                            options.res,
+                            result.primaryUserId,
+                            result.recipeUserId,
```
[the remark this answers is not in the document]
rishabhpoddar: User exposed functions:
- createNewSession(req, res, userId, ...) -> documented in session recipe
- createNewSessionForPrimaryUser(req, res, userId, recipeUserId, ...) -> will be exposed to users, but not documented

Recipe function:
createNewSession(req, res, userId, recipeUserId, ...)

rishabhpoddar (changes requested):

Issue #508, opened 2023-03-06
BaseRequest getHeaderValue should consider case sensitivity for AWS Lambda


nkshah2:
AWS Lambda does not guarantee that headers received are always lower cased, for example when using the dashboard recipe the header received by the APIs are "Authorization" and not "authorization" (which is what the dashboard recipe tries to read). 

The `getHeaderValue` function in the BaseRequest implementation for AWS Lambda does not check for case sensitivity when trying to read headers and as a result dashboard APIs always throw a 401.

PR #509, opened 2023-03-06
fix: Fix issue of BaseRequest getHeaderValue not handling case insensitivity

nkshah2:
## Summary of change
- Adds a utility function that fetches a value from any object in a case insensitive way
- Adds a utils test file to test the functionality of the new function

## Related issues

-  #508 

## Test Plan

Adds tests

## Documentation changes

None needed

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Update version and CHANGELOG

### DIFF
--- a/test/utils.test.js
+++ b/test/utils.test.js
@@ -1,0 +1,20 @@
+const assert = require("assert");
+const { getFromObjectCaseInsensitive } = require("../lib/build/utils");
+
+describe("SuperTokens utils test", () => {
+    it("Test getFromObjectCaseInsensitive", () => {
+        const testObj = {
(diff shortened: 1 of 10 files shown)


test/utils.test.js
```diff
@@ -0,0 +1,20 @@
+const assert = require("assert");
+const { getFromObjectCaseInsensitive } = require("../lib/build/utils");
+
```
rishabhpoddar, line 3: add tests for the framework itself

rishabhpoddar (changes requested):

PR #511, opened 2023-03-14
review changes

rishabhpoddar:
https://github.com/supertokens/supertokens-node/pull/501


lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -585,11 +585,16 @@ export default function getAPIImplementation(): APIInterface {
             let user = response.user;
 
             let recipeUser = user.loginMethods.find((u) => u.recipeId === "emailpassword" && u.email === email);
+
+            if (recipeUser === undefined) {
+                throw new Error("Race condition error - please call this API again");
```
rishabhpoddar, line 590:
```suggestion
                throw new Error("Should never come here");
```

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -657,11 +663,15 @@ export default function getAPIImplementation(): APIInterface {
             let user = response.user;
             let recipeUser = user.loginMethods.find((u) => u.recipeId === "emailpassword" && u.email === email);
 
+            if (recipeUser === undefined) {
+                throw new Error("Race condition error - please call this API again");
```
rishabhpoddar, line 667: what should we do here instead of throwing this error?

PR #512, opened 2023-03-14
feat: access token to jwt

porcellus:
## Summary of change

Matching new CDI. Some related changes (e.g.: renaming session data) will be done in separate PRs.

## Related issues

-  

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-  [ ] Changelog
-  [ ] More testing


lib/ts/recipe/session/accessToken.ts
```diff
@@ -31,21 +32,44 @@ export async function getInfoFromAccessToken(
     timeCreated: number;
 }> {
     try {
-        verifyJWT(jwtInfo, jwtSigningPublicKey);
-        const payload = jwtInfo.payload;
+        // From the library examples
+        const { payload } = await jose.jwtVerify(jwtInfo.rawTokenString, jwks).catch(async (error) => {
```
rishabhpoddar, line 36: Is the `.catch` async? Cause then you should convert this to promise style and await and then catch the error, otherwise `validateAccessTokenStructure` in the next line will start running

lib/ts/recipe/session/accessToken.ts
```diff
@@ -31,21 +32,44 @@ export async function getInfoFromAccessToken(
...
+            // We only want to opt-into this for V2 access tokens
+            if (jwtInfo.version === 2 && error?.code === "ERR_JWKS_MULTIPLE_MATCHING_KEYS") {
```
rishabhpoddar, line 38: what about version < 2?

lib/ts/recipe/session/accessToken.ts
```diff
@@ -31,21 +32,44 @@ export async function getInfoFromAccessToken(
...
+
+            throw error;
+        });
 
         // This should be called before this function, but the check is very quick, so we can also do them here
-        validateAccessTokenStructure(payload);
+        validateAccessTokenStructure(payload, jwtInfo.version);
 
         // We can mark these as defined (the ! after the calls), since validateAccessTokenPayload checks this
+        let userId = jwtInfo.version === 2 ? sanitizeStringInput(payload.userId)! : sanitizeStringInput(payload.sub)!;
```
rishabhpoddar, line 61: lesser than equal to 2?

lib/ts/recipe/session/accessToken.ts
```diff
@@ -31,21 +32,44 @@ export async function getInfoFromAccessToken(
...
+        let expiryTime =
+            jwtInfo.version === 2 ? sanitizeNumberInput(payload.expiryTime)! : sanitizeNumberInput(payload.exp)! * 1000;
+        let timeCreated =
+            jwtInfo.version === 2
```
rishabhpoddar, line 65: lesser than equal to 2?

lib/ts/recipe/session/accessToken.ts
```diff
@@ -31,21 +32,44 @@ export async function getInfoFromAccessToken(
...
+                ? sanitizeNumberInput(payload.timeCreated)!
+                : sanitizeNumberInput(payload.iat)! * 1000;
+        let userData = jwtInfo.version === 2 ? payload.userData : payload;
```
rishabhpoddar, line 68: lesser than equal to 2?

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -18,6 +18,16 @@ import STError from "./error";
 import { SessionClaim, SessionClaimValidator, SessionContainerInterface, TokenTransferMethod } from "./types";
 import { Helpers } from "./recipeImplementation";
```
rishabhpoddar, line 19: We should remove updateAccessTokenPayload since this is a breaking change anyway, and we do not want people to call updateAccessTokenPayload thinking that it's a merge.

lib/ts/recipe/session/index.ts
```diff
@@ -44,6 +44,7 @@ export default class SessionWrapper {
         userId: string,
         accessTokenPayload: any = {},
         sessionData: any = {},
+        useDynamicAccessTokenSigningKey?: boolean,
         userContext: any = {}
```
rishabhpoddar, line 48: We should remove updateAccessTokenPayload since this is a breaking change anyway, and we do not want people to call updateAccessTokenPayload thinking that it's a merge.

lib/ts/recipe/session/jwt.ts
```diff
@@ -45,12 +45,21 @@ export function parseJWTWithoutSignatureVerification(jwt: string): ParsedJWTInfo
...
     }
 
+    // V1&V2 is functionally identical, plus all legacy tokens should be V2 now.
+    let version = 2;
+    // V2 or older tokens did not save the key id;
     // checking header
     if (!HEADERS.has(splittedInput[0])) {
-        throw new Error("JWT header mismatch");
+        const parsedHeader = JSON.parse(Buffer.from(splittedInput[0], "base64").toString());
+        if (parsedHeader.typ !== "JWT" || parsedHeader.version !== "3" || parsedHeader.kid === undefined) {
```
rishabhpoddar, line 54: !== "3" -> should be < 3?

lib/ts/recipe/session/recipe.ts
```diff
@@ -79,39 +78,17 @@ export default class SessionRecipe extends RecipeModule {
                 issuer: this.config.jwt.issuer,
                 override: this.config.override.openIdFeature,
             });
-
```
rishabhpoddar, line 82: we will always need the openIdRecipe now cause we will / should always expose the jwks endpoint now?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -80,41 +49,29 @@ export default function getRecipeInterface(
...
-                response.accessTokenValidity,
-                response.refreshTokenValidity,
-                response.jwtSigningPublicKeyList
-            );
-
-            updateJwtSigningPublicKeyInfo(
-                response.jwtSigningPublicKeyList,
-                response.jwtSigningPublicKey,
-                response.jwtSigningPublicKeyExpiryTime
+    const JWKS: ReturnType<typeof createRemoteJWKSet>[] = querier.getUrlsForPath("/.well-known/jwks.json").map((url) =>
```
rishabhpoddar, line 52: We should allow the user to also specify the jwks endpoint cause in m2m case, someone may use our backend sdk without connected to a core and those getSession function calls should verify by querying the apiDomain's jwks endpoint?

This is a very high level thought - to discuss

lib/ts/recipe/session/index.ts
```diff
@@ -44,6 +44,7 @@ export default class SessionWrapper {
         userId: string,
         accessTokenPayload: any = {},
         sessionData: any = {},
```
rishabhpoddar, line 46: need to add getSession / verifySession that do not depend on the req / res (our previous discussion about this)

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -83,87 +72,73 @@ export async function getSession(
...
                 throw err;
             }
-
-            // If we reached a key older than the token and failed to validate the token,
-            // that means it was signed by a key newer than the cached list.
-            // In this case we go to the server.
-            if (timeCreated >= key.createdAt) {
-                foundASigningKeyThatIsOlderThanTheAccessToken = true;
-                break;
+            if (timeCreated <= Date.now() - 60000) {
```
rishabhpoddar, line 125: would be good to have an explanation of this magic number?

lib/ts/recipe/session/sessionFunctions.ts
```diff
@@ -83,87 +72,73 @@ export async function getSession(
...
 
-    // If the token was created before the oldest key in the cache but hasn't expired, then a config value must've changed.
-    // E.g., the access_token_signing_key_update_interval was reduced, or access_token_signing_key_dynamic was turned on.
-    // Either way, the user needs to refresh the access token as validating by the server is likely to do nothing.
-    if (!foundASigningKeyThatIsOlderThanTheAccessToken) {
-        throw new STError({
-            message: "Access token has expired. Please call the refresh API",
-            type: STError.TRY_REFRESH_TOKEN,
-        });
-    }
```
rishabhpoddar, line 159: won't this check be required for version <= 2 of the access token?

lib/ts/recipe/session/recipeImplementation.ts
```diff
@@ -163,6 +120,9 @@ export default function getRecipeInterface(
                 helpers,
                 userId,
                 disableAntiCSRF,
+                useDynamicAccessTokenSigningKey !== undefined
+                    ? useDynamicAccessTokenSigningKey
+                    : config.useDynamicAccessTokenSigningKey === false,
```
rishabhpoddar, line 125:
```suggestion
                useDynamicAccessTokenSigningKey !== undefined
                    ? !useDynamicAccessTokenSigningKey
                    : config.useDynamicAccessTokenSigningKey === false,
```

rishabhpoddar: Should we change createNewSession to take useStaticKeys with default to false instead of using useDynamicAccessTokenSigningKey? Cause the core too takes useStaticKeys

rishabhpoddar (changes requested):

lib/ts/recipe/session/index.ts
porcellus: That'll be done in a separate PR

lib/ts/recipe/session/recipe.ts
porcellus: We don't always need to expose the JWKS endpoint from the backend. The FE doesn't need it, and the m2m/multiple backends use-case is fairly rare, I think.

coreDriverInterfaceSupported.json
```diff
@@ -1,4 +1,4 @@
 {
     "_comment": "contains a list of core-driver interfaces branch names that this core supports",
-    "versions": ["2.8", "2.9", "2.10", "2.11", "2.12", "2.13", "2.14", "2.15", "2.16", "2.17", "2.18"]
+    "versions": ["2.19"]
```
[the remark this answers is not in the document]
porcellus: I don't think the things covered by the FDI will change - only the structure of the access token will change.
If we assume that the access token payload is accessed through `getAccessTokenPayload`, then the only change is that extra props will appear plus the JWT added as a prop will disappear. This is a big change and may require migration on the FE, but the FE SDKs can remain the same.

lib/ts/recipe/session/accessToken.ts
porcellus: Since V1 and V2 are functionally the same, we never have version = 1

lib/ts/recipe/session/recipeImplementation.ts
porcellus: Let's discuss this - but I don't see a significant advantage of always exposing it:
- in the m2m use-case they need the recipe to create the JWTs anyway. 
- in the multiple backends use case (where multiple backends can verify, but only a single backend can create sessions), I think it still makes sense that this endpoint has to be enabled

lib/ts/recipe/session/sessionFunctions.ts
porcellus: This is basically replaced by the max-age check

lib/ts/recipe/session/types.ts
```diff
@@ -27,23 +27,6 @@ export type KeyInfo = {
 };
```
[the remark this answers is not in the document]
porcellus: This will be done in a separate PR

lib/ts/recipe/session/types.ts
```diff
@@ -143,9 +128,9 @@ export type TypeNormalisedInput = {
     }) => TokenTransferMethod | "any";
 
     invalidClaimStatusCode: number;
+    exposeAccessTokenToFrontendInCookieBasedAuth: boolean;
     jwt: {
         enable: boolean;
-        propertyNameInAccessTokenPayload: string;
         issuer?: string;
     };
```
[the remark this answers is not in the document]
porcellus: it also initializes the OpenId recipe. Since we wanted to keep the enable option we also have to keep the issuer.

package.json
```diff
@@ -39,6 +39,7 @@
         "co-body": "6.1.0",
         "cookie": "0.4.0",
```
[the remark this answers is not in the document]
porcellus: It is.

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -93,6 +103,10 @@ export default class Session implements SessionContainerInterface {
     // Any update to this function should also be reflected in the respective JWT version
     async mergeIntoAccessTokenPayload(accessTokenPayloadUpdate: any, userContext?: any): Promise<void> {
         const updatedPayload = { ...this.getAccessTokenPayload(userContext), ...accessTokenPayloadUpdate };
+        for (const key of protectedProps) {
+            delete updatedPayload[key];
+        }
```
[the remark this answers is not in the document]
porcellus: I didn't change this, but I see no big advantage to moving it elsewhere, which would also change how the `regenerateAccessToken` function works, changing the interface.

lib/ts/recipe/session/accessToken.ts
```diff
@@ -31,21 +32,44 @@ export async function getInfoFromAccessToken(
...
+                    try {
+                        return await jose.jwtVerify(jwtInfo.rawTokenString, publicKey);
+                    } catch (innerError) {
+                        if (innerError?.code === "ERR_JWS_SIGNATURE_VERIFICATION_FAILED") {
+                            continue;
+                        }
+                        throw innerError;
+                    }
+                }
+                throw new jose.errors.JWSSignatureVerificationFailed();
```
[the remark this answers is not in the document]
porcellus: It's tested by adding verifying a v2 access token, but I also added a specific `ProcessState` to make it more explicit.

lib/ts/recipe/session/jwt.ts
porcellus: I think it's better to be checking the string match here:
- If we add support for future access token versions, we likely should add new validation rules anyway
- it's a string that we do not need to parse here.

lib/ts/recipe/session/recipeImplementation.ts
porcellus: I think these were the names In the ADR. I'm open to changing them, but defaulting to true seemed more natural.

lib/ts/recipe/session/index.ts
porcellus: https://github.com/supertokens/supertokens-node/pull/520

lib/ts/recipe/session/types.ts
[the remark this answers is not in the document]
porcellus: #519

coreDriverInterfaceSupported.json
```diff
@@ -1,4 +1,4 @@
 {
     "_comment": "contains a list of core-driver interfaces branch names that this core supports",
-    "versions": ["2.8", "2.9", "2.10", "2.11", "2.12", "2.13", "2.14", "2.15", "2.16", "2.17", "2.18"]
```
rishabhpoddar, line 3: To discuss affect on:
- websocket auth
- graphql docs
- using regular jwt lib vs getSession
- m2m auth
- one API being used for m2m and being called from frontend

rishabhpoddar (changes requested):

coreDriverInterfaceSupported.json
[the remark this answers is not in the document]
rishabhpoddar: You should add a test on the frontend SDK related to the new exposeAccessTokenToTheFrontend boolean. For example, we want to make sure that refreshing still works in cookie based even if the access token is exposed to the frontend. 

I'm keeping this comment open until new tests are added related to this change.

lib/ts/recipe/session/types.ts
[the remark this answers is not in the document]
rishabhpoddar: But the issuer is not used when creating  a new access token for the session? Maybe we should, and maybe we should extract the issuer outside of the jwt config. So the jwt config will only have enable or not

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -92,14 +102,52 @@ export default class Session implements SessionContainerInterface {
...
+        if (response === undefined) {
+            throw new STError({
+                message: "Session does not exist anymore",
+                type: STError.UNAUTHORISED,
+            });
+        }
+
+        this.userDataInAccessToken = response.session.userDataInJWT;
+        if (response.accessToken !== undefined) {
+            this.accessToken = response.accessToken.token;
```
rishabhpoddar, line 131: - if exposeToFrontend is true, then we also need to set the access token in header? Where else is this missed?
- consider moving this part of setting tokens to a single place which checks for the new config (exposeToFrontend), and sets the tokens accordingly

lib/ts/recipe/session/sessionClass.ts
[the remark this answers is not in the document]
rishabhpoddar: Removing the protected props is a new thing? I suggest that you do not do this check at all in the backend SDK since the core will anyway throw an error.

lib/ts/recipe/session/jwt.ts
rishabhpoddar: There could be future versions added that is still a JWT but have some other claim (similar to refreshTokenHash1) which is required. So this file doesn't change, but instead the accessToken file will only change

rishabhpoddar: > it's a string that we do not need to parse here.

You already parse it: `version = Number.parseInt(parsedHeader.version)`

lib/ts/recipe/session/recipe.ts
rishabhpoddar: Yea.. but we should i think. It's expected and easier docs + less configuration for the user. The m2m, and multiple backend is not really rare - most medium / large users have this need.

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: To discuss

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: Agreed about the defaulting to true, so then we should change the core's API - this inconsistency is not necessary and causes confusion in the long run

lib/ts/recipe/session/sessionClass.ts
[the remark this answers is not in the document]
porcellus: Right - I was looking at the wrong `delete`. Anyway, we do this here because `getAccessTokenPayload` will return the protected props, meaning the check in the core endpoint will always fail. We could:
1. do this here (and we should do it at this level, I think, because I don't think we should change the `regenerateAccessToken` behaviour)
2. provide a `mergeIntoAccessTokenPayload` API on the CDI.

I think 2. is the better solution, but we've been putting it off for a while, and it's not critical.

lib/ts/recipe/session/types.ts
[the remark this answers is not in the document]
rishabhpoddar: We decided to get rid of the issuer entirely from the sdk.

lib/ts/recipe/session/sessionClass.ts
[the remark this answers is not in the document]
rishabhpoddar: - We can remove the protected props in the regerateAccessToken recipe implementation
- We shuold get rid of the regerateAccessToken function from session index

lib/ts/recipe/session/recipe.ts
rishabhpoddar: We also decided to getting rid of `jwt: {enable: true}`, and to initlaise openid inside session anyway

lib/ts/recipe/session/recipeImplementation.ts
rishabhpoddar: We decided that we will not do this and instead recommend people to use a standard jwt verification lib with claim checking

coreDriverInterfaceSupported.json
rishabhpoddar: # websocket auth
- cookie based
   - enabled exposeAccessTokenToFrontend
   - on frontend get the access token using session.getAccessTokens
   - send that to websocket connection as part of the message
   - use standard jwt lib in websocket to verify the token pointing to the jwks endpoint
   - session refreshing will not be an issue cause when we call session.getAccessTokens on the frontend, it ensures that the access token is valid.
- header based
   - on frontend get the access token using session.getAccessTokens
   - send that to websocket connection as part of the message
   - use standard jwt lib in websocket to verify the token pointing to the jwks endpoint
   - session refreshing will not be an issue cause when we call session.getAccessTokens on the frontend, it ensures that the access token is valid.

# graphql 
TODO

porcellus: ## GraphQL

The existing guide should still work and it's still necessary if we recommend people to use `getSession`. It could be made easier however:
1. By recommending `getSessionWithoutModifyingResponse` and handling setting things in the response themselves.
2. By recommending standard JWT libraries. This will limit how they interface with our sessions, but they only need to handle sending a 401 + the GQL servers generally have guides about JWT auth

No change necessary/recommended on the FE, since regular interception should already take care of everything.

## using regular jwt lib vs getSession

- This will enable people to use standard JWT verification libraries instead of our `getSession` on the backend.
- It can make integration with API gateway and other similar things easier as there are existing guides/built-in solutions for JWT based auth for most.
- These verification libs generally send 401s for invalid tokens, which can make integration even easier
- `exposeAccessTokenToFrontend` makes it possible to use cookie based auth on some (i.e.: the auth domain) and header based on other domains (using `Session.getAccessToken()` and setting it as the auth header manually for other domains.

No change necessary/recommended on the FE, since regular interception should already take care of everything.

## m2m auth

No change - the tokens issued to users in general should not be used in M2M communication. The JWT recipe should be used to create the tokens used in M2M communication, and the JWKs endpoint exposed both by the core and the BE SDK can be used for validation.

## one API being used for m2m and being called from frontend

- Now it can use a single JWT validation middleware
- It then needs to validate where the request came from - e.g.: by checking the sub
- Existing implementations will keep working, `verifySession` can still differentiate the access tokens from jwts created otherwise

## Session sharing across API domains (not just sub-domains)

- Header-based sessions
  - no change
- Cookie-based sessions
  - The auth domain can use cookies after this change
  - This means that the refresh token is still stored securely in an HTTPOnly cookie.
  - Steps:
    - enable `exposeAccessTokenToFrontend`
    - on frontend get the access token using `Session.getAccessToken`
    - attach the access token as an authorization header to requests made to API domains
    - handle 401s coming from API domains by calling `attemptRefreshingSession`

rishabhpoddar: Sounds good.

lib/ts/recipe/session/sessionClass.ts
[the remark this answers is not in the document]
rishabhpoddar: - On second thought, if we are removing the protected prop right before calling the core - and we are OK with that, it's as good as just putting that check in the core itself? I prefer if it's in the core as lesser work for backend sdks.
- If we are silently ignoring protected props for mergeIntoAccessTokenPayload, should we also silently be ignoring during createNewSession / refreshSession? It's strange that in some place it throws an error, and in some place it's ignored.

TO DISCUSS

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionClass.ts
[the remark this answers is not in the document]
rishabhpoddar: We finally decided to move it into mergeIntoAccessTokenPayload (in sessionClass and in index.ts of this function) before merging the user input into the existing payload

CHANGELOG.md
```diff
@@ -9,9 +9,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+-   Added new optional `useDynamicAccessTokenSigningKey` param to `createNewSession`
+-   Removed deprecated `updateAccessTokenPayload` and `regenerateAccessToken` from the Session recipe interface
+-   Removed `getAccessTokenLifeTimeMS` and `getRefreshTokenLifeTimeMS` functions
+
+## Changes
+
+-   The Session recipe now always initializes the OpenID recipe if it hasn't been initialized.
+-   Refactored how access token validation is done
+-   Removed the handshake call to improve start-up times
+-   Added support for new access token version
```
rishabhpoddar, line 49: Need to have a migration guide:
- For example, on the frontend, if you getting the jwt you are doing getAccessTokenPayloadSecurely().jwt ->
```
let jwt = null;
let accessTokenPayload = await getAccessTokenPayloadSecurely()
if (accessTokenPayload.jwt === undefined) {
   jwt = await session.getAccessToken();
} else {
   jwt = accessTokenPayload.jwt;
}
```
- Another example is on the backend, if you are using sAccessToken to get he JWT and then do JWT verification of it, now it will be directly done on the sAccessToken (you need to detect the right version and do it based on that).
- You need to update the core.

CHANGELOG.md
```diff
@@ -9,9 +9,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking Changes
 
+-   Added support for CDI version `2.19`
+-   Dropped support for CDI version `2.8`-`2.18`
+-   Changed the interface and configuration of the Session recipe, see below for details. If you do not use the Session recipe directly and do not provide custom configuration, then no migration is necessary.
+-   `getAccessTokenPayload` will now return standard (`sub`, `iat`, `exp`) claims along the user defined ones in `getAccessTokenPayload`
```
rishabhpoddar, line 15: along with the non standard claim list

rishabhpoddar (changes requested):

CHANGELOG.md
```diff
@@ -9,9 +9,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking Changes
```
rishabhpoddar, line 10: test that older backend SDK tests still work with the new core

rishabhpoddar (changes requested):

CHANGELOG.md
porcellus: Ran all tests with the old node SDK (master). All tests passed, the only change required was the name of core config values we set in the tests.

PR #519, opened 2023-03-17
feat!: rename sessionData handling functions

porcellus:
## Summary of change

Rename sessionData handling functions as discussed on slack

## Related issues

-   

## Test Plan

No behaviour change

## Documentation changes

No behaviour change, but will need to update the name in the docs. Done later at the same time as the rest of the rework.

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/session/types.ts
```diff
@@ -256,7 +256,11 @@ export type RecipeInterface = {
     revokeMultipleSessions(input: { sessionHandles: string[]; userContext: any }): Promise<string[]>;
 
     // Returns false if the sessionHandle does not exist
```
rishabhpoddar, line 258: the input to createNewSession still uses `sessionData`. Please rename that to `sessionDataInDatabase` as well. Is there any other place where sessionData is being used?

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #528, opened 2023-03-30
Emailverification accountlinking

bhumilsarvaiya:


lib/ts/recipe/dashboard/api/userdetails/userEmailVerifyTokenPost.ts
```diff
@@ -19,13 +21,20 @@ export const userEmailVerifyTokenPost = async (_: APIInterface, options: APIOpti
         });
     }
```
rishabhpoddar, line 22: Should the frontend sent recipeUserId?

lib/ts/recipe/emailverification/index.ts
```diff
@@ -38,22 +38,9 @@ export default class Wrapper {
     > {
         const recipeInstance = Recipe.getInstanceOrThrowError();
```
rishabhpoddar, line 39: - We probably don't need `getEmailForUserId` on a per recipe anymore cause we can directly use supertokens.getUser function if the config for this is not defined
- We should rename `getEmailForUserId` to `getEmailForRecipeUserId`

lib/ts/recipe/emailverification/api/implementation.ts
```diff
@@ -5,62 +5,228 @@ import { GeneralErrorResponse } from "../../../types";
...
+                    // one that just got verified and that we ARE doing post login
+                    // account linking. So this is only for (Case 3) and (Case 4)
+
+                    // In both case 3 and case 4, we do not want to change anything in the
+                    // current session in terms of user ID or email verification claim (since
+                    // both of these refer to the current logged in user and not the newly
+                    // linked user's account). Instead, we just want to remove the
+                    // account linking claim from the session.
+
+                    // TODO: remove account linking claim
```
rishabhpoddar, line 125: TODO..

lib/ts/recipe/emailverification/api/implementation.ts
```diff
@@ -5,62 +5,228 @@ import { GeneralErrorResponse } from "../../../types";
...
+                        // if the session's primary user ID is equal to the
+                        // primary user ID that the account was linked to, then
+                        // this means that the new account became a primary user (Case 1)
+                        // We also have the sub cases here that the account that just
+                        // got verified was already linked to the session's primary user ID,
+                        // but either way, we don't need to change any user ID.
+
+                        // In this case, all we do is to update the emailverification claim
+                        try {
+                            // TODO: fetchValue should take a recipeId as well now
```
rishabhpoddar, line 72: TODO

rishabhpoddar (changes requested):

lib/ts/recipe/emailverification/api/implementation.ts
```diff
@@ -5,62 +5,228 @@ import { GeneralErrorResponse } from "../../../types";
...
-                    if ((err as Error).message === "UNKNOWN_USER_ID") {
-                        throw new SessionError({
-                            type: SessionError.UNAUTHORISED,
-                            message: "Unknown User ID provided",
-                        });
+            if (verifyTokenResponse.status === "EMAIL_VERIFICATION_INVALID_TOKEN_ERROR") {
+                return verifyTokenResponse;
+            }
+
+            let primaryUserIdThatTheAccountWasLinkedTo = await AccountLinking.getInstanceOrThrowError().createPrimaryUserIdOrLinkAccounts(
```
rishabhpoddar, line 33: There is a comment in the lucid chart:

> There is a race condidtion where post account link callback can be called twice. One when the user clicks on the email verification link which has a session and one when the email verify API is called with the same session. The race condition happens if both happens at the same time. But, it is very rare as this would require user to refresh the page manually while clicking on the email verfication link

How is this correct? Cause the email verify GET API never does account linking right? So the post account linking callback will not get called ever..

lib/ts/recipe/emailverification/api/implementation.ts
```diff
@@ -5,62 +5,230 @@ import { GeneralErrorResponse } from "../../../types";
...
+                        };
+                    } else {
+                        // if the session's primary user ID is NOT equal to the
+                        // primary user ID that the account that it was linked to, then
+                        // this means that the new account got linked to another primary user (Case 2)
+
+                        // In this case, we need to update the session's user ID by creating
+                        // a new session
+
+                        // TODO: revoke all session belonging to session.getRecipeUserId()
```
rishabhpoddar, line 98: TODO

rishabhpoddar (changes requested):

lib/build/recipe/emailverification/types.d.ts
```diff
@@ -90,16 +90,16 @@ export declare type RecipeInterface = {
...
         email: string;
         userContext: any;
     }): Promise<{
         status: "OK";
     }>;
     unverifyEmail(input: {
-        userId: string;
+        recipeUserId: string;
         email: string;
         userContext: any;
```
rishabhpoddar, line 104: need to make sure that email verification has recipeUserId everywhere and not just userId

lib/ts/recipe/emailverification/recipe.ts
```diff
@@ -209,4 +216,107 @@ export default class Recipe extends RecipeModule {
...
+                    // if the session's primary user ID is equal to the
+                    // primary user ID that the account was linked to, then
+                    // this means that the new account became a primary user (Case 1)
+                    // We also have the sub cases here that the account that just
+                    // got verified was already linked to the session's primary user ID,
+                    // but either way, we don't need to change any user ID.
+
+                    // In this case, all we do is to update the emailverification claim
+                    try {
+                        // TODO: fetchValue should take a recipeId as well now
```
rishabhpoddar, line 265: TODO

lib/ts/recipe/emailverification/recipe.ts
```diff
@@ -209,4 +216,107 @@ export default class Recipe extends RecipeModule {
...
+                    return;
+                } else {
+                    // if the session's primary user ID is NOT equal to the
+                    // primary user ID that the account that it was linked to, then
+                    // this means that the new account got linked to another primary user (Case 2)
+
+                    // In this case, we need to update the session's user ID by creating
+                    // a new session
+
+                    // TODO: revoke all session belonging to session.getRecipeUserId()
```
rishabhpoddar, line 288: TODO

rishabhpoddar (changes requested):

lib/ts/recipe/emailverification/api/implementation.ts
```diff
@@ -5,62 +5,136 @@ import { GeneralErrorResponse } from "../../../types";
...
+                        options.res,
+                        user.id,
+                        recipeUserId,
+                        session.getAccessTokenPayload(),
+                        await session.getSessionDataFromDatabase()
+                    );
+                }
+            }
+
+            // TODO: update account linking claim from db...
```
rishabhpoddar, line 84: TODO

rishabhpoddar (changes requested):

lib/ts/recipe/emailverification/recipe.ts
```diff
@@ -194,6 +199,9 @@ export default class Recipe extends RecipeModule {
             }
         }
 
+        // TODO: we should just call the getUser function here and loop through
```
rishabhpoddar, line 202: TODO

lib/ts/recipe/emailverification/recipe.ts
```diff
@@ -209,4 +217,105 @@ export default class Recipe extends RecipeModule {
     addGetEmailForUserIdFunc = (func: GetEmailForUserIdFunc): void => {
         this.getEmailForUserIdFuncsFromOtherRecipes.push(func);
```
rishabhpoddar, line 218: where and how can we add a resync for the account linking claim? For example, if the if it's left in the session and the table doesn't have it, we need to update the session at some point independently to the email verification API being completed.

rishabhpoddar (changes requested):

PR #545, opened 2023-04-20
feat: make accessTokenPath configurable

Qdea:
## Summary of change

Due to implementation differences, when using the supertokens-node package, it would make sense to allow the user to configure the accessTokenPath instead of using the path statically. In the current PR, an optional accessTokenPath parameter has been added to TypeNormalisedInput, when specified, a custom path will be used in the user config. If the parameter is not specified, then the default value is taken, which was in the package before the changes.

In addition, minor syntax bugs have been fixed and the documentation in the docs folder, generated using npm run build-docs, has been updated.

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/recipe/session/recipe.ts
```diff
@@ -74,7 +74,7 @@ export default class SessionRecipe extends RecipeModule {
 
         this.isInServerlessEnv = isInServerlessEnv;
 
-        if (this.config.jwt.enable === true) {
+        if (this.config.jwt.enable) {
```
rishabhpoddar, line 77: this is an unrelated change. Please revert it :)

lib/ts/recipe/session/types.ts
```diff
@@ -129,6 +129,7 @@ export type TypeInput = {
 
 export type TypeNormalisedInput = {
     refreshTokenPath: NormalisedURLPath;
+    accessTokenPath?: NormalisedURLPath;
```
rishabhpoddar, line 132: the normalised input should not be optional. You should add `accessTokenPath` to the `TypeInput` type, and then in the `validateAndNormaliseUserInput` function [here](https://github.com/supertokens/supertokens-node/blob/1119aa4866/lib/ts/recipe/session/utils.ts#L127), you should set the normalised input to this value if provided by the user, else to `normaliseURLPathOrThrowError("/")`

lib/ts/recipe/session/utils.ts
```diff
@@ -207,7 +207,7 @@ export function validateAndNormaliseUserInput(
     let accessTokenPayloadJWTPropertyName = "jwt";
     let issuer: string | undefined;
 
-    if (config !== undefined && config.jwt !== undefined && config.jwt.enable === true) {
+    if (config !== undefined && config.jwt !== undefined && config.jwt.enable) {
```
rishabhpoddar, line 210: please revert this change as it's not related to this PR

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -167,7 +165,7 @@ export function setCookie(
     if (pathType === "refreshTokenPath") {
         path = config.refreshTokenPath.getAsStringDangerous();
     } else if (pathType === "accessTokenPath") {
-        path = "/";
+        path = config.accessTokenPath ? config.accessTokenPath.getAsStringDangerous() : "/";
```
rishabhpoddar, line 168:
```suggestion
        path = config.accessTokenPath.getAsStringDangerous() === "" ? "/" : config.accessTokenPath.getAsStringDangerous();
```

rishabhpoddar: This change is suggested cause we will not make `accessTokenPath` optional in the normalised input type

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -143,14 +143,12 @@ export function setHeader(res: BaseResponse, name: string, value: string) {
 
 /**
  *
+ * @param config
```
rishabhpoddar, line 146: please also consider adding a test about this new setting.

rishabhpoddar (changes requested):

lib/ts/recipe/session/cookieAndHeaders.ts
Qdea: Hi, this is not new setting, sorry. It just my IDE change this
image
bc of things like that
image
automatically.

lib/ts/recipe/session/utils.ts
Qdea: Ok, will do

lib/ts/recipe/session/recipe.ts
Qdea: Ok, will do

lib/ts/recipe/session/cookieAndHeaders.ts
Qdea: Done

lib/ts/recipe/session/types.ts
Qdea: Done

lib/ts/normalisedURLPath.ts
```diff
@@ -43,7 +43,7 @@ export default class NormalisedURLPath {
     };
 }
 
-function normaliseURLPathOrThrowError(input: string): string {
+export function normaliseURLPathOrThrowError(input: string): string {
```
rishabhpoddar, line 46: You do not need to export this since the `NormalisedURLPath` class is already exported. Use that instead

lib/ts/recipe/session/utils.ts
```diff
@@ -229,6 +232,7 @@ export function validateAndNormaliseUserInput(
 
     return {
         refreshTokenPath: appInfo.apiBasePath.appendPath(new NormalisedURLPath(REFRESH_API_PATH)),
+        accessTokenPath: appInfo.apiBasePath.appendPath(new NormalisedURLPath(accessTokenPath)),
```
rishabhpoddar, line 235:
```suggestion
        accessTokenPath,
```

rishabhpoddar (changes requested):

lib/ts/recipe/session/cookieAndHeaders.ts
rishabhpoddar: Ah no. I meant that you should add a test about testing the new accessTokenPath config

lib/ts/recipe/session/utils.ts
Qdea: So i dont need to append this?

lib/ts/recipe/session/cookieAndHeaders.ts
Qdea: and where should I put this?

rishabhpoddar: Here are a few examples: https://github.com/supertokens/supertokens-node/blob/master/test/config.test.js

Qdea: Thanks, pushed tests

rishabhpoddar (approved):

PR #565, opened 2023-05-13
feat: tsup, pnpm and new build system

productdevbook:
- dynamic build -> export `dist`
- pnpm support
- esm, cjs support

### DIFF
--- a/lib/ts/framework/utils.ts
+++ b/lib/ts/framework/utils.ts
@@ -138,7 +138,7 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(
             let err = await new Promise((resolve) => {
                 let resolvedCalled = false;
                 if (request.readable) {
-                    jsonParser(request, new ServerResponse(request), (e) => {
+                    jsonParser(request, new ServerResponse(request), (e: any) => {
                         if (!resolvedCalled) {
                             resolvedCalled = true;
                             resolve(e);
@@ -175,7 +175,7 @@ export async function assertFormDataBodyParserHasBeenUsedForExpressLikeRequest(
     let parser = urlencoded({ extended: true });
     let err = await new Promise((resolve) => {
         if (request.readable) {
-            parser(request, new ServerResponse(request), (e) => {
+            parser(request, new ServerResponse(request), (e:any) => {
                 resolve(e);
             });
         } else {
--- a/lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
+++ b/lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
@@ -71,7 +71,10 @@ function defaultCreateAndSendCustomSms(_: NormalisedAppinfo) {
                          * if the error is thrown from API, the response object
                          * will be of type `{err: string}`
                          */
+                        // TODO: check if this is the correct way to check for error
+                        // @ts-ignore
                         if (err.response.data.err !== undefined) {
+                            // @ts-ignore
                             throw Error(err.response.data.err);
                         } else {
                             throw err;
--- a/package.json
+++ b/package.json
@@ -2,14 +2,158 @@
     "name": "supertokens-node",
     "version": "14.0.2",
     "description": "NodeJS driver for SuperTokens core",
-    "main": "index.js",
+    "exports": {
+        ".": {
+            "import": "./dist/index.mjs",
+            "types": "./dist/index.d.ts",
+            "require": "./dist/index.js"
+        },
+        "./nextjs": {
+            "import": "./dist/nextjs.mjs",
+            "types": "./dist/nextjs.d.ts",
+            "require": "./dist/nextjs.js"
+        },
+        "./recipe/dashboard": {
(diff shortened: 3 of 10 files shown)


lib/ts/framework/utils.ts
```diff
@@ -138,7 +138,7 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(
             let err = await new Promise((resolve) => {
                 let resolvedCalled = false;
                 if (request.readable) {
-                    jsonParser(request, new ServerResponse(request), (e) => {
+                    jsonParser(request, new ServerResponse(request), (e: any) => {
```
rishabhpoddar, line 141: can we not use `any` type?

lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
```diff
@@ -71,7 +71,10 @@ function defaultCreateAndSendCustomSms(_: NormalisedAppinfo) {
                          * if the error is thrown from API, the response object
                          * will be of type `{err: string}`
                          */
+                        // TODO: check if this is the correct way to check for error
+                        // @ts-ignore
```
rishabhpoddar, line 75: can we not add `@ts-ignore` to so many files?

rishabhpoddar (changes requested):

rishabhpoddar:
hey @productdevbook could you also tell if this a breaking change for the user? For example, will the import statements for the user change in any way for any recipe / function etc?

lib/ts/framework/utils.ts
productdevbook: you can use, i did any as i don't know what types are

lib/ts/recipe/passwordless/smsdelivery/services/backwardCompatibility/index.ts
productdevbook: there were ts errors in a few places only during the build, I closed them temporarily with this. We can do this better in the next pr.

productdevbook:
It should not affect it in any way, because I took the same paths. I can even create a playground for testing purposes.

productdevbook:
image

rishabhpoddar:
Okay. Im going to try and remove all the newly added `@ts-ignore` annotations before merging this PR.

package.json
```diff
@@ -2,14 +2,205 @@
...
+        "dist",
+        "LICENSE",
+        "README.md"
+    ],
     "scripts": {
         "test": "TEST_MODE=testing npx mocha --timeout 500000",
-        "build-check": "cd lib && npx tsc -p tsconfig.json --noEmit && cd ../test/with-typescript && npx tsc -p tsconfig.json --noEmit",
-        "build": "cd lib && rm -rf build && npx tsc -p tsconfig.json && cd ../test/with-typescript && npx tsc -p tsconfig.json --noEmit && cd ../.. && npm run post-build",
+        "build-check": "npx tsc -p tsconfig.json --noEmit && cd ../test/with-typescript && npx tsc -p tsconfig.json --noEmit",
+        "build": "tsup",
```
rishabhpoddar, line 200: the build command earlier used to also check the compilation of `../test/with-typescript` - this was done as a way to make sure that the interface exposed to the user doesn't break unintentionally.

rishabhpoddar (changes requested):

package.json
productdevbook: you don't need them, tsup sets them all in dist. There is no need for that test, since that user will no longer have ts errors.

productdevbook:
> Okay. Im going to try and remove all the newly added `@ts-ignore` annotations before merging this PR.
> 
> Also, i see the you have removed the build folder and instead using `dist`, but `dist` itself is not part of the git repo (which is fine). This would affect the npm publish process right? Also, if someone is using the github repo directly in the `npm` command (like `npm i git+https://github.com:supertokens/supertokens-node#14.0`), how would they go about generating the required `dist` folder (for the package to work)

yes, used dist. example `npm publish` fist build working and after send. package.json in

```
  "files": [
        "dist",
        "LICENSE",
        "README.md"
    ],
```

only send data.

productdevbook:
`npm i git+https://github.com:supertokens/supertokens-node#14.0` 

I do not know about this. normally it should not be kept on dist vs github in any way.

rishabhpoddar:
> I do not know about this. normally it should not be kept on dist vs github in any way.

Right, we do have requirements that we need the sdk to work even with directly using the git repo. Therefore, we would need the `dist` folder to be a part of the github repo as well (which is why earlier we had the build folder as well). 

If you can think of a better solution to this, it would be great.

productdevbook:
I can't find a solution to this, people who will use it will pr and create and export their github dist using their own url. This should not be in the main library. The solution is actually that simple. We already do the same thing in many libraries.

1. pr and build and push. and me github url copy and paste.

rishabhpoddar:
I'm afraid that won't work for us. Maybe we can add a post npm install hook to do the TS compilation for such users. But we do require some solution that universally works even when pulling the GH repo directly. This is a hard requirement for us i'm afraid.

PR #576, opened 2023-05-30
feat: Introduce TOTP recipe

KShivendu:
## Summary of change

Introduce TOTP recipe

## Issue link
- https://github.com/supertokens/supertokens-core/issues/554

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
```
rishabhpoddar, line 8: please resolve conflicts

rishabhpoddar (changes requested):

CHANGELOG.md
KShivendu: done

lib/ts/recipe/totp/recipeImplementation.ts
```diff
@@ -0,0 +1,96 @@
...
+            let response = await querier.sendPostRequest(
+                new NormalisedURLPath("/recipe/totp/device"),
+                copyAndRemoveUserContext(input)
+            );
+            if (response.status !== "OK") {
+                return response;
+            }
+
+            let issuerName = config.issuer;
+            let userIdentifier = "user@example.com"; // TODO: Fetch this based on first factor?
```
KShivendu, line 40: We can also leave it empty.

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,167 @@
...
+        return [
+            {
+                id: CREATE_DEVICE_API,
+                disabled: this.apiImpl.createDevicePOST === undefined,
+                method: "post",
+                pathWithoutApiBasePath: new NormalisedURLPath(CREATE_DEVICE_API),
+            },
+            {
+                id: VERIFY_DEVICE_API,
+                disabled: this.apiImpl.createDevicePOST === undefined,
```
rishabhpoddar, line 100: shouldnt this be verify device?

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,167 @@
...
+                method: "post",
+                pathWithoutApiBasePath: new NormalisedURLPath(VERIFY_DEVICE_API),
+            },
+            {
+                id: VERIFY_CODE_API,
+                disabled: this.apiImpl.createDevicePOST === undefined,
```
rishabhpoddar, line 106: shouldnt this be verify code?

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,167 @@
...
+                method: "post",
+                pathWithoutApiBasePath: new NormalisedURLPath(VERIFY_CODE_API),
+            },
+            {
+                id: REMOVE_DEVICE_API,
+                disabled: this.apiImpl.createDevicePOST === undefined,
```
rishabhpoddar, line 112: remove device..

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,167 @@
...
+                method: "post",
+                pathWithoutApiBasePath: new NormalisedURLPath(REMOVE_DEVICE_API),
+            },
+            {
+                id: LIST_DEVICE_API,
+                disabled: this.apiImpl.createDevicePOST === undefined,
```
rishabhpoddar, line 118: list device

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+ * under the License.
+ */
+
+import OverrideableBuilder from "supertokens-js-override";
+import { BaseRequest, BaseResponse } from "../../framework";
+import { NormalisedAppinfo } from "../../types";
+import { SessionContainer } from "../session";
+
+export type TypeInput = {
+    issuer: string;
```
rishabhpoddar, line 22: what should the value of this be from the user's point of view?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+    defaultSkew?: number;
+    defaultPeriod?: number;
+    allowUnverifiedDevice?: boolean;
```
rishabhpoddar, line 25: what is the use case of making this true for the user?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+export type RecipeInterface = {
+    createDevice: (input: {
+        userId: string;
+        deviceName: string;
+        skew?: number;
+        period?: number;
+        userContext: any;
+    }) => Promise<{ status: "OK"; qr: string } | { status: "DEVICE_ALREADY_EXISTS_ERROR" }>;
```
rishabhpoddar, line 56: what is qr? a better variable name is needed

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+import OverrideableBuilder from "supertokens-js-override";
+import { BaseRequest, BaseResponse } from "../../framework";
+import { NormalisedAppinfo } from "../../types";
+import { SessionContainer } from "../session";
```
rishabhpoddar, line 19:
```suggestion
import type { SessionContainer } from "../session";
```

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+    req: BaseRequest;
+    res: BaseResponse;
+};
+
+export type APIInterface = {
+    createDevicePOST?: (input: {
+        session: SessionContainer;
+        deviceName: string;
+        period: number;
+        skew: number;
```
rishabhpoddar, line 109: this does not come from the frontend right? So we do not need these as part of the input to this API?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK"; qr: string } | { status: "DEVICE_ALREADY_EXISTS_ERROR" }>; // TODO: Consider seperately mentioning templates vars used in qr (issuerName, secret, provisioner, etc)
```
rishabhpoddar, line 112: Please stop making PRs with TODO comments. Either solve the TODO and then make a PR, or then add it to the a todo checklist in the main issue related to this PR.

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+    verifyCodePOST?: (input: {
+        session: SessionContainer;
+        totp: string;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<
+        | { status: "OK" | "INVALID_TOTP_ERROR" | "TOTP_NOT_ENABLED_ERROR" }
```
rishabhpoddar, line 119: This API should not be called at all in case TOTP_NOT_ENABLED_ERROR. Therefore, if the rceipe function returns this, we can throw a JS error from this API - so no need to return this status code?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+        | { status: "LIMIT_REACHED_ERROR"; retryAfterMs: number }
+    >;
+    verifyDevicePOST?: (input: {
+        session: SessionContainer;
+        deviceName: string;
+        totp: string;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<
+        | { status: "OK" | "INVALID_TOTP_ERROR" | "TOTP_NOT_ENABLED_ERROR" | "UNKNOWN_DEVICE_ERROR" }
```
rishabhpoddar, line 129: This API should not be called at all in case TOTP_NOT_ENABLED_ERROR. Therefore, if the rceipe function returns this, we can throw a JS error from this API - so no need to return this status code?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+        | { status: "LIMIT_REACHED_ERROR"; retryAfterMs: number }
+    >;
+    removeDevicePOST?: (input: {
+        session: SessionContainer;
+        deviceName: string;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK" | "TOTP_NOT_ENABLED_ERROR" }>;
```
rishabhpoddar, line 137: Do not need to return TOTP_NOT_ENABLED_ERROR

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+        userId: string;
+        existingDeviceName: string;
+        newDeviceName: string;
+        userContext: any;
+    }) => Promise<{ status: "OK" | "TOTP_NOT_ENABLED_ERROR" | "DEVICE_ALREADY_EXISTS_ERROR" | "UNKNOWN_DEVICE_ERROR" }>;
+    removeDevice: (input: {
+        userId: string;
+        deviceName: string;
+        userContext: any;
+    }) => Promise<{ status: "OK" | "TOTP_NOT_ENABLED_ERROR" }>;
```
rishabhpoddar, line 84: the OK status has `didDeviceExist` as well.

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+        deviceName: string;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK" | "TOTP_NOT_ENABLED_ERROR" }>;
+    listDevicesGET?: (input: {
+        session: SessionContainer;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK" | "TOTP_NOT_ENABLED_ERROR" }>;
+};
```
rishabhpoddar, line 143: You need to add API for checking if totp is setup for a user - or can the frontend know this based on the output of listDevicesGET (if this returns TOTP_NOT_ENABLED_ERROR, then it means it's not enabled, else it is)

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,143 @@
...
+        session: SessionContainer;
+        deviceName: string;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK" | "TOTP_NOT_ENABLED_ERROR" }>;
+    listDevicesGET?: (input: {
+        session: SessionContainer;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK" | "TOTP_NOT_ENABLED_ERROR" }>;
```
rishabhpoddar, line 142: the OK status needs to have the list of devices?

rishabhpoddar (changes requested):

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,167 @@
...
+        throw new Error("Initialisation not done. Did you forget to call the SuperTokens.init function?");
+    }
+
+    static init(config: TypeInput): RecipeListFunction {
+        return (appInfo, isInServerlessEnv) => {
+            if (Recipe.instance === undefined) {
+                Recipe.instance = new Recipe(Recipe.RECIPE_ID, appInfo, isInServerlessEnv, config);
+                return Recipe.instance;
+            } else {
+                throw new Error("Passwordless recipe has already been initialised. Please check your code for bugs.");
```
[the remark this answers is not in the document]
KShivendu: done

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,167 @@
...
+            let builder = new OverrideableBuilder(APIImplementation());
+            this.apiImpl = builder.override(this.config.override.apis).build();
+        }
+    }
+
+    static getInstanceOrThrowError(): Recipe {
+        if (Recipe.instance !== undefined) {
+            return Recipe.instance;
+        }
+        throw new Error("Initialisation not done. Did you forget to call the SuperTokens.init function?");
```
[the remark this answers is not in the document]
KShivendu: done

lib/ts/recipe/totp/recipe.ts
KShivendu: done

lib/ts/recipe/totp/recipe.ts
KShivendu: done

lib/ts/recipe/totp/recipe.ts
KShivendu: done

lib/ts/recipe/totp/recipe.ts
KShivendu: done

lib/ts/recipe/totp/types.ts
KShivendu: It represents the app name that will be displayed in the Authy app. Should we just reuse the value from app info?

lib/ts/recipe/totp/types.ts
KShivendu: qr is QR code info. Frontend should render this as a QR code.

lib/ts/recipe/totp/types.ts
KShivendu: Oo this is something new for me.

lib/ts/recipe/totp/types.ts
KShivendu: Okay good point.

lib/ts/recipe/totp/types.ts
KShivendu: Yes they can know it from that. But we can still create an API.

lib/ts/recipe/totp/types.ts
KShivendu: Fixed

lib/ts/recipe/totp/types.ts
KShivendu: Can come from the frontend. But it can complicate things for the end user. So we can keep it optional?

lib/ts/recipe/totp/types.ts
KShivendu: Reusing the value from app info unless user overrides it with the config.

lib/ts/recipe/totp/types.ts
KShivendu: How about calling it `qrCode`?

lib/ts/version.ts
```diff
@@ -12,7 +12,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-export const version = "14.1.1";
+export const version = "14.1.2";
```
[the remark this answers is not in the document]
KShivendu: done

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "supertokens-node",
-    "version": "14.1.1",
+    "version": "14.1.2",
```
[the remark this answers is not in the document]
KShivendu: done

lib/ts/recipe/totp/types.ts
KShivendu: Actually adding up with MFA, I think this can go wrong.

Original effect of this was that if a user scans the QR but doesn't verify at the time of TOTP registration, they will still be prompted for a TOTP code in the next login (since `devices.length > 0` mean isFactorEnabledForUser gives `True`)

But if you think about it, if user doesn't scan the QR and closes the tab, they will be still be prompted for a TOTP code in the next login attempt. But now since they never scanned it, they don't have any way to enter a valid TOTP. 

There's no way to differentiate these two cases. So I think we should remove this param altogether or just set it to False.

lib/ts/recipe/totp/types.ts
rishabhpoddar: I see no purpose on having this come from the frontend at all. Remove this

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,168 @@
...
+    listDevicesGET?: (input: {
+        session: SessionContainer;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK"; devices: { name: string; period: number; skew: number; verified: boolean }[] }>;
+    isTotpEnabledGET?: (input: {
+        session: SessionContainer;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK"; isEnabled: boolean }>;
```
rishabhpoddar, line 156: this API not needed - since MFA will have an API for is factor setup?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,168 @@
...
+        session: SessionContainer;
+        deviceName: string;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK"; didDeviceExist: boolean }>;
+    listDevicesGET?: (input: {
+        session: SessionContainer;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK"; devices: { name: string; period: number; skew: number; verified: boolean }[] }>;
```
rishabhpoddar, line 151: this API should return TOTP_NOT_ENABLED_ERROR as well.

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,168 @@
...
+import { BaseRequest, BaseResponse } from "../../framework";
+import RecipeUserId from "../../recipeUserId";
+import { NormalisedAppinfo } from "../../types";
+import type { SessionContainer } from "../session";
+
+export type TypeInput = {
+    issuer?: string;
+    defaultSkew?: number;
+    defaultPeriod?: number;
+    getEmailOrPhoneForRecipeUserId?: GetEmailOrPhoneForRecipeUserIdFunc;
```
rishabhpoddar, line 26: since the arg in create device is called `userIdentifierInfo`, we should call this function getUserIdentifierInfoForUserId`. Also note that we do not call this function recipe user ID cause we do not want to pass it a recipe user ID, but instead the primary user ID

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    getAllCORSHeaders = (): string[] => {
+        return [];
+    };
+
+    isErrorFromThisRecipe = (err: any): err is STError => {
+        return STError.isErrorFromSuperTokens(err) && err.fromRecipe === TotpRecipe.RECIPE_ID;
+    };
+
+    getEmailOrPhoneForRecipeUserId: GetEmailOrPhoneForRecipeUserIdFunc = async (recipeUserId, userContext) => {
```
rishabhpoddar, line 195: function name should change

rishabhpoddar, line 195: the input is also a userID, not a recipe user id

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,168 @@
...
+    }) => Promise<{ status: "OK"; devices: { name: string; period: number; skew: number; verified: boolean }[] }>;
+    isTotpEnabledGET?: (input: {
+        session: SessionContainer;
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<{ status: "OK"; isEnabled: boolean }>;
+};
+
+export type GetEmailOrPhoneForRecipeUserIdFunc = (
+    recipeUserId: RecipeUserId,
```
rishabhpoddar, line 160:
```suggestion
    userId: string,
```

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,168 @@
...
+    userContext: any
+) => Promise<
+    | {
+          status: "OK";
+          info: string;
+      }
+    | { status: "EMAIL_AND_PHONE_DO_NOT_EXIST_ERROR" | "UNKNOWN_USER_ID_ERROR" }
```
rishabhpoddar, line 167:
```suggestion
    | { status: "USER_IDENTIFIER_INFO_DOES_NOT_EXIST_ERROR" | "UNKNOWN_USER_ID_ERROR" }
```

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,243 @@
...
+            };
+        }
+
+        const primaryLoginMethod = user.loginMethods.find((method) => method.recipeUserId.getAsString() === user!.id);
+
+        if (primaryLoginMethod === undefined) {
+            return {
+                status: "UNKNOWN_USER_ID_ERROR",
+            };
+        }
```
rishabhpoddar, line 217: This is not true. There can be other login methods even if the primary user's login method is removed. If this is undefined, then you can " // fallback on trying the first email or phone number" method

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,243 @@
...
+            return await listDevicesAPI(this.apiImpl, options);
+        }
+
+        return false;
+    };
+
+    handleError = async (err: STError, _: BaseRequest, res: BaseResponse): Promise<void> => {
+        if (err.fromRecipe === TotpRecipe.RECIPE_ID) {
+            if (err.type === TotpError.TOTP_NOT_ENABLED_ERROR) {
+                sendNon200ResponseWithMessage(res, "TOTP is not enabled for the user", 400); // bad req
```
rishabhpoddar, line 181: 400 is not a good resposne for this. Search for some other status code maybe

rishabhpoddar (changes requested):

lib/ts/recipe/totp/types.ts
KShivendu: good point. I agree

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,243 @@
...
+                disabled: this.apiImpl.listDevicesGET === undefined,
+                method: "get",
+                pathWithoutApiBasePath: new NormalisedURLPath(LIST_DEVICE_API),
+            },
+            {
+                id: IS_TOTP_ENABLED_API,
+                disabled: this.apiImpl.isTotpEnabledGET === undefined,
+                method: "get",
+                pathWithoutApiBasePath: new NormalisedURLPath(IS_TOTP_ENABLED_API),
+            },
```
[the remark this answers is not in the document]
KShivendu: Yes. Removed

lib/ts/recipe/totp/recipe.ts
KShivendu: 403 (Forbidden) seems better

lib/ts/recipe/totp/types.ts
KShivendu: Good point. Removed the API

lib/ts/recipe/totp/types.ts
KShivendu: done.

lib/ts/recipe/totp/types.ts
KShivendu: Removed.

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,168 @@
...
+    recipeId: string;
+    isInServerlessEnv: boolean;
+    req: BaseRequest;
+    res: BaseResponse;
+};
+
+export type APIInterface = {
+    createDevicePOST?: (input: {
+        session: SessionContainer;
+        deviceName: string;
```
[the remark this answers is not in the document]
KShivendu: Tried MFA in AWS console?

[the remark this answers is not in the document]
KShivendu: Answered on Slack. AWS Console has this.

lib/ts/recipe/totp/types.ts
KShivendu: Done.

lib/ts/recipe/totp/types.ts
rishabhpoddar: @KShivendu missed this?

lib/ts/recipe/totp/types.ts
rishabhpoddar: You said this is removed, but it's still there in the API input, but just made optional. Remove it entirely from the API input?

lib/ts/recipe/totp/types.ts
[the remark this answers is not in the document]
rishabhpoddar: In this case, the input should be optional? How will the frontend know if it should ask for the device name or not?

lib/ts/recipe/totp/types.ts
rishabhpoddar: @KShivendu missed this?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,162 @@
...
+import OverrideableBuilder from "supertokens-js-override";
+import { BaseRequest, BaseResponse } from "../../framework";
+import { NormalisedAppinfo } from "../../types";
+import type { SessionContainer } from "../session";
+
+export type TypeInput = {
+    issuer?: string;
+    defaultSkew?: number;
+    defaultPeriod?: number;
+    getEmailOrPhoneForRecipeUserId?: GetUserIdentifierInfoForUserIdFunc;
```
rishabhpoddar, line 25: missed changing this?

lib/ts/recipe/totp/types.ts
rishabhpoddar: change not made properly.

lib/ts/recipe/totp/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    static init = TotpRecipe.init;
+
+    static Error = SuperTokensError;
+
+    static createDevice(input: { userId: string; deviceName: string; skew: number; period: number }) {
+        return TotpRecipe.getInstanceOrThrowError().recipeInterfaceImpl.createDevice({
+            userContext: {},
+            ...input,
+        });
+    }
```
KShivendu, line 29: Directly pass args instead of using `input` object.

lib/ts/recipe/totp/types.ts
KShivendu: Removed from API input.

lib/ts/recipe/totp/types.ts
KShivendu: Not sure what do you mean. I removed the `isTotpEnabled` API. 

See 

- https://github.com/supertokens/supertokens-node/blob/8fde4a6a105abf945efb5e33dffa4b0f0b4362db/lib/ts/recipe/totp/api/implementation.ts#L20
- https://github.com/supertokens/supertokens-node/blob/8fde4a6a105abf945efb5e33dffa4b0f0b4362db/lib/ts/recipe/totp/types.ts#L111

lib/ts/recipe/totp/types.ts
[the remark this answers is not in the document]
KShivendu: Okay. So we have two ways:
- Always use a default value for `deviceName` column (say "TOTP Device") unless frontend passes it while calling the API.  
- Our frontend SDK can check if any device exists. If yes, ask for new device name. Otherwise, use default value (say "TOTP Device")

lib/ts/recipe/totp/types.ts
KShivendu: Done now.

lib/ts/recipe/totp/types.ts
KShivendu: https://github.com/supertokens/supertokens-node/pull/576/files#r1223923156

lib/ts/recipe/totp/types.ts
KShivendu: > Will the listDevices return 0 or 1 result? If it returns 0, and if the user creates a new device, what will happen to the older one? How will that be cleaned up?

As per the current implementation, it will return 1 if device has been created. So user will asked to enter TOTP while logging in next time. We can make listDevices so that, it also takes allowUnverifiedDevices param (or we can filter by verified status on the SDK level). But for the device cleanup part, user needs to remove the unused device on their own. We can automate this based on expiry maybe if we don't allow any unverified devices for logging in.

rishabhpoddar: I don't want to change the core. What do other providers do in this case - a user adds a device, but does not verify it, and closes the browser. Then the user signs in - do they see a totp screen? Or do they have to setup a new device?

lib/ts/recipe/totp/types.ts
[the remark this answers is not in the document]
rishabhpoddar: We can't rely on the frontend to always pass in a device name even if one device already exists. Therefore, you need to handle this case on the API as well - if a device already exists with "TOTP Device", then we need to use another name like "TOTP Device 2" or something.

Actually, we can just use "Device 1", "Device 2" etc by default.. unless the frontend passes a device name, then we use that. In case that device name already exists, we will return a `DEVICE_ALREADY_EXISTS_ERROR`

lib/ts/recipe/totp/types.ts
rishabhpoddar: 403 is a very specific error - reserved for claim validation failures. So we can't reuse that here. Instead, we should return `TOTP_NOT_ENABLED_ERROR`

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+            const { session, options, ...rest } = input;
+            const args = { ...rest, userId: session.getUserId() };
+            let response = await input.options.recipeImplementation.verifyDevice(args);
+
+            if (response.status === "TOTP_NOT_ENABLED_ERROR") {
+                throw new STError({
+                    type: STError.TOTP_NOT_ENABLED_ERROR,
+                    message: "TOTP is not enabled for this user",
+                });
+            }
```
rishabhpoddar, line 62:
```suggestion
           
```

lib/ts/recipe/totp/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import TotpRecipe from "./recipe";
+import SuperTokensError from "./error";
+
+export default class Wrapper {
+    static init = TotpRecipe.init;
+
+    static Error = SuperTokensError;
+
+    static createDevice(input: { userId: string; deviceName: string; skew: number; period: number }) {
```
rishabhpoddar, line 24:
```suggestion
    static createDevice(input: { userId: string; deviceName: string; skew?: number; period?: number }) {
```

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    config: TypeNormalisedInput;
+
+    recipeInterfaceImpl: RecipeInterface;
+
+    apiImpl: APIInterface;
+
+    isInServerlessEnv: boolean;
+
+    constructor(recipeId: string, appInfo: NormalisedAppinfo, isInServerlessEnv: boolean, config: TypeInput) {
```
rishabhpoddar, line 55:
```suggestion
    constructor(recipeId: string, appInfo: NormalisedAppinfo, isInServerlessEnv: boolean, config?: TypeInput) {
```

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,227 @@
...
+        if (TotpRecipe.instance !== undefined) {
+            return TotpRecipe.instance;
+        }
+        throw new Error("Initialisation not done. Did you forget to call the SuperTokens.init or totp.init function?");
+    }
+
+    static init(config?: TypeInput): RecipeListFunction {
+        return (appInfo, isInServerlessEnv) => {
+            if (TotpRecipe.instance === undefined) {
+                TotpRecipe.instance = new TotpRecipe(TotpRecipe.RECIPE_ID, appInfo, isInServerlessEnv, config ?? {});
```
rishabhpoddar, line 82:
```suggestion
                TotpRecipe.instance = new TotpRecipe(TotpRecipe.RECIPE_ID, appInfo, isInServerlessEnv, config);
```

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,227 @@
...
+                return TotpRecipe.instance;
+            } else {
+                throw new Error("TOTP recipe has already been initialised. Please check your code for bugs.");
+            }
+        };
+    }
+
+    static reset() {
```
rishabhpoddar, line 90: you need to call this function in the test folder - in utils.js

lib/ts/recipe/totp/recipeImplementation.ts
```diff
@@ -0,0 +1,100 @@
...
+        let result = {
+            ...input,
+        };
+        delete result.userContext;
+        return result;
+    }
+
+    return {
+        createDevice: async function (input) {
+            const { userContext, userIdentifierInfo, ...rest } = input;
```
rishabhpoddar, line 31: need to add logic for coming up with the name in case the user has not given it

lib/ts/recipe/totp/recipeImplementation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+            let issuerName = config.issuer;
+            let userIdentifier = userIdentifierInfo;
+            return {
+                status: "OK",
+                issuerName,
+                userIdentifier,
+                secret: response.secret,
+                qrCodeString: encodeURI(
+                    `otpauth://totp/${issuerName}${userIdentifier ? ":" + userIdentifier : ""}` +
```
rishabhpoddar, line 49:
```suggestion
                    `otpauth://totp/${issuerName}${userIdentifier !== undefined ? (":" + userIdentifier) : ""}` +
```

rishabhpoddar (changes requested):

lib/ts/recipe/totp/types.ts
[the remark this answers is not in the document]
KShivendu: Done.

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,227 @@
...
+            // fallback on trying the first email
+            return { info: user.emails[0], status: "OK" };
+        } else if (primaryLoginMethod.phoneNumber !== undefined) {
+            return {
+                info: primaryLoginMethod.phoneNumber,
+                status: "OK",
+            };
+        } else if (user.phoneNumbers.length > 0) {
+            return { info: user.phoneNumbers[0], status: "OK" };
+        }
```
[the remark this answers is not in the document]
KShivendu: Done.

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,227 @@
...
+            };
+        }
+
+        const primaryLoginMethod = user.loginMethods.find((method) => method.recipeUserId.getAsString() === user!.id);
+
+        if (primaryLoginMethod === undefined) {
+            return {
+                status: "UNKNOWN_USER_ID_ERROR",
+            };
+        }
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,227 @@
...
+
+        return false;
+    };
+
+    handleError = async (err: STError, _: BaseRequest, res: BaseResponse): Promise<void> => {
+        if (err.fromRecipe === TotpRecipe.RECIPE_ID) {
+            if (err.type === TotpError.TOTP_NOT_ENABLED_ERROR) {
+                sendNon200ResponseWithMessage(res, "TOTP is not enabled for the user", 403); // bad req
+            }
+        }
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/index.ts
```diff
@@ -0,0 +1,75 @@
...
+        return TotpRecipe.getInstanceOrThrowError().recipeInterfaceImpl.createDevice({
+            userContext: {},
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/api/createDevice.ts
```diff
@@ -0,0 +1,49 @@
...
+    let session = await Session.getSession(options.req, options.res);
+
+    let { deviceName } = await options.req.getJSONBody();
+
+    if (deviceName === undefined) {
+        throw new STError({
+            type: STError.BAD_INPUT_ERROR,
+            message: "Please provide deviceName",
+        });
+    }
```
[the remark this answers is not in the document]
KShivendu: done

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,160 @@
...
+    recipeId: string;
+    isInServerlessEnv: boolean;
+    req: BaseRequest;
+    res: BaseResponse;
+};
+
+export type APIInterface = {
+    createDevicePOST?: (input: {
+        session: SessionContainer;
+        deviceName: string;
```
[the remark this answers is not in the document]
KShivendu: done

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+ * under the License.
+ */
+
+import { APIInterface } from "../types";
+import STError from "../error";
+import TotpRecipe from "../recipe";
+
+export default function getAPIImplementation(): APIInterface {
+    return {
+        createDevicePOST: async function (input) {
```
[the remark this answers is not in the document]
KShivendu: TODO when we will merge with MFA.

lib/ts/recipe/totp/api/listDevices.ts
```diff
@@ -0,0 +1,38 @@
...
+import { makeDefaultUserContextFromAPI } from "../../../utils";
+
+import Session from "../../session";
+
+export default async function listDevices(apiImplementation: APIInterface, options: APIOptions): Promise<boolean> {
+    if (apiImplementation.listDevicesGET === undefined) {
+        return false;
+    }
+
+    let session = await Session.getSession(options.req, options.res);
```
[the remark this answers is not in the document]
KShivendu: Why? It should show device list only when someone is fully authenticated, right?

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,160 @@
...
+        options: APIOptions;
+        userContext: any;
+    }) => Promise<
+        | { status: "OK"; issuerName: string; secret: string; userIdentifier?: string; qrCodeString: string }
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+            const { session, options, ...rest } = input;
+            let userIdentifierInfo = undefined;
+            const emailOrPhoneInfo = await TotpRecipe.getInstanceOrThrowError().getUserIdentifierInfoForUserId(
+                session.getUserId(),
+                input.userContext
+            );
+            if (emailOrPhoneInfo.status === "OK") {
+                userIdentifierInfo = emailOrPhoneInfo.info;
+            }
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+import { APIInterface } from "../types";
+import STError from "../error";
+import TotpRecipe from "../recipe";
+
+export default function getAPIImplementation(): APIInterface {
+    return {
+        createDevicePOST: async function (input) {
+            const { session, options, ...rest } = input;
```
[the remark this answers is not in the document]
KShivendu: Removed.

lib/ts/recipe/totp/recipe.ts
KShivendu: done

lib/ts/recipe/totp/recipe.ts
KShivendu: done

lib/ts/recipe/totp/recipe.ts
KShivendu: done

lib/ts/recipe/totp/recipeImplementation.ts
KShivendu: Not here. In the API implementation, right?

lib/ts/recipe/totp/types.ts
KShivendu: We decided to remove `TOTP_NOT_ENABLED_ERROR` from the core response. So this doesn't apply anymore.

lib/ts/recipe/totp/api/verifyCode.ts
```diff
@@ -0,0 +1,49 @@
...
+import { makeDefaultUserContextFromAPI } from "../../../utils";
+
+import Session from "../../session";
+
+export default async function verifyCode(apiImplementation: APIInterface, options: APIOptions): Promise<boolean> {
+    if (apiImplementation.verifyCodePOST === undefined) {
+        return false;
+    }
+
+    let session = await Session.getSession(options.req, options.res);
```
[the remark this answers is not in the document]
KShivendu: done

lib/ts/recipe/totp/api/removeDevice.ts
```diff
@@ -0,0 +1,48 @@
...
+import { makeDefaultUserContextFromAPI } from "../../../utils";
+
+import Session from "../../session";
+
+export default async function removeDevice(apiImplementation: APIInterface, options: APIOptions): Promise<boolean> {
+    if (apiImplementation.removeDevicePOST === undefined) {
+        return false;
+    }
+
+    let session = await Session.getSession(options.req, options.res);
```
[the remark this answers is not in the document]
KShivendu: No. I think we shouldn't remove do it here. Removal of device should be possible only when you're fully logged in. Otherwise, someone can just remove all the devices after cracking the first factor.

lib/ts/recipe/totp/api/createDevice.ts
```diff
@@ -0,0 +1,49 @@
...
+import { makeDefaultUserContextFromAPI } from "../../../utils";
+
+import Session from "../../session";
+
+export default async function createDevice(apiImplementation: APIInterface, options: APIOptions): Promise<boolean> {
+    if (apiImplementation.createDevicePOST === undefined) {
+        return false;
+    }
+
+    let session = await Session.getSession(options.req, options.res);
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+                throw new STError({
+                    type: STError.TOTP_NOT_ENABLED_ERROR,
+                    message: "TOTP is not enabled for this user",
+                });
+            }
+
+            return response;
+        },
+
+        removeDevicePOST: async function (input) {
```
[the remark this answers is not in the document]
KShivendu: Instead just keeping the validators intact in `removeDevice`

lib/ts/recipe/totp/api/verifyDevice.ts
```diff
@@ -0,0 +1,56 @@
...
+import { makeDefaultUserContextFromAPI } from "../../../utils";
+
+import Session from "../../session";
+
+export default async function verifyDevice(apiImplementation: APIInterface, options: APIOptions): Promise<boolean> {
+    if (apiImplementation.verifyDevicePOST === undefined) {
+        return false;
+    }
+
+    let session = await Session.getSession(options.req, options.res);
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+            const { session, options, ...rest } = input;
+            const args = { ...rest, userId: session.getUserId() };
+            let response = await input.options.recipeImplementation.verifyCode(args);
+
+            if (response.status === "TOTP_NOT_ENABLED_ERROR") {
+                throw new STError({
+                    type: "TOTP_NOT_ENABLED_ERROR",
+                    message: "TOTP is not enabled for this user",
+                });
+            }
```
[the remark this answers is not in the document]
KShivendu: Done

lib/ts/recipe/totp/api/implementation.ts
KShivendu: Done

lib/ts/recipe/totp/index.ts
KShivendu: Done

lib/ts/recipe/totp/types.ts
KShivendu: Most providers don't allow more than one device. Only AWS does. So we used DevTools and found that in AWS devices are created when you pass a device name but those devices remain "unassigned" (equivalent to verified in our case)

These unverfied devices weren't getting deleted (atleast not instantly) and aren't shown to the users. If user creates a device with the same name, frontend calls delete device API to remove the existing one and create a new one.

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -14,82 +14,84 @@
...
+                    userContext,
                 });
+                const devices = listDevicesResponse.status === "OK" ? listDevicesResponse.devices : [];
+                existingDeviceCount = devices.length;
+                verifiedDeviceCount = devices.filter((device) => device.verified).length;
+                deviceName = `TOTP Device ${existingDeviceCount + 1}`; // Assuming no one creates a device in the same format
             }
-            return response;
+
+            if (verifiedDeviceCount > 0) {
```
rishabhpoddar, line 52: if the deviceName is provided by the user, the verifiedDeviceCount will be 0, cause you set the value to it only if `deviceName === undefined`

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -14,82 +14,84 @@
...
 
-            if (response.status === "TOTP_NOT_ENABLED_ERROR") {
-                throw new STError({
-                    type: STError.TOTP_NOT_ENABLED_ERROR,
-                    message: "TOTP is not enabled for this user",
-                });
-            }
+            // TODO: Uncomment when MFA is implemented
+            // userContext.flow = 'signin';
+            // await completeFactorInSession(session, 'totp', userContext);
```
rishabhpoddar, line 69: TODO

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -14,82 +14,84 @@
...
+            let response = await options.recipeImplementation.verifyCode(args);
 
-            if (response.status === "TOTP_NOT_ENABLED_ERROR") {
-                throw new STError({
-                    type: STError.TOTP_NOT_ENABLED_ERROR,
-                    message: "TOTP is not enabled for this user",
-                });
-            }
+            // TODO: Uncomment when MFA is implemented
+            // userContext.flow = 'signin';
```
rishabhpoddar, line 68: we cannot use userContext in our code

lib/ts/recipe/totp/recipe.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    isErrorFromThisRecipe = (err: any): err is STError => {
+        return STError.isErrorFromSuperTokens(err) && err.fromRecipe === TotpRecipe.RECIPE_ID;
+    };
+
+    getUserIdentifierInfoForUserId: GetUserIdentifierInfoForUserIdFunc = async (userId, userContext) => {
+        if (this.config.getUserIdentifierInfoForUserId !== undefined) {
+            const userRes = await this.config.getUserIdentifierInfoForUserId(userId, userContext);
+            if (userRes.status !== "UNKNOWN_USER_ID_ERROR") {
+                return userRes;
```
rishabhpoddar, line 179: need to add test to make sure that TOTP recipe can be used with a custom user ID that is not known to supertokens

lib/ts/recipe/totp/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    static createDevice(input: {
+        userId: string;
+        deviceName: string;
+        skew?: number;
+        period?: number;
+        userContext?: any;
+    }) {
+        return TotpRecipe.getInstanceOrThrowError().recipeInterfaceImpl.createDevice({
+            userContext: input.userContext === undefined ? {} : input.userContext,
+            ...input,
```
rishabhpoddar, line 33: userContext has to go after `...input` in all these functions

lib/ts/recipe/totp/api/listDevices.ts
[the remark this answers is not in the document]
rishabhpoddar: Not necessarily. This API needs to be called from the frontend in order to know which flow to pick:
- Create device flow on the frontend
- Enter TOTP flow on the frontend.

lib/ts/recipe/totp/api/removeDevice.ts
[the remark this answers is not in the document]
rishabhpoddar: Okay. But then we need to remove all other validators, and just keep the allFactorsCompleted validator.

lib/ts/recipe/totp/types.ts
rishabhpoddar: - I think we should just remove the config for `allowUnverifiedDevice` from the core alltogether (since we are anyway changing the core now).
- If devices are unverified in the core, they remain as unverified. If a new device is created with the same name as the unverified device, the core should just replace the existing unverified device with the new one.
- Finally, move the name generation strategy to the core - so the createDevice API in the core should take an optional device name, and the name will be determined based on the number of verified devices. This also means that we no longer need to have the name generation logic on the backend SDK level. Keep in mind that the core can still return device already exists if the name is given by the user, and that name belongs to a verified device already.

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,99 @@
...
+            return response;
+        },
+
+        verifyDevicePOST: async function ({ session, options, deviceName, totp, userContext }) {
+            const args = { userId: session.getUserId(), deviceName, totp, userContext };
+            let response = await options.recipeImplementation.verifyDevice(args);
+
+            // TODO: Uncomment when MFA is implemented
+            // userContext.flow = 'signup';
+            // await completeFactorInSession(session, 'totp', userContext);
```
rishabhpoddar, line 80: - TODO
- we should not use userContext in our code

lib/ts/recipe/totp/recipeImplementation.ts
```diff
@@ -0,0 +1,101 @@
...
+
+            let issuerName = config.issuer;
+            let userIdentifier = userIdentifierInfo;
+            return {
+                status: "OK",
+                issuerName,
+                userIdentifier,
+                secret: response.secret,
+                qrCodeString: encodeURI(
+                    `otpauth://totp/${issuerName}${userIdentifier === undefined ? ":" + userIdentifier : ""}` +
```
rishabhpoddar, line 50:
```suggestion
                    `otpauth://totp/${issuerName}${userIdentifier !== undefined ? ":" + userIdentifier : ""}` +
```

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,99 @@
...
+            } else if (emailOrPhoneInfo.status === "UNKNOWN_USER_ID_ERROR") {
+                throw new SessionError({
+                    type: SessionError.UNAUTHORISED,
+                    message: "Unknown User ID provided",
+                });
+            }
+
+            let existingDeviceCount = 0;
+            let verifiedDeviceCount = 0;
+            if (deviceName === undefined) {
```
rishabhpoddar, line 40: - The device name generation should happen in the core, so remove the logic here. I see no benefit of doing it in the backend SDK level.
- See last comment in https://github.com/supertokens/supertokens-node/pull/576#discussion_r1212621133

rishabhpoddar (changes requested):

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -14,82 +14,84 @@
...
-                    message: "TOTP is not enabled for this user",
+            let existingDeviceCount = 0;
+            let verifiedDeviceCount = 0;
+            if (deviceName === undefined) {
+                // We need to set the device name:
+                const listDevicesResponse = await options.recipeImplementation.listDevices({
+                    userId: session.getUserId(),
+                    userContext,
                 });
+                const devices = listDevicesResponse.status === "OK" ? listDevicesResponse.devices : [];
```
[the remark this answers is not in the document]
KShivendu: Done.

lib/ts/recipe/totp/api/implementation.ts
KShivendu: ?

lib/ts/recipe/totp/api/listDevices.ts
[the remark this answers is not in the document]
KShivendu: > which flow to pick

This will be handled by `/auth/mfa/factor/is-enabled` API which does this for TOTP:
```js
if (input.factorId === "totp") {
    const res = await listDevices({ userId: input.session.getUserId() });
    const verifiedDevicesCount = res.status === "OK" ? res.devices.filter((d) => d.verified).length : 0;
    return verifiedDevicesCount > 0;
}
```

lib/ts/recipe/totp/api/removeDevice.ts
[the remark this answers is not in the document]
KShivendu: > remove all other validators, and just keep the allFactorsCompleted validator

Hmm. Then this should happen in all TOTP APIs where we are overriding validators.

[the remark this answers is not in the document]
KShivendu: TODO

lib/ts/recipe/totp/index.ts
KShivendu: Done

lib/ts/recipe/totp/types.ts
```diff
@@ -0,0 +1,159 @@
...
+            builder: OverrideableBuilder<RecipeInterface>
+        ) => RecipeInterface;
+        apis: (originalImplementation: APIInterface, builder: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+
+export type RecipeInterface = {
+    createDevice: (input: {
+        userId: string;
+        deviceName: string;
```
rishabhpoddar, line 52: should be optional as well

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+            if (verifiedDeviceCount > 0) {
+                // TODO: We need to assert that all factors have been completed
+                // before actually creating the device.
+                // await session.assertClaims(MfaClaim.validators.hasCompletedAllFactors(), userContext);
+            }
+
+            if (deviceName === undefined) {
+                // We need to set the device name:
+                deviceName = `TOTP Device ${devices.length + 1}`; // Assuming no one creates a device in the same format
+            }
```
rishabhpoddar, line 54: this happens in the core, so we don't need it here

lib/ts/recipe/totp/api/implementation.ts
```diff
@@ -0,0 +1,97 @@
...
+                userId: session.getUserId(),
+                userContext,
+            });
+            const devices = listDevicesResponse.devices;
+            const verifiedDeviceCount = devices.filter((device) => device.verified).length;
+
+            if (verifiedDeviceCount > 0) {
+                // TODO: We need to assert that all factors have been completed
+                // before actually creating the device.
+                // await session.assertClaims(MfaClaim.validators.hasCompletedAllFactors(), userContext);
```
rishabhpoddar, line 48: this will also change to use the isAllowedToSetupFactor func

rishabhpoddar (changes requested):

PR #579, opened 2023-06-05
account linking third party

rishabhpoddar:
## TODO
- [x] in index.ts, add linkAccountsPostSession using third party info
- [x] thirdpartypasswordless changes
- [x] add checking for email change not allowed condition in `mockCreateNewOrUpdateEmailOfRecipeUser`
- [x] delete isVerified boolean check in signinup index function in all recipes related to thirdparty
- [x] make listUsersByAccountInfo in index.ts task optional boolean instead of compulsory
- [x] in linkAccountWithUserFromSession, right before we create a new recipe user, should we call isSignUpAllowed?
- [x] Should we return authoCodeResponse in linkAccountWithUserFromSession for third party API if email of new account needs to be verified? Cause if we don't, then how else will the dev get the access token? And if we do, then how will the dev save it AFTER the email has been verified?
- [x] Change field name from newEmail to email etc in linkAccountsWithSession in email verification recipe
- [x] check in test for linkAccountPostSession (functional level), that if needs to be verified is returned, then the linking info is stored in to fetch table


lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -682,6 +758,31 @@ export default class Recipe extends RecipeModule {
...
+             * of the account). If that current session is NOT that primary user, then it
+             * would have failed in the above check itself.
+             *
+             * - This email is not verified, and there exists a primary user for this email / phone
+             * number. The only way it can come here is if that primary user is the current session's
+             * user as well (else the above check would fail). So we are OK with linking even though
+             * the newUser is not verified cause they have logged into the primary user's account
+             * when calling this function proving ownership AND the emails are the same.
+             *
+             */
```
rishabhpoddar, line 784: add tests to simulate these three cases and check that linking is successful.

PR #584, opened 2023-06-11
changes to sign in flow for third party and email password to check for email verification

rishabhpoddar:
## TODO
- [x] new tests
- [x] sign in for third party recipe should also give sign in not allowed.
- [x] sign in for third partt should mark the email as verified if other linked recipes have the same email and those are verified, or if provider says that the email is verified.
- [x] sign in for email password recipe should also give sign in not allowed.
- [x] sign in for email password should mark the email as verified if other linked recipes have the same email and those are verified.
- [x] Is `EMAIL_CHANGE_NOT_ALLOWED_ERROR` an appropriate response status text for  `mockCreateNewOrUpdateEmailOfRecipeUser` cause there may be no email change.. just a regular sign in that's blocked, cause maybe that email was marked as unverified.
- [x] Make sure that the use of email verification recipe in the new functions is done by first checking that the recipe is initialised.
- [x] Address TODOs saying that that block of code should not be happening in the core.
- [x] Attempt account linking or primary user creation during sign in process as well for third party and for email password recipes
- [x] Check for if sign in is allowed should be done after marking email as verified in the sign in APIs which should happen after the credentials are verfied.
- [x] Does isEmailChangedAllowed need to take isVerfied? Can't it just query the email verification recipe?
- [x] Marking email as verified in the linkAccounts should happen in the recipe function
- [x] We need to check that the isSignInAllowed check in sign in APIs is not affected by the account linking attempt that happens in the sign in recipe function. 
- [x] What’s the diff between isEmailChangeAllowed and isSignInAllowed?
- [x] What’s the worst that could happen if we do the signInUp and isSignInAllowed returns false?


lib/ts/recipe/emailpassword/recipeImplementation.ts
```diff
@@ -191,7 +192,15 @@ export default function getRecipeInterface(
                 }
             }
             let isAccountLinkingEnabled = false;
-            if (input.email !== undefined) {
+
+            // TODO: maybe this all should not happen here at all, and should be exposed
+            // as another function in account linking recipe called isEmailChangeAllowed -
+            // cause this is a recipe level function and we want to let devs call this
+            // independently of the user calling it (via an API for example).
```
rishabhpoddar, line 199: TODO

lib/ts/recipe/thirdparty/mockCore.ts
```diff
@@ -28,22 +35,42 @@ export async function mockCreateNewOrUpdateEmailOfRecipeUser(
...
+                            }
+                        });
+                    }
                 }
             }
+        } else if (isAccountLinkingEnabled) {
+            // this means that we are signing in a recipe user id
+            // TODO: this part should not be here, and should actually be in the backend
+            // SDK - as a function like isEmailChangeAllowed and isSignUpAllowed which is
+            // only called during the sign in API.
```
rishabhpoddar, line 65: TODO

lib/ts/recipe/thirdparty/mockCore.ts
```diff
@@ -53,6 +80,50 @@ export async function mockCreateNewOrUpdateEmailOfRecipeUser(
         email: { id: email },
     });
 
+    if (response.status === "OK" && (shouldMarkInputEmailVerified || isVerified)) {
+        // TODO: this part should ideally be in the backend SDK and not in the core. In the
+        // backend SDK, this should be a part of the sign in recipe function, and createNewUser recipe function.
```
rishabhpoddar, line 85: TODO

lib/ts/recipe/accountlinking/recipeImplementation.ts
```diff
@@ -228,6 +228,7 @@ export default function getRecipeImplementation(querier: Querier, config: TypeNo
             }
 
             if (accountsLinkingResult.status === "OK" && !accountsLinkingResult.accountsAlreadyLinked) {
+                // TODO: call verifyEmailForRecipeUserIfLinkedAccountsAreVerified
```
rishabhpoddar, line 231: TODO

PR #585, opened 2023-06-12
feat: add initial interface for oauth2 recipe

porcellus:
## Summary of change

add initial interface for oauth2 recipe

## Related issues

-  

## Test Plan

N/A

## Documentation changes

N/A

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/build/recipe/oauth2/types.d.ts
+++ b/lib/build/recipe/oauth2/types.d.ts
@@ -2,6 +2,7 @@
 import { BaseRequest, BaseResponse } from "../../framework";
 import OverrideableBuilder from "supertokens-js-override";
 import { GeneralErrorResponse, JSONObject } from "../../types";
+import { SessionContainer } from "../session";
 export declare type TypeInput = {
     override?: {
         functions?: (
@@ -28,11 +29,17 @@ export declare type APIOptions = {
     req: BaseRequest;
     res: BaseResponse;
 };
-export declare type RecipeInterface = {
-    tokensPOST(input: {
-        payload?: any;
-        validitySeconds?: number;
-        useStaticSigningKey?: boolean;
+export declare type APIInterface = {
+    tokenPOST(input: {
+        clientId: string;
+        grantType: GrantType;
+        clientSecret?: string;
+        code?: string;
+        codeVerifier?: string;
+        refreshToken?: string;
+        redirectUri?: string;
+        scope: string[];
+        options: APIOptions;
         userContext: any;
     }): Promise<
         | {
@@ -46,76 +53,177 @@ export declare type RecipeInterface = {
     >;
     authorizeGET(input: {
         clientId: string;
-        responseType: string;
-        redirectUri?: string;
-        scopes: string[];
+        responseType: ResponseType[];
+        redirectUri: string;
+        scope: string[];
         state?: string;
         codeChallenge?: string;
-        codeChallengeMethod?: string;
-        prompt?: string;
+        codeChallengeMethod?: CodeChallengeMethod;
+        prompt?: OAuthPromptType;
         nonce?: string;
-        acr_values?: string;
-        maxAge?: number;
-        options: APIOptions;
-        userContext: any;
-    }): Promise<{
-        redirectUrl: string;
-    }>;
-    authorizePOST(input: {
-        clientId: string;
-        responseType: string;
-        redirectUri?: string;
-        scopes: string[];
-        state?: string;
-        codeChallenge?: string;
-        codeChallengeMethod?: string;
-        prompt?: string;
-        nonce?: string;
-        acr_values?: string;
-        maxAge?: number;
-        options: APIOptions;
-        userContext: any;
-    }): Promise<{
-        redirectUrl: string;
-    }>;
-    getRedirectUrlPOST(input: {
-        clientId: string;
-        responseType: string;
-        redirectUri?: string;
-        scopes: string[];
-        state?: string;
-        codeChallenge?: string;
-        codeChallengeMethod?: string;
-        prompt?: string;
-        nonce?: string;
-        acr_values?: string;
-        maxAge?: number;
+        session?: SessionContainer;
         options: APIOptions;
         userContext: any;
     }): Promise<
         | {
-              status: "OK";
+              redirectUrl: string;
+          }
+        | GeneralErrorResponse
+    >;
+    authorizePOST(input: {
+        clientId: string;
+        responseType: ResponseType[];
+        redirectUri: string;
+        scope: string[];
+        state?: string;
+        codeChallenge?: string;
+        codeChallengeMethod?: CodeChallengeMethod;
+        prompt?: OAuthPromptType;
+        nonce?: string;
+        session?: SessionContainer;
+        options: APIOptions;
+        userContext: any;
+    }): Promise<
+        | {
               redirectUrl: string;
           }
         | GeneralErrorResponse
     >;
     userInfoGET(input: {
+        oauth2AccessToken: string;
         options: APIOptions;
         userContext: any;
-    }): Promise<{
-        status: "OK";
-        sub: string;
-        email?: string;
-        email_verified?: boolean;
-        phoneNumber?: string;
-    }>;
+    }): Promise<
+        | JSONObject
+        | {
+              status: "UNAUTHORISED";
+              message: string;
+          }
+    >;
 };
-export declare type APIInterface = {
+export declare type RecipeInterface = {
+    getUserInfo(input: { oauth2AccessToken: string; userContext: any }): Promise<JSONObject>;
+    createAuthCode(input: {
+        clientId: string;
+        responseType: ResponseType;
+        redirectUri?: string;
+        scopes: string[];
+        codeChallenge?: string;
+        codeChallengeMethod?: CodeChallengeMethod;
+        prompt?: OAuthPromptType;
+        nonce?: string;
+        allQueryParams: string;
+        session: SessionContainer;
+        userContext: any;
+    }): Promise<
+        | {
+              status: "OK";
+              code?: string;
+              idToken?: string;
+          }
+        | {
+              status: "CLIENT_ERROR";
+              error: AuthorizationErrorCode;
+              errorDescription: string;
+              errorURI: string;
+          }
+        | {
+              status: "AUTH_ERROR";
+              error: string;
+          }
+    >;
+    createTokens(
+        input:
+            | {
+                  clientId: string;
+                  grantType: "authorization_code";
+                  clientSecret?: string;
+                  code?: string;
+                  codeVerifier?: string;
+                  redirectUri: string;
+                  scope?: string[];
+                  allQueryParams: string;
+                  userContext: any;
+              }
+            | {
+                  clientId: string;
+                  grantType: "client_credentials";
+                  clientSecret: string;
+                  scope?: string[];
+                  userContext: any;
+              }
+            | {
+                  clientId: string;
+                  grantType: "refresh_token";
+                  clientSecret?: string;
+                  refreshToken?: string;
+                  scope?: string[];
+                  userContext: any;
+              }
+    ): Promise<
+        | {
+              status: "OK";
+              tokenType: "Bearer";
+              accessToken: string;
+              refreshToken?: string;
+              idToken?: string;
+              expiresIn: number;
+          }
+        | {
+              status: "CLIENT_ERROR";
+              error: TokensErrorCode;
+          }
+    >;
+    verifyAccessToken(input: {
+        clientId: string;
+        accessToken: string;
+    }): Promise<{
+        sessionHandle?: string;
+        scopes: string[];
+        timeCreated: number;
+        lastUpdated: number;
+        timeAccessTokenExpires: number;
+        timeRefreshTokenExpires?: number;
+    }>;
+    getTokenInfo(input: {
+        clientId: string;
+        refreshToken?: string;
+        authCode?: string;
+    }): Promise<{
+        sessionHandle?: string;
+        scopes: string[];
+        timeCreated: number;
+        lastUpdated: number;
+        timeAccessTokenExpires: number;
+        timeRefreshTokenExpires?: number;
+    }>;
+    revokeToken(
+        input:
+            | {
+                  clientId: string;
+              }
+            | {
+                  sessionHandle: string;
+              }
+            | {
+                  accessToken: string;
+              }
+    ): Promise<void>;
+    revokeAuthCodes(
+        input:
+            | {
+                  clientId: string;
+              }
+            | {
+                  sessionHandle: string;
+              }
+    ): Promise<void>;
     buildAccessToken(input: {
         clientId: string;
         userId?: string;
         sessionHandle?: string;
         scopes: string[];
+        userContext: any;
     }): Promise<{
         payload?: JSONObject;
         lifetimeInSecs?: number;
@@ -125,31 +233,55 @@ export declare type APIInterface = {
         userId?: string;
         sessionHandle?: string;
         scopes: string[];
+        userContext: any;
     }): Promise<{
         payload?: JSONObject;
         lifetimeInSecs?: number;
     }>;
     getRefreshTokenLifetime(): Promise<number | undefined>;
-    buildAuthorizeRedirectUrl(input: {
-        clientId: string;
-        redirectUri: string;
-        error?: string;
-        code?: string;
-        idToken?: string;
-    }): Promise<string>;
     validateScopes(input: {
         clientId: string;
         userId?: string;
         sessionHandle?: string;
         scopes: string[];
+        userContext: any;
     }): Promise<
         | {
               status: "OK";
               grantedScopes: string[];
           }
         | {
-              status: "SCOPE_ERROR";
+              status: "INVALID_SCOPE_ERROR";
+              message: string;
+          }
+        | {
+              status: "ACCESS_DENIED_ERROR";
               message: string;
           }
     >;
+    shouldIssueRefreshToken(input: {
+        clientId: string;
+        userId?: string;
+        sessionHandle?: string;
+        scopes: string[];
+        userContext: any;
+    }): Promise<boolean>;
 };
+export declare type ResponseType = "code" | "id_token" | "token";
+export declare type CodeChallengeMethod = "S256";
+export declare type GrantType = "authorization_code" | "client_credentials" | "refresh_token" | "password";
+export declare type OAuthPromptType = "none" | "login" | "consent" | "select_account";
+export declare type AuthorizationErrorCode =
+    | "invalid_request"
+    | "unauthorized_client"
+    | "access_denied"
+    | "unsupported_response_type"
+    | "invalid_scope"
+    | "server_error"
+    | "temporarily_unavailable";
+export declare type TokensErrorCode =
+    | "invalid_request"
+    | "invalid_client"
+    | "invalid_grant"
+    | "unsupported_grant_type"
+    | "invalid_scope";
--- a/lib/build/recipe/oauth2/types.d.ts
+++ b/lib/build/recipe/oauth2/types.d.ts
@@ -1,0 +1,155 @@
+// @ts-nocheck
+import { BaseRequest, BaseResponse } from "../../framework";
+import OverrideableBuilder from "supertokens-js-override";
+import { GeneralErrorResponse, JSONObject } from "../../types";
+export declare type TypeInput = {
+    override?: {
+        functions?: (
+            originalImplementation: RecipeInterface,
+            builder?: OverrideableBuilder<RecipeInterface>
+        ) => RecipeInterface;
+        apis?: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+export declare type TypeNormalisedInput = {
+    override: {
+        functions: (
+            originalImplementation: RecipeInterface,
+            builder?: OverrideableBuilder<RecipeInterface>
+        ) => RecipeInterface;
+        apis: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+export declare type APIOptions = {
+    recipeImplementation: RecipeInterface;
+    config: TypeNormalisedInput;
+    recipeId: string;
+    isInServerlessEnv: boolean;
+    req: BaseRequest;
+    res: BaseResponse;
+};
+export declare type RecipeInterface = {
+    tokensPOST(input: {
+        payload?: any;
+        validitySeconds?: number;
+        useStaticSigningKey?: boolean;
+        userContext: any;
+    }): Promise<
+        | {
+              token_type: "Bearer";
+              access_token: string;
+              refresh_token?: string;
+              id_token?: string;
+              expires_in: number;
+          }
+        | GeneralErrorResponse
+    >;
+    authorizeGET(input: {
+        clientId: string;
+        responseType: string;
+        redirectUri?: string;
+        scopes: string[];
+        state?: string;
+        codeChallenge?: string;
+        codeChallengeMethod?: string;
+        prompt?: string;
+        nonce?: string;
+        acr_values?: string;
+        maxAge?: number;
+        options: APIOptions;
+        userContext: any;
+    }): Promise<{
+        redirectUrl: string;
+    }>;
+    authorizePOST(input: {
+        clientId: string;
+        responseType: string;
+        redirectUri?: string;
+        scopes: string[];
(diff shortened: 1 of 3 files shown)


lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+    config: TypeNormalisedInput;
+    recipeId: string;
+    isInServerlessEnv: boolean;
+    req: BaseRequest;
+    res: BaseResponse;
+};
+
+export type RecipeInterface = {
+    tokensPOST(input: {
+        payload?: any;
```
rishabhpoddar, line 51: shouldn't there be payload for access token and id token separately?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+        validitySeconds?: number;
```
rishabhpoddar, line 52: shouldn't there be validity for access token, refresh token and id token deparately?

rishabhpoddar, line 52: We should be consistent with the naming in the CDI spec vs in this interface

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+        useStaticSigningKey?: boolean;
```
rishabhpoddar, line 53: This is not added in the CDI spec. Is this mentioned in any ADR? Should i add this to the CDI spec?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+export type APIOptions = {
+    recipeImplementation: RecipeInterface;
+    config: TypeNormalisedInput;
+    recipeId: string;
+    isInServerlessEnv: boolean;
+    req: BaseRequest;
+    res: BaseResponse;
+};
+
+export type RecipeInterface = {
```
rishabhpoddar, line 49: should this be APIInterface? The funciton names have GET / POST etc in them - which are in APIInterfaces and not recipeInterfaces.

lib/ts/recipe/oauth2/types.ts
rishabhpoddar, line 51: shouldn't this function also take auth code / client secret / client id / refresh token?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+    tokensPOST(input: {
+        payload?: any;
+        validitySeconds?: number;
+        useStaticSigningKey?: boolean;
+        userContext: any;
+    }): Promise<
+        | {
+              token_type: "Bearer";
```
rishabhpoddar, line 57: we have not discussed token_type yet - is this something that the core should know about at all?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+              access_token: string;
+              refresh_token?: string;
+              id_token?: string;
+              expires_in: number;
```
rishabhpoddar, line 61: this value of expires_at is corresponding to which token? We should have a better name for it.

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+          }
+        | GeneralErrorResponse
+    >;
+    // The OpenId spec states that we have to support both POST and GET
+    authorizeGET(input: {
+        clientId: string;
+        responseType: string;
+        redirectUri?: string;
```
rishabhpoddar, line 69: always required right?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+              refresh_token?: string;
+              id_token?: string;
+              expires_in: number;
+          }
+        | GeneralErrorResponse
+    >;
+    // The OpenId spec states that we have to support both POST and GET
+    authorizeGET(input: {
+        clientId: string;
+        responseType: string;
```
rishabhpoddar, line 68: what is the meaning of responseType?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+        scopes: string[];
+        state?: string;
+
+        // PKCE
+        codeChallenge?: string;
+        codeChallengeMethod?: string;
+
+        // OpenIdConnect
+        prompt?: string;
+        nonce?: string;
```
rishabhpoddar, line 79: to discuss

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        // Not sure if we want to support these
+        acr_values?: string;
+        maxAge?: number;
```
rishabhpoddar, line 83: if we don't need to support it, then we shouldn't for now

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        options: APIOptions;
```
rishabhpoddar, line 85: missing param for `mode` (online or offline)

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+        userContext: any;
+    }): Promise<{
+        redirectUrl: string; // This will end up as a 302, encoding both error and success
```
rishabhpoddar, line 88: So the client app will query this API (or the POST one), and then this will redirect to the websiteDomain on the authorize path?

Why not just ask the client app to redirect to the website domain path directly? 

Also, if this errors out (Cause maybe client id is missing - then what is the flow after the redirection?)

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+              status: "OK";
+              redirectUrl: string;
+          }
+        | GeneralErrorResponse
+    >;
+
+    userInfoGET(input: {
+        options: APIOptions;
+        userContext: any;
+    }): Promise<{ status: "OK"; sub: string; email?: string; email_verified?: boolean; phoneNumber?: string }>; // We should define this in a way that doesn't require adding all claims here
```
rishabhpoddar, line 147: shouldn't the input to this API be access token?

rishabhpoddar, line 147: > We should define this in a way that doesn't require adding all claims here

I didn't understand

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+        acr_values?: string;
+        maxAge?: number;
+
+        options: APIOptions;
+        userContext: any;
+    }): Promise<{
+        redirectUrl: string; // This will end up as a 302, encoding both error and success
+    }>;
+    // This is intended to be used when we want to do the redirection in the FE (e.g.: header based auth)
+    getRedirectUrlPOST(input: {
```
rishabhpoddar, line 115: didn't quite understand the point of this API

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+    }): Promise<{ payload?: JSONObject; lifetimeInSecs?: number }>; // Returning undefined for these props means we use the default from Core
+    buildIdToken(input: {
+        clientId: string;
+        userId?: string;
+        sessionHandle?: string;
+        scopes: string[];
+    }): Promise<{ payload?: JSONObject; lifetimeInSecs?: number }>; // Returning undefined for these props means we use the defaults from BE SDKs
+    getRefreshTokenLifetime(): Promise<number | undefined>; // Returning undefined means we use the default from Core
+
+    buildAuthorizeRedirectUrl(input: {
```
rishabhpoddar, line 165: why is this function needed? What would be the default logic of this function?

rishabhpoddar (changes requested):

lib/ts/recipe/oauth2/types.ts
porcellus: No, we only have `Bearer` tokens. If we ever choose to implement another token type, it'll involve far more DB changes than adding a token type, so we do not need to have this concept on the core level right now.

lib/ts/recipe/oauth2/types.ts
porcellus: This name is the one set by the standard, we shouldn't change it (although we could choose to remove it). It is defined as:
> Expiration time of the Access Token in seconds since the response was generated.

lib/ts/recipe/oauth2/types.ts
porcellus: Not missing from here. Offline access is requested through the `offline_access` according to the OId specs, and I didn't see a way to specifically request it in base OAuth2. The basic idea was that we'd choose if we should issue a refresh token in the BE SDK (in some overrideable function, i.e.: `shouldIssueRefreshToken`).
We could remove the related param from CDI and rely on scopes in the core as well, but this seemed like a more flexible solution.

lib/ts/recipe/oauth2/types.ts
porcellus: I meant that we should define the result of the function in a way that:
- will not require us to change it in case we add a built-in scope that adds some info int the user info
- will make it easy to override (TS could be complaining about types if we make it too strict)
For now, I made it into `JSONObject`; it should fit any object that we can be serialize without an issue.

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,179 @@
...
+        clientId: string;
+        redirectUri: string;
+        error?: string;
+        code?: string;
+        idToken?: string;
+    }): Promise<string>;
+
+    validateScopes(input: {
```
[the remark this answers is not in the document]
porcellus: We've discussed this on the call, but:
- We return true for some built-in scopes (`supertokens.profile`, `openid`)
- Anything else is denied (not removed from the list, since it's easier to debug if it fails loudly)

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+        sessionHandle?: string;
+        scopes: string[];
+        userContext: any;
+    }): Promise<boolean>;
+};
+
+export type ResponseType = "code" | "id_token" | "token"; // We do not support "token", but it's a possible input value.
+export type CodeChallengeMethod = "S256";
+
+export type GrantType = "authorization_code" | "client_credentials" | "refresh_token" | "password"; // We do not support "password", but it's a possible input value.
```
rishabhpoddar, line 282: If we do not support password, what happens if the user gives that?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+export type APIInterface = {
+    tokenPOST(input: {
+        clientId: string;
+        grantType: GrantType;
+        clientSecret?: string;
+        code?: string;
+        codeVerifier?: string;
+        refreshToken?: string;
+        redirectUri?: string;
+        scope: string[];
```
rishabhpoddar, line 59: should it be scope or scopes?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+        | {
+              redirectUrl: string; // This will end up as a 302, encoding both error and success
+          }
+        | GeneralErrorResponse
+    >;
+    authorizePOST(input: {
+        clientId: string;
+        responseType: ResponseType[]; // add enum
+        redirectUri: string;
+        scope: string[];
```
rishabhpoddar, line 103: scope or scopes?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+        state?: string;
```
rishabhpoddar, line 104: how is this input going to be used? The core does not have this as input / output anywhere

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        // PKCE
+        codeChallenge?: string;
+        codeChallengeMethod?: CodeChallengeMethod;
+
+        // OpenIdConnect
+        prompt?: OAuthPromptType;
+        nonce?: string;
```
rishabhpoddar, line 87: instead of these, can we just use `queryString` param? Also, queryParams should be an input to the API interface as well?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
```
rishabhpoddar, line 112: replace with queryString?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+        userContext: any;
+    }): Promise<JSONObject | { status: "UNAUTHORISED"; message: string }>;
+};
+
+export type RecipeInterface = {
+    getUserInfo(input: {
+        oauth2AccessToken: string;
+
+        userContext: any;
+    }): Promise<JSONObject>;
```
rishabhpoddar, line 137: is there a core endpoint that's required for this? Or else how will this get the info?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        // PKCE
+        codeChallenge?: string;
+        codeChallengeMethod?: CodeChallengeMethod;
+
+        // OpenIdConnect
+        prompt?: OAuthPromptType;
+        nonce?: string;
+
+        allQueryParams: string;
```
rishabhpoddar, line 153: do we need prompt and nonce if we are passing in allQueryParams?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        session: SessionContainer;
```
rishabhpoddar, line 155: recipe functions don't take session - so pass in user ID / session handle. The reason is cause these function should be callable in offline mode.

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+          }
+        | { status: "CLIENT_ERROR"; error: AuthorizationErrorCode; errorDescription: string; errorURI: string } // these are redirected back to the client
+        | { status: "AUTH_ERROR"; error: string } // This gets a redirection to the auth frontend
+    >;
+
+    createTokens(
+        input:
+            | {
+                  clientId: string;
+                  grantType: "authorization_code";
```
rishabhpoddar, line 172: the core already tries to determine the grant type based on the input. I don't think this is needed in the backend sdk?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+              tokenType: "Bearer";
+              accessToken: string;
+              refreshToken?: string;
+              idToken?: string;
+              expiresIn: number;
+          }
+        | { status: "CLIENT_ERROR"; error: TokensErrorCode } // This results in a 400 with the error code added. The status specifies that this is a client error to be consistent with the createAuthCode
+    >;
+
+    verifyAccessToken(input: {
```
rishabhpoddar, line 213:
```suggestion
    verifyAccessTokenFromDatabase(input: {
```

rishabhpoddar: Just so that users know that this will query the db (as opposed to thinking that this is stateless)

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+        clientId: string;
+        accessToken: string;
+    }): Promise<{
+        sessionHandle?: string;
+        scopes: string[];
+        timeCreated: number;
+        lastUpdated: number;
+        timeAccessTokenExpires: number;
+        timeRefreshTokenExpires?: number;
```
rishabhpoddar, line 222: Add:
- queryString
- the hash values for refresh and access token
 (basically all the stuff that the table contains)

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+    }>;
+
+    getTokenInfo(input: {
+        clientId: string;
+        refreshToken?: string;
+        authCode?: string;
```
rishabhpoddar, line 228: also add sessionHandle?: string

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+    }): Promise<{
+        sessionHandle?: string;
+        scopes: string[];
+        timeCreated: number;
+        lastUpdated: number;
+        timeAccessTokenExpires: number;
+        timeRefreshTokenExpires?: number;
```
rishabhpoddar, line 235: add all the fields that the table contains

lib/ts/recipe/oauth2/types.ts
rishabhpoddar: Need to discuss this. Im not sure i understand

rishabhpoddar (changes requested):

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+              refreshToken?: string;
+              idToken?: string;
+              expiresIn: number;
+          }
+        | { status: "CLIENT_ERROR"; error: TokensErrorCode } // This results in a 400 with the error code added. The status specifies that this is a client error to be consistent with the createAuthCode
+    >;
+
+    verifyAccessToken(input: {
+        clientId: string;
+        accessToken: string;
```
rishabhpoddar, line 215: missing user context

rishabhpoddar: Make sure to add it everywhere

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+    }): Promise<{
+        sessionHandle?: string;
+        scopes: string[];
+        timeCreated: number;
+        lastUpdated: number;
+        timeAccessTokenExpires: number;
+        timeRefreshTokenExpires?: number;
+    }>;
```
rishabhpoddar, line 223: missing unauthorised status code

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+              expiresIn: number;
+          }
+        | { status: "CLIENT_ERROR"; error: TokensErrorCode } // This results in a 400 with the error code added. The status specifies that this is a client error to be consistent with the createAuthCode
+    >;
+
+    verifyAccessToken(input: {
+        clientId: string;
+        accessToken: string;
+    }): Promise<{
+        sessionHandle?: string;
```
rishabhpoddar, line 217: missing ok status code

lib/ts/recipe/oauth2/types.ts
rishabhpoddar, line 137: missing UNAUTHORISED status

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+    shouldIssueRefreshToken(input: {
+        clientId: string;
+        userId?: string;
+        sessionHandle?: string;
+        scopes: string[];
+        userContext: any;
+    }): Promise<boolean>;
+};
+
+export type ResponseType = "code" | "id_token" | "token"; // We do not support "token", but it's a possible input value.
```
rishabhpoddar, line 279: if we don't supoprt token, we should not have it - otherwise what will the recipe function do if that is passed?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+export type RecipeInterface = {
+    getUserInfo(input: {
+        oauth2AccessToken: string;
+
+        userContext: any;
+    }): Promise<JSONObject>;
+
+    createAuthCode(input: {
+        clientId: string;
+        responseType: ResponseType;
```
rishabhpoddar, line 141: this should be an array right?

rishabhpoddar (changes requested):

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,298 @@
...
+    recipeId: string;
+    isInServerlessEnv: boolean;
+    req: BaseRequest;
+    res: BaseResponse;
+};
+
+export type APIInterface = {
+    tokenPOST(input: {
+        clientId: string;
+        grantType: GrantType;
```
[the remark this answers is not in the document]
porcellus: It is.

lib/ts/recipe/oauth2/types.ts
porcellus: This maps to a query param called `scope` and I wanted to keep that consistent. Internally, we use plural everywhere because it indicates that it is an array, but I didn't want to make this change on the API level. Although, we are making a change from snake_case to camelCase anyway, so if you have a strong preference I can rename this as well.

lib/ts/recipe/oauth2/types.ts
porcellus: It's just reflected redirection url. No need to store it.

lib/build/recipe/oauth2/types.d.ts
```diff
@@ -0,0 +1,287 @@
...
+        userId?: string;
+        sessionHandle?: string;
+        scopes: string[];
+        userContext: any;
+    }): Promise<{
+        payload?: JSONObject;
+        lifetimeInSecs?: number;
+    }>;
+    getRefreshTokenLifetime(): Promise<number | undefined>;
+    validateScopes(input: {
```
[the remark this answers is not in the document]
porcellus: Yep.

lib/ts/recipe/oauth2/types.ts
porcellus: we return w/ `unsupported_grant_type`

lib/ts/recipe/oauth2/types.ts
porcellus: No, this is done in the BE SDK. Basically it uses the session handle (and the user id). The result of this basically built the same way as the id token.

lib/ts/recipe/oauth2/types.ts
porcellus: It's important to note that it could be there (at least on the API level), since it's part of the protocol. We return with `unsupported_response_type`

lib/ts/recipe/oauth2/types.ts
porcellus: This will make typing easier when overriding and we already have this information from the API and I planned to expose grant type related `createTokens` variants in the index file to make calling and typing that part easier. (`createTokensWithAuthCode`, etc.)

lib/ts/recipe/oauth2/types.ts
rishabhpoddar: I think internally, we should also call it `scope` then.

lib/ts/recipe/oauth2/types.ts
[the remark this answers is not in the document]
rishabhpoddar: Right yea. Not in the APIInterface, but in the recipe interface for creating tokens (createTokens function)?

lib/ts/recipe/oauth2/types.ts
```diff
@@ -0,0 +1,342 @@
...
+        userContext: any;
+    }): Promise<JSONObject | { status: "UNAUTHORISED_ERROR"; message: string }>;
+};
+
+export type RecipeInterface = {
+    getUserInfo(input: {
+        oauth2AccessToken: string;
+
+        userContext: any;
+    }): Promise<JSONObject | { status: "UNAUTHORISED_ERROR" }>;
```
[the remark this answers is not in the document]
rishabhpoddar: This is only in the recipe interface cause it will make consuming the response easier. The API interface function doens't need status OK

rishabhpoddar (changes requested):

lib/ts/recipe/oauth2/types.ts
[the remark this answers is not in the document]
porcellus: We could add it there, but if we do, we need something (i.e.: a callback or a default value at least) to control it when it's created through the API. 
If we want to add this, I think we should add an overrideable function (e.g.: `shouldUseStaticSigningKey`) which will return false by default. People can override this to always return true or they could decide based on the context. This flexible and covers all usecases. It's a bit more cumbersome to use when calling manually, but that should be rare and this means there is a single way to do this.

PR #599, opened 2023-06-23
refactor: edge function compatibility

porcellus:
## Summary of change

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/recipe/thirdparty/providers/apple.ts
```diff
@@ -37,45 +37,41 @@ type TypeThirdPartyProviderAppleConfig = {
...
+            .setIssuedAt()
+            .setExpirationTime("180days")
+            .setAudience("https://appleid.apple.com")
+            .setSubject(getActualClientIdFromDevelopmentClientId(clientId))
+            .sign(crypto.createPrivateKey(privateKey.replace(/\\n/g, "\n")));
     }
 
-    function get(redirectURI: string | undefined, authCodeFromRequest: string | undefined): TypeProviderGetResponse {
+    // trying to generate a client secret, in case client has not passed the values correctly
+    const clientSecretPromise = getClientSecret(
```
rishabhpoddar, line 59: in the older impl, if the config related to the secret provided by the user was invalid, then calling Apple.init({...}) itself would fail - does this still happen? (given that this is an async function now. If it doesn't happen anymore, that's OK - will the error now be thrown when we actually try and call the `await get` function?

lib/ts/recipe/thirdparty/providers/utils.ts
```diff
@@ -1,30 +1,13 @@
...
-    const client = jwksClient({
-        jwksUri,
-    });
-    function getKey(header: any, callback: any) {
-        client.getSigningKey(header.kid, function (_, key: any) {
-            var signingKey = key.publicKey || key.rsaPublicKey;
-            callback(null, signingKey);
-        });
-    }
+    const jwks = jose.createRemoteJWKSet(new URL(jwksUri));
```
rishabhpoddar, line 8: just to confirm - since this function will be called many times (each time a social auth login happens with apple), this won't end up creating multiple instances caches for their jwks right?

test/auth-react-server/utils.js
```diff
@@ -14,7 +14,7 @@
  */
 const { exec } = require("child_process");
```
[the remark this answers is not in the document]
rishabhpoddar: @iresharma will be doing this.

rishabhpoddar (changes requested):

package.json
```diff
@@ -34,33 +34,32 @@
...
     "devDependencies": {
+        "@types/inflation": "^2.0.1",
+        "body-parser": "1.20.1",
         "@hapi/hapi": "^20.2.0",
         "@koa/router": "^10.1.1",
         "@loopback/core": "2.16.2",
         "@loopback/repository": "3.7.1",
         "@loopback/rest": "9.3.0",
         "@types/aws-lambda": "8.10.77",
         "@types/co-body": "^5.1.1",
```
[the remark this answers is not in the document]
porcellus: We use `body-parser` in some tests and `@types/inflation` is necessary since we import inflation.

lib/ts/framework/utils.ts
```diff
@@ -131,62 +176,51 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(
...
-            let parser = urlencoded({ extended: true });
-            let err = await new Promise((resolve) => parser(request, new ServerResponse(request), resolve));
-            if (err !== undefined) {
-                throw new STError({
-                    type: STError.BAD_INPUT_ERROR,
-                    message: "API input error: Please make sure to pass valid url encoded form in the request body",
-                });
-            }
-        }
     }
```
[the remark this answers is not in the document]
porcellus: body-parser only parses the body, so it doesn't really do anything for "delete" and "get" requests and it's completely unrelated to `request.query` (https://expressjs.com/en/api.html#req.query)

lib/ts/recipe/thirdparty/providers/apple.ts
porcellus: yeah, it throws on the first get, which is at the start of the first sign-in process using the provider.

lib/ts/recipe/thirdparty/providers/utils.ts
porcellus: no, it creates a single instance... but then immediately discards it. which is not technically a problem (and the way it used to work), but I'll fix it to re-use the cache anyway.

rishabhpoddar (approved):

PR #604, opened 2023-06-23
feat: Introduce MFA recipe

KShivendu:
## Summary of change

Introduce MFA recipe

## Related issues

-   https://github.com/supertokens/supertokens-core/issues/554

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -740,6 +742,17 @@ export default function getAPIImplementation(): APIInterface {
               }
             | GeneralErrorResponse
         > {
+            let session = await Session.getSession(
+                options.req,
+                options.res,
+                { overrideGlobalClaimValidators: async (_) => [], sessionRequired: false },
+                userContext
+            );
```
rishabhpoddar, line 750: the api interface should take an optional session.

lib/ts/recipe/mfa/index.ts
```diff
@@ -0,0 +1,78 @@
...
+            tenantId: "public", // TODO: Pass a variable
+            userId: session.getUserId(),
+            factorId,
+            userContext: userContext === undefined ? {} : userContext,
+        });
+    }
+
+    static async getAllFactorsEnabledForUser(session: SessionContainer, userContext?: any) {
+        return await MfaRecipe.getInstanceOrThrowError().recipeInterfaceImpl.getAllFactorsEnabledForUser({
+            tenantId: "public", // TODO: Pass a variable
```
rishabhpoddar, line 46: TODOs - change hardcoded public tenant id everywhere

lib/ts/recipe/mfa/index.ts
```diff
@@ -0,0 +1,78 @@
...
+export let init = Wrapper.init;
+export let Error = Wrapper.Error;
+
+export let enableFactorForUser = Wrapper.enableFactorForUser;
+export let disableFactorForUser = Wrapper.disableFactorForUser;
+export let getAllFactorsEnabledForUser = Wrapper.getAllFactorsEnabledForUser;
+export let isFactorAlreadySetup = Wrapper.isFactorAlreadySetup;
+
+export let completeFactorInSession = Wrapper.completeFactorInSession;
+export let checkAllowedAsFirstFactor = checkAllowedAsFirstFactorUtil;
```
rishabhpoddar, line 78: why the change in name (imported has Until, but exported doesn't)?

lib/ts/recipe/mfa/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const isAllowed = allowedFirstFactors.indexOf(factorId) !== -1;
+
+    if (!isAllowed) {
+        throw new STError({
+            type: STError.BAD_INPUT_ERROR,
+            message: "The provided factor is not allowed as a first factor. Please check your recipe configuration.",
+        });
+    }
+}
```
rishabhpoddar, line 52: instead of using this function in the APIs, what you should do is (for example in email password sign in):
- before calling getSession, check if emailpassword is allowed as a first factor, by calling `MfaRecipe.getInstanceOrThrowError().recipeInterfaceImpl.getFirstFactors` directly. 
- If it is allowed, then give sessionRequired: false in getSession, else give it as true. This way, if the sign in API is called without a session, and it's not allowed as a first factor, the user will get a 401 error - which is expected (as opposed to a 400 error)

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -788,14 +801,32 @@ export default function getAPIImplementation(): APIInterface {
...
+                    userContext
+                );
+                if (linkAccountResponse.status !== "OK") {
+                    // Recipe user id OR account info are already linked with another primary user id
+                    throw new Error("Error in linking accounts. Contact support.");
+                    // return {status: "CANNOT_LINK_ACCOUNTS_ERROR"};
+                }
+            }
+
+            userContext.flow = "signin";
```
rishabhpoddar, line 827: don't modify userContext in our code

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -788,14 +801,32 @@ export default function getAPIImplementation(): APIInterface {
...
 
-            let session = await Session.createNewSession(
-                options.req,
-                options.res,
-                emailPasswordRecipeUser.recipeUserId,
-                {},
-                {},
-                userContext
-            );
+            if (session === undefined) {
```
rishabhpoddar, line 804: update the branch

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -788,14 +801,32 @@ export default function getAPIImplementation(): APIInterface {
...
+                    emailPasswordRecipeUser.recipeUserId,
+                    session.getUserId(),
+                    userContext
+                );
+                if (linkAccountResponse.status !== "OK") {
+                    // Recipe user id OR account info are already linked with another primary user id
+                    throw new Error("Error in linking accounts. Contact support.");
+                    // return {status: "CANNOT_LINK_ACCOUNTS_ERROR"};
+                }
+            }
```
rishabhpoddar, line 825: we should not do anything in the else branch.

rishabhpoddar (changes requested):

Issue #342, opened 2022-07-05
Refactor appInfo


rishabhpoddar:
- [ ] Allow multiple websiteDomains. Right now we accept just one value, which causes issues when sharing the same backend for multiple frontends (dev env, testing env)
   - We should accept an array of domains and determine the cookie settings according to that.
   - This causes issues with knowing which domain to use for email verification and reset password links - how can we fix that?
   - What if we want to set sameSite for cookie depending on the origin server? For example, if it's querying from localhost, we set it to none, else we set it to lax (if it's a same top level domain)

- [ ] The name websiteDomain is confusing for use where we have mobile apps. Maybe we can change the name to frontendDomain?

- [ ] We should accept non https / http URLs for frontend native apps.

- [ ] Issue with multiple sub domains:
   - If login is on one sub domain (auth.example.com), value of websiteDomain can be `https://auth.example.com`, but then the name websiteDomain is confusing. It has to imply that this is the URL in which the login UI is shown.
   - If showing a login page per sub domain:
      - What will be the value of websiteDomain array? We can't just hardcode all the domains since they might be dynamically generated (when a new user signs up, they will get a new sub domain automatically)
      - How will the reset password / email verification links be generated in this case?

bhumilsarvaiya:
- have two functions:
    - get host value for request: determine the apiDomain
    - get origin value for request: determine websiteDomain

- allow user to set allowed origins: function | string | string[]
If type is string:
  - It's the same as right now.

If type is string[]:
  - The default implementation of sameSite:
      - read from origin header, and if it matches any of these string[], then determine value based on origin
      - if it doesn't match or is undefined, then determine value based on all the values in string[].
  - For reset password / email verification links:
    - read from origin header, and if it matches any of these string[], then use that domain.
    - if it doesn't match or is undefined, then we use the 0th element of the string[]

If type is function (userContext) => Promise<string>:
  - The default implementation of sameSite:
    - then you just call the function and use the result to determine value of sameSite.
  - For reset password / email verification links:
    - then you just call the function and use the result to determine value of sameSite then use that domain.

bhumilsarvaiya:
### How the above solution solves the above problem:

> We should accept an array of domains and determine the cookie settings according to that.
- we accept array of allowed website domains

> This causes issues with knowing which domain to use for email verification and reset password links - how can we fix that?
- the domain for email verification and reset password links will be dependent on the origin of the incoming request
[...]

PR #706, opened 2023-10-04
feat: Refactor app info type to allow adding multiple website domains

nkshah2:
## Summary of change

- Allows for multiple frontend domains to be configured
- Adds `origin` to app info which can be used to provide a function as the frontend domain instead of `websiteDomain` which only accepts a string

## Related issues

- https://github.com/supertokens/supertokens-node/issues/342

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Unit tests
-   [ ] Update CHANGELOG


lib/ts/recipe/session/sessionClass.ts
```diff
@@ -167,7 +167,9 @@ export default class Session implements SessionContainerInterface {
                     this.accessToken,
                     this.frontToken,
                     this.helpers.config,
-                    this.reqResInfo.transferMethod
+                    this.reqResInfo.transferMethod,
+                    this.reqResInfo.req,
+                    {}
```
rishabhpoddar, line 172: why passing in {}? If the user has provided a userContext, pass that in, else create a default user context (like it's done in the apis).

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -272,15 +274,25 @@ export default class Session implements SessionContainerInterface {
...
 
-            setAccessTokenInResponse(res, this.accessToken, this.frontToken, this.helpers.config, transferMethod);
+            setAccessTokenInResponse(
+                res,
+                this.accessToken,
+                this.frontToken,
+                this.helpers.config,
+                transferMethod,
+                info.req,
+                {}
```
rishabhpoddar, line 284: why passing in {}? Pass the userContext as an optional param into this func signature (so this is not a breaking change for the SDK then). And then in this function, if the userContext is undefined, create a default one like it's done in the APIs.

Also, during impl, first make it compulsory, so that the types complain in all places where this function is called, and then make it optional

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -272,15 +274,25 @@ export default class Session implements SessionContainerInterface {
...
                 setToken(
                     this.helpers.config,
                     res,
                     "refresh",
                     this.refreshToken.token,
                     this.refreshToken.expiry,
-                    transferMethod
+                    transferMethod,
+                    info.req,
+                    {}
```
rishabhpoddar, line 295: See above comment.

lib/ts/recipe/dashboard/api/analytics.ts
```diff
@@ -71,9 +71,12 @@ export default async function analyticsPost(
         };
     }
 
-    const { apiDomain, websiteDomain, appName } = options.appInfo;
+    const { apiDomain, getOrigin: websiteDomain, appName } = options.appInfo;
     const data = {
-        websiteDomain: websiteDomain.getAsStringDangerous(),
+        websiteDomain: websiteDomain({
+            request: undefined,
+            userContext: {},
```
rishabhpoddar, line 78: create default user context here. Not an empty object. There should be no `{}` for user context in any of your changes.

lib/ts/recipe/emailverification/utils.ts
```diff
@@ -68,9 +69,15 @@ export function getEmailVerifyLink(input: {
     token: string;
     recipeId: string;
     tenantId: string;
+    userContext: any;
 }): string {
     return (
-        input.appInfo.websiteDomain.getAsStringDangerous() +
+        input.appInfo
+            .getOrigin({
+                request: getRequestFromUserContext(input.userContext),
```
rishabhpoddar, line 77: This is not the right way for us to get the request. We should not call `getRequestFromUserContext` here. You need to modify `getEmailVerifyLink` to take in a request object | undefined, and pass that in here. 

Now all the places where `getEmailVerifyLink` is being called, needs to pass in a request object (for example in our APIs). There are a few cases where we don't have the request object like in EmailVerification. createEmailVerificationLink function, or EmailVerification.sendEmailVerificationEmail, but in those cases, you can use `getRequestFromUserContext`.

lib/ts/recipe/passwordless/recipe.ts
```diff
@@ -243,7 +244,12 @@ export default class Recipe extends RecipeModule {
         const appInfo = this.getAppInfo();
 
         let magicLink =
-            appInfo.websiteDomain.getAsStringDangerous() +
+            appInfo
+                .getOrigin({
+                    request: getRequestFromUserContext(input.userContext),
```
rishabhpoddar, line 249: do not call this function here. Instead, make this function accept a req | undefined and pass that in here. All places where we are calling createMagicLink, need to then pass in the req object. If those places don't have the req object, then they can call getRequestFromUserContext

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -253,9 +254,15 @@ export function getPasswordResetLink(input: {
     token: string;
     recipeId: string;
     tenantId: string;
+    userContext: any;
 }): string {
     return (
-        input.appInfo.websiteDomain.getAsStringDangerous() +
+        input.appInfo
+            .getOrigin({
+                request: getRequestFromUserContext(input.userContext),
```
rishabhpoddar, line 262: do not call getRequestFromUserContext here. Instead, make this function take a req | undefined. All the places that call getPasswordResetLink then need to pass in the request object if they have it, or then they can call getRequestFromUserContext if they can't get the request object (which should only be the case for index function exposed to the user).

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipe.ts
```diff
@@ -65,7 +65,7 @@ export default class SessionRecipe extends RecipeModule {
         this.config = validateAndNormaliseUserInput(this, appInfo, config);
         logDebugMessage("session init: antiCsrf: " + this.config.antiCsrf);
         logDebugMessage("session init: cookieDomain: " + this.config.cookieDomain);
-        logDebugMessage("session init: cookieSameSite: " + this.config.cookieSameSite);
+        logDebugMessage("session init: cookieSameSite: " + this.config.getCookieSameSite);
```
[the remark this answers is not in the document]
nkshah2: "Are there any cases where you haven't changed the name of the variable"

Not sure what you mean

lib/ts/recipe/passwordless/recipe.ts
nkshah2: This would make it a breaking change

lib/ts/recipe/session/utils.ts
```diff
@@ -137,19 +137,20 @@ export function validateAndNormaliseUserInput(
...
 
     let cookieSameSite: (input: {
         request: BaseRequest | undefined;
         userContext: any;
     }) => "strict" | "lax" | "none" = (input: { request: BaseRequest | undefined; userContext: any }) => {
+        let protocolOfWebsiteDomain = getURLProtocol(
+            appInfo
+                .getOrigin({
+                    request: undefined,
+                    userContext: {},
```
rishabhpoddar, line 149: you need to pass in the userContext from the input....

lib/ts/recipe/session/types.ts
```diff
@@ -75,7 +75,11 @@ export type TypeInput = {
     }) => TokenTransferMethod | "any";
 
     errorHandlers?: ErrorHandlers;
-    antiCsrf?: "VIA_TOKEN" | "VIA_CUSTOM_HEADER" | "NONE";
+    antiCsrf?:
+        | "VIA_TOKEN"
+        | "VIA_CUSTOM_HEADER"
+        | "NONE"
+        | ((input: { request: BaseRequest | undefined; userContext: any }) => "VIA_CUSTOM_HEADER" | "NONE");
```
rishabhpoddar, line 82:
```suggestion
```

rishabhpoddar: There should only be `origin` which is a user facing function.

lib/ts/recipe/session/recipe.ts
```diff
@@ -63,9 +63,11 @@ export default class SessionRecipe extends RecipeModule {
     constructor(recipeId: string, appInfo: NormalisedAppinfo, isInServerlessEnv: boolean, config?: TypeInput) {
         super(recipeId, appInfo);
         this.config = validateAndNormaliseUserInput(this, appInfo, config);
-        logDebugMessage("session init: antiCsrf: " + this.config.antiCsrf);
+        logDebugMessage("session init: antiCsrf: " + this.config.antiCsrfFunctionOrString);
```
rishabhpoddar, line 66: you should not do this style of printing cause antiCsrfFunctionOrString can be a function.. check for that here and then print it if it's a string. Else print that it's a function.

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -169,7 +176,7 @@ export default class Session implements SessionContainerInterface {
                     this.helpers.config,
                     this.reqResInfo.transferMethod,
                     this.reqResInfo.req,
-                    {}
+                    makeDefaultUserContextFromAPI(this.reqResInfo.req)
```
rishabhpoddar, line 179: user Context is from the input. If that's undefined, then call the makeDefaultUserContextFromAPI

lib/ts/recipe/session/sessionClass.ts
```diff
@@ -55,7 +56,13 @@ export default class Session implements SessionContainerInterface {
             // If we instead clear the cookies only when revokeSession
             // returns true, it can cause this kind of a bug:
             // https://github.com/supertokens/supertokens-node/issues/343
-            clearSession(this.helpers.config, this.reqResInfo.res, this.reqResInfo.transferMethod);
+            clearSession(
+                this.helpers.config,
+                this.reqResInfo.res,
+                this.reqResInfo.transferMethod,
+                this.reqResInfo.req,
+                userContext
```
rishabhpoddar, line 64: check if it's undefined, and if it's not, then pass it, else call the makeDefaultUserContextFromAPI

lib/ts/recipe/dashboard/api/analytics.ts
```diff
@@ -75,7 +76,7 @@ export default async function analyticsPost(
     const data = {
         websiteDomain: websiteDomain({
             request: undefined,
-            userContext: {},
+            userContext: makeDefaultUserContextFromAPI(options.req),
```
rishabhpoddar, line 79:
```suggestion
            userContext: {},
```

lib/ts/recipe/passwordless/recipe.ts
```diff
@@ -212,11 +212,13 @@ export default class Recipe extends RecipeModule {
             | {
                   email: string;
                   tenantId: string;
+                  request: BaseRequest | undefined;
                   userContext?: any;
               }
             | {
                   phoneNumber: string;
                   tenantId: string;
+                  request: BaseRequest | undefined;
```
rishabhpoddar, line 221: this needs to be used somewhere in this function.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #708, opened 2023-10-05
feat(mfa): initial types

porcellus:
## Summary of change

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,150 @@
...
+        functions: (
+            originalImplementation: RecipeInterface,
+            builder?: OverrideableBuilder<RecipeInterface>
+        ) => RecipeInterface;
+        apis: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+
+export type RecipeInterface = {
+    getFirstFactors: (input: { tenantId: string }) => Promise<string[]>;
```
rishabhpoddar, line 96: don't need this

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,150 @@
...
+    c: Record<string, number>;
+    n: string[];
+};
+
+export type TypeInput = {
+    firstFactors?: string[];
+
+    getMFARequirementsForSession?: (
+        userId: string,
+        recipeUserId: RecipeUserId,
```
rishabhpoddar, line 50: - add an optional session object.
- add enabled factors for user.
- add completed factor for user. If a session does not exist, we get this from the recipe that is creating the session.

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,150 @@
...
+    | MFARequirement
+)[];
+
+export type MFAClaimValue = {
+    c: Record<string, number>;
+    n: string[];
+};
+
+export type TypeInput = {
+    firstFactors?: string[];
```
rishabhpoddar, line 46: We need to finalise how factor IDs works in relation to passwordless recipe. Since passwordless means a bunch of methods of logging in (otp-phone, opt-email, link-phone, link-email).

lib/ts/recipe/multifactorauth/types.ts
rishabhpoddar: Also check the last call recording to confirm if there is anything that is missed.

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export type RecipeInterface = {
+    getFirstFactors: (input: { tenantId: string }) => Promise<string[]>;
+    isAllowedToSetupFactor: (input: {
+        tenantId: string;
+        session: SessionContainerInterface;
+        factorId: string;
+        userContext: any;
+    }) => Promise<boolean>;
+    getFactorsSetupForUser: (input: { userId: string; userContext: any }) => Promise<string[]>;
```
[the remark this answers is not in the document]
porcellus: We need to come up with a different name for these two concepts then (here's what I meant here and everywhere else by setup and enabled:

- Setup: a factor is set up (I guess I could've capitalized it differently in the names) if the user can "complete" it, e.g.: totp is setup if there is a verified totp device and the factor is enabled.
- Enabled: a factor is enabled if `enableFactorForUser` or `enableFactorForTenant` was called.

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,167 @@
...
+};
+
+export type TypeNormalisedInput = {
+    firstFactors?: string[];
+
+    getGlobalMFARequirements: (
+        userId: string,
+        recipeUserId: RecipeUserId,
+        tenantId: string,
+        session: SessionContainer | undefined,
```
rishabhpoddar, line 85:
```suggestion
        session: SessionContainer,
```

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    override: {
+        functions: (
+            originalImplementation: RecipeInterface,
+            builder?: OverrideableBuilder<RecipeInterface>
+        ) => RecipeInterface;
+        apis: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+
```
rishabhpoddar, line 110: Instead of having factor id for passwordless where we have "otp-mobile" etc.. we just have it as "passwordless", and in passwordless init, we define:
```
passwordless.init({
	contact_method: “…”,
	flow_type: “..”,
	mfa: {
		contact_method: “…”
	}
})
oneOf: […]
```
- The `mfa` object doesn't have flow type cause we don't support magic link

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,167 @@
...
+export type MFAClaimValue = {
+    c: Record<string, number>;
+    n: string[];
+};
+
+export type TypeInput = {
+    firstFactors?: string[];
+
+    getGlobalMFARequirements?: (
```
rishabhpoddar, line 48:
```suggestion
    getMFARequirementsDuringAuth?: (
```

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,129 @@
...
+export type TypeNormalisedInput = {
+    firstFactors?: string[];
+
+    getMFARequirementsForAuth: (
+        session: SessionContainer | undefined,
+        factorsSetUpByTheUser: string[],
+        completedFactors: Record<string, number>,
+        userContext: any
+    ) => Promise<MFARequirementList> | MFARequirementList;
+    getMFARequirementsForFactorSetup: (
```
rishabhpoddar, line 79: this should go away

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,129 @@
...
+        ) => RecipeInterface;
+        apis: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+
+export type RecipeInterface = {
+    isAllowedToSetupFactor: (input: {
+        session: SessionContainerInterface;
+        factorId: string;
+        userContext: any;
```
rishabhpoddar, line 100: - add the output of calling `getMFARequirementsForAuth` along with the above. The variable name can be `mfaRequirementsForAuth`.
- In the index level of this function, we will call getMFARequirementsForAuth for them (it will not be an input to that function)

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,129 @@
...
+    }) => Promise<boolean>;
+    getFactorsSetupForUser: (input: { userId: string; tenantId: string; userContext: any }) => Promise<string[]>;
```
rishabhpoddar, line 102:
```suggestion
    getFactorsSetupForUser: (input: { session: SessionContainer; userContext: any }) => Promise<string[]>;
```

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,129 @@
...
+export type APIInterface = {
+    mfaInfoGET: (input: {
+        options: APIOptions;
+        session: SessionContainerInterface;
+        userContext: any;
+    }) => Promise<
+        | {
+              status: "OK";
+              factors: {
+                  canComplete: string[];
```
rishabhpoddar, line 123:
```suggestion
                  isAlreadySetup: string[];
```

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,129 @@
...
+                  canSetup: string[];
```
rishabhpoddar, line 124:
```suggestion
                  isAllowedToSetup: string[];
```

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,129 @@
...
+        functions: (
+            originalImplementation: RecipeInterface,
+            builder?: OverrideableBuilder<RecipeInterface>
+        ) => RecipeInterface;
+        apis: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+
+export type RecipeInterface = {
+    isAllowedToSetupFactor: (input: {
```
rishabhpoddar, line 97: do we need markFactorAsCompleteInSession here too? I think it was required for certain flows in the google doc.

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
rishabhpoddar, line 97: add `markFactorAsCompleteInSession` to this

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
porcellus: Isn't there a use-case where a dev wants to call this to get info about an arbitrary user? We may event want to call it like that our dashboard impl.

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,120 @@
...
+    | MFARequirement
+)[];
+
+export type MFAClaimValue = {
+    c: Record<string, number>;
+    n: string[];
+};
+
+export type TypeInput = {
+    firstFactors?: string[];
```
rishabhpoddar, line 44: Should we make the type input something like this?
```
firstFactors?: ("emailpassword" | "thirdparty" | "otp-email" | "otp-phone" | "link-email" | "link-phone" | {type: custom, id: string})[];
```

This way, users won't make a mistake in terms of our pre support factors.

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,120 @@
...
+        ) => RecipeInterface;
+        apis: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+
+export type RecipeInterface = {
+    isAllowedToSetupFactor: (input: {
+        session: SessionContainer;
+        factorId: string;
+        requirementsForAuth: MFARequirementList;
```
rishabhpoddar, line 71:
```suggestion
        mfaRequirementsForAuth: MFARequirementList;
```

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,120 @@
...
+        factorsSetUpByTheUser: string[];
+        defaultRequiredFactorsForUser: string[];
+        defaultRequiredFactorsForTenant: string[];
+        completedFactors: Record<string, number>;
```
rishabhpoddar, line 75: do we need these as inputs? I don't mind it, but last time we discussed, it wasn't a part of the input - did you think of a use case for having these?

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
porcellus: We do not strictly need them, however I've found that factorsSetUpByTheUser can be useful if someone wants to implement a logic similar to our default `isAllowed...` logic. Basically we have all of these available wherever we are calling this (since we just called `getMFARequirements...`) and since there is only a single input object ignoring extra props is not an issue.

lib/ts/recipe/multifactorauth/types.ts
porcellus: I think that's a good compromise.

porcellus: I'll make this change since I think it's fine, but note that we are breaking the pattern of userrules and EV where we accept any user id

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -53,7 +57,15 @@ export type TypeInput = {
...
-    firstFactors?: string[];
+    firstFactors?: (
+        | "emailpassword"
+        | "thirdparty"
+        | "otp-email"
+        | "otp-phone"
+        | "link-email"
+        | "link-phone"
+        | { type: "custom"; id: string }
+    )[];
```
rishabhpoddar, line 68:
```suggestion
    firstFactors?: string[];
```

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -0,0 +1,132 @@
...
+ * under the License.
+ */
+
+import { BaseRequest, BaseResponse } from "../../framework";
+import OverrideableBuilder from "supertokens-js-override";
+import { GeneralErrorResponse, User } from "../../types";
+import { SessionContainer } from "../session";
+import { SessionContainerInterface } from "../session/types";
+
+export type MFARequirement = string;
```
rishabhpoddar, line 22: make this into the         | "emailpassword"
        | "thirdparty"
        | "otp-email"
        | "otp-phone"
        | "link-email"
        | "link-phone"
        | { type: "custom"; id: string }

rishabhpoddar (changes requested):

PR #732, opened 2023-10-26
feat: added `debug` flag in the `SuperTokenConfig` in the `init()` for logging

IamMayankThakur:
## Summary of change

- Added `debug` flag in the `init()`. Debug logs will be printed if this flag is set to true.
- The old way of settings the environment variable `DEBUG=com.supertokens` will continue to work.

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/supertokens.ts
```diff
@@ -49,6 +50,9 @@ export default class SuperTokens {
     telemetryEnabled: boolean;
 
     constructor(config: TypeInput) {
+        if (config.debug) {
```
rishabhpoddar, line 53: - we also want it to work for with the env var like before (backwards compatibility)
- Doing implicit if is not a good idea. Make it `if (config.debug === true)`

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "supertokens-node",
```
rishabhpoddar, line 2: missing tests..?
- Should not log if debug is undefined or false
- Should log if debug is set to true
- Should log if env var is set

lib/ts/supertokens.ts
```diff
@@ -49,6 +50,9 @@ export default class SuperTokens {
...
+            debug.enable("com.supertokens");
```
rishabhpoddar, line 54: - bad idea to hard code this when we already have `SUPERTOKENS_DEBUG_NAMESPACE` in the logger.ts file.
- Ideally this boolean is passed to the logger.ts file and that then enables the debug logging.. we do not want to directly interact with the `debug` lib in this file.

rishabhpoddar (changes requested):

test/config.test.js
```diff
@@ -1192,6 +1194,64 @@ describe(`configTest: ${printPath("[test/config.test.js]")}`, function () {
         }
     });
 
+    it("testing that the debug mode is set", async function () {
```
rishabhpoddar, line 1197: this is not how you test if the logging is working.. you need to actually listen to the console and see if it outputs in there?

lib/ts/supertokens.ts
```diff
@@ -48,7 +48,16 @@ export default class SuperTokens {
 
     telemetryEnabled: boolean;
 
+    debugEnabled: boolean;
```
rishabhpoddar, line 51: this variable should not be required, and even if it is, it should be in the logger.ts file

lib/ts/logger.ts
```diff
@@ -23,12 +23,14 @@ const SUPERTOKENS_DEBUG_NAMESPACE = "com.supertokens";
 */
 
 function logDebugMessage(message: string) {
-    if (debug.enabled(SUPERTOKENS_DEBUG_NAMESPACE)) {
-        debug(SUPERTOKENS_DEBUG_NAMESPACE)(
-            `{t: "${new Date().toISOString()}", message: \"${message}\", file: \"${getFileLocation()}\" sdkVer: "${version}"}`
-        );
-        console.log();
-    }
+    debug(SUPERTOKENS_DEBUG_NAMESPACE)(
```
rishabhpoddar, line 26: why aren't you checking for `if (debug.enabled(SUPERTOKENS_DEBUG_NAMESPACE)) {`?

rishabhpoddar (changes requested):

lib/ts/logger.ts
IamMayankThakur: We already call `debug.enable()` when initialising the logger, so ideally this should always return `true`. Added this check, just for safety.

lib/ts/supertokens.ts
IamMayankThakur: It was only used in the tests, removed it.

test/config.test.js
IamMayankThakur: Done. overrode the `debug.log` to capture logs.

PR #753, opened 2023-12-05
fix(Fastify session ts): wrong request session injected into the handler function

kriskw1999:
### Description:

**sandbox:**
- Here you can find a codesandbox with the issue 
- NOTE: take a look at the error of Typescript in console, don't worry about the lint on the code it is an issue with code sandbox
- https://codesandbox.io/p/devbox/fastify-typescript-forked-k38hsl?file=%2Findex.ts%3A42%2C4

**Issue:**
- The `verifySession` function in the Fastify application incorrectly uses the SuperTokens' `SessionRequest` type, leading to a type mismatch.
- When using Fastify's `get` method, the expected dynamic request type from the handler function was being overshadowed by the SuperTokens type.

**Solution:**
- Proposed a more flexible approach by introducing a new generic to the `verifySession` function.
- Modified the `verifySession` function to accept a generic type parameter `T`, extending from Fastify's `OriginalFastifyRequest`.

**Technical Details:**
- **Original Code:**
  ```javascript
  // Problematic implementation
  export declare function verifySession(
      options?: VerifySessionOptions
  ): (req: SessionRequest, res: FastifyReply) => Promise<void>;
  ```

- **Proposed Solution:**
  ```javascript
  // Revised implementation with generic type
  export declare function verifySession<T extends OriginalFastifyRequest>(
      options?: VerifySessionOptions
  ): (req: SessionRequest<T>, res: FastifyReply) => Promise<void>;
  ```

**Impact:**
- Ensures correct type inference for Fastify route handlers using `verifySession`.
- Enhances type safety and compatibility with Fastify's dynamic type resolution mechanism.
- It should not change any existing code because just decorates the new type with a generic providing a default value to it

**Additional Notes:**
- please note that you cannot add this generic without changing the interface into the type because it shows an error of potentially wrong interface implementation.
- please note this is just a proposal to fix these kind of issues, feel free to propose an alternative solution!


porcellus:
The code change looks good to me. Before merging, we'd need:
- Some note added into the changelog
- Increased patch version (see the original PR template)

I'm happy to take over as well if you don't want to do those parts.

kriskw1999:
@porcellus ok now I guess that I have done all the stuff. If I missed something let me know!

PR #785, opened 2024-01-24
feat: Add getInitialSessionAuthContext to compute SessionAuth context

sasha240100:
## Summary of change

Adds `getInitialSessionAuthContext` to support computing initial `SessionAuth` context (to be used with `initialSessionAuthContext` property from  https://github.com/supertokens/supertokens-auth-react/pull/789)

## Related issues

-   https://github.com/supertokens/supertokens-auth-react/issues/790
-   https://github.com/supertokens/supertokens-auth-react/pull/789

## Test Plan

Tested locally on a NextJS implementation using supertokens. `getInitialSessionAuthContext(session)` creates an SSR session context that works like expected when passed into `initialSessionAuthContext={}` of `<SessionAuth />` component. Also covered by unit tests

## Documentation changes

TODO.

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Add to documentation

### DIFF
--- a/lib/ts/nextjs.ts
+++ b/lib/ts/nextjs.ts
@@ -237,13 +237,13 @@ export default class NextJS {
             doesSessionExist: false,
             accessTokenPayload: {},
             invalidClaims: [],
-            userId: '',
-        } as SSRSessionContextType
+            userId: "",
+        } as SSRSessionContextType;
 
         if (session) {
-            initialContext.doesSessionExist = true
-            initialContext.accessTokenPayload = await session.getAccessTokenPayload()
-            initialContext.userId = await session.getUserId()
+            initialContext.doesSessionExist = true;
+            initialContext.accessTokenPayload = await session.getAccessTokenPayload();
+            initialContext.userId = await session.getUserId();
         }
 
         return initialContext;
--- a/lib/ts/nextjs.ts
+++ b/lib/ts/nextjs.ts
@@ -231,25 +231,22 @@ export default class NextJS {
     }
 
     static async getInitialSessionAuthContext(session: SessionContainer | undefined): Promise<SSRSessionContextType> {
-        if (session) {
-            return {
-                isContextFromSSR: true,
-                loading: false,
-                doesSessionExist: true,
-                accessTokenPayload: await session.getAccessTokenPayload(),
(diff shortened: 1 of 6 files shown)


lib/ts/types.ts
```diff
@@ -113,3 +114,13 @@ export type User = {
     // the recipeUserId can be converted to string from the RecipeUserId object type.
     toJson: () => any;
 };
+
+export type SSRSessionContextType = {
+    isContextFromSSR: true;
+    loading: false;
+    doesSessionExist: boolean;
+    accessTokenPayload: any;
+    invalidClaims: ClaimValidationError[]; // TODO: Find if it's possible to compute invalidClaims on BE
```
porcellus, line 123: As mentioned in the auth-react PR, if we can't find a way to do these properly (which I don't think we can, not right now at least), the claim validation related props should be removed from this type.

lib/ts/types.ts
```diff
@@ -113,3 +114,13 @@ export type User = {
     // the recipeUserId can be converted to string from the RecipeUserId object type.
     toJson: () => any;
 };
+
+export type SSRSessionContextType = {
+    isContextFromSSR: true;
+    loading: false;
```
porcellus, line 120: These should not be required since they are static values.

porcellus (changes requested):

lib/ts/types.ts
sasha240100: But they are expected by `supertokens-auth-react`, if they will be optional - that means we would have to polyfill them inside `SessionAuth` component, and in that case I would rather remove them completely from `supertokens-node`

lib/ts/nextjs.ts
```diff
@@ -230,6 +230,24 @@ export default class NextJS {
         return result;
     }
 
+    static async getInitialSessionAuthContext(session: SessionContainer | undefined): Promise<SSRSessionContextType> {
```
rishabhpoddar, line 233: I don't think we need this function. And if we are going to keep it, why is it in the nextjs file?

rishabhpoddar (changes requested):

PR #788, opened 2024-01-31
feat: move account linking related MFA things into account linking recipe

porcellus:
## Summary of change

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -597,262 +590,60 @@ export default function getAPIImplementation(): APIInterface {
...
+                // TODO: We do not have this info, but considering this false should be OK.
+                // Also, this part will change, since we'll be checking the existence of the EP user anyway
+                // Plus the fake email things.
+                isVerified: false,
+                tenantId,
+                userContext,
+                session,
+            });
+            if (preAuthChecks.status === "SIGN_IN_UP_NOT_ALLOWED") {
+                throw new Error("This should never happen: post-auth sign in/up checks should not fail for sign up");
```
rishabhpoddar, line 609:
```suggestion
                throw new Error("This should never happen: pre-auth sign in/up checks should not fail for sign in");
```

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,280 @@
...
+    }: {
+        accountInfo: AccountInfoWithRecipeId;
+        tenantId: string;
+        factorIds: string[];
+        isSignUp: boolean;
+        isVerified: boolean;
+        session?: SessionContainerInterface;
+        userContext: UserContext;
+    }): Promise<
+        { status: "OK"; validFactorIds: string[]; isFirstFactor: boolean } | { status: "SIGN_IN_UP_NOT_ALLOWED" }
```
rishabhpoddar, line 35:
```suggestion
        { status: "OK"; validFactorIds: string[]; isFirstFactor: boolean } | { status: "SIGN_UP_NOT_ALLOWED" }
```

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -888,485 +679,90 @@ export default function getAPIImplementation(): APIInterface {
...
-                }
+            const overwriteSessionDuringSignInUp = SessionRecipe.getInstanceOrThrowError().config
+                .overwriteSessionDuringSignInUp;
 
-                const linkRes = await AccountLinking.getInstance().recipeInterfaceImpl.linkAccounts({
-                    recipeUserId: recipeUserId,
-                    primaryUserId: sessionUser.id,
+            let signUpResponse;
+            if (session === undefined || overwriteSessionDuringSignInUp) {
+                signUpResponse = await options.recipeImplementation.signUp({
```
rishabhpoddar, line 726: add a comment here saying that we have this and not call createNewRecipeUser, (even though semantically for this API we could) because users use signUp override to do things post sign up and not createNewRecipeUser override

rishabhpoddar (changes requested):

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,280 @@
...
+                    );
+                    await Session.revokeAllSessionsForUser(
+                        session.getUserId(),
+                        false,
+                        session.getTenantId(),
+                        userContext
+                    );
+
+                    // create a new session keeping the payloads from the input session
+                    respSession = await Session.createNewSession(
```
rishabhpoddar, line 240: I'm not really comfortable with doing this.. it feels weird. What do you think about these instead:
- If we reach a case where we have to link the session user to the secondary factor user, we instead return a support code; OR
- We just logout the user from their current session and so they have to sign in again.

rishabhpoddar (changes requested):

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,280 @@
...
+            ) {
+                return { status: "SIGN_IN_UP_NOT_ALLOWED" };
+            }
+        }
+
+        // If nothing failed, we return OK
+        return { status: "OK", validFactorIds, isFirstFactor: !wantToLinkSessionUser };
+    },
+
+    postAuthChecks: async function ({
```
rishabhpoddar, line 175: https://supertokens.slack.com/archives/D02B887A2MQ/p1706808245781919?thread_ts=1706804574.214459&cid=D02B887A2MQ

rishabhpoddar: The above change is required to the logic overall

rishabhpoddar (changes requested):

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+                // In all apis besides passwordless createCode, we know exactly which factor we are signing into, so in those cases,
+                // this is basically just checking if the single factor is allowed to be setup or not.
+                // For createCode (if the FE didn't pass the factor id exactly, which it should for MFA),
+                // we filter whatever is allowed. If any of them are allowed, createCode can happen.
+                // The filtered list can be used to select email templates. As an example:
+                // If the flowType for passwordless is USER_INPUT_CODE_AND_MAGIC_LINK and but only the otp-email factor is allowed to be set up
+                // then we do not want to include a link in the email.
+                for (const id of factorIds) {
+                    try {
+                        if (!factorsAlreadySetUp.includes(id)) {
```
rishabhpoddar, line 180:
```suggestion
                        if (isSignUp) {...} else { /* assert that id is in factorsAlreadySetUp */}
```

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+                            await MultiFactorAuth.assertAllowedToSetupFactorElseThrowInvalidClaimError(
+                                session!,
```
rishabhpoddar, line 182: we should take the authenticating user into account here and when we decide value of wantToLinkSessionUser

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,426 @@
...
+    ): { status: T; reason: string } {
+        if (errorCodeMap[resp.status] !== undefined) {
+            return {
+                status: errorStatus,
+                reason: errorCodeMap[resp.status]!,
+            };
+        }
+        throw new Error("Should never come here: unmapped error status");
+    },
+    preAuthChecks: async function ({
```
rishabhpoddar, line 31: split this up into smaller functions that have clear branches

rishabhpoddar (changes requested):

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -131,184 +134,208 @@ export default class Recipe extends RecipeModule {
...
-            );
-
-            if (!shouldDoAccountLinking.shouldAutomaticallyLink) {
-                logDebugMessage(
-                    "createPrimaryUserIdOrLinkAccounts not linking because shouldAutomaticallyLink is false"
-                );
-                return user;
+            if (linkRes.status === "NO_LINK") {
+                return { status: "OK", user: inputUser };
             }
```
rishabhpoddar, line 192: i really prefer if we don't introduce new status codes.. please reuse existing ones!

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -131,184 +134,208 @@ export default class Recipe extends RecipeModule {
...
+                    throw new SessionError({
+                        type: SessionError.UNAUTHORISED,
+                        message: "Session user not found",
+                    });
+                } else if (makeSessionUserPrimaryRes.status === "OK") {
+                    sessionUser = makeSessionUserPrimaryRes.user;
+                } else {
+                    // All other statuses signify that we can't make the session user primary
+                    // Which means we can't continue
+                    return { status: "NON_PRIMARY_SESSION_USER_OTHER_PRIMARY_USER" };
```
rishabhpoddar, line 237: please reuse existing status codes

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -131,184 +134,208 @@ export default class Recipe extends RecipeModule {
     }
 
     // this function returns the user ID for which the session will be created.
-    createPrimaryUserIdOrLinkAccounts = async ({
+    createPrimaryUserIdOrLinkByAccountInfoOrLinkToSessionIfProvided = async ({
```
rishabhpoddar, line 137: in this function, why can't we just call the authCheck function and then based on if isFirstFactor = true, we call `tryLinkingByAccountInfo`, else we try and link to the session?

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -131,184 +134,208 @@ export default class Recipe extends RecipeModule {
...
+            if (this.isLinked(sessionUser, inputUser)) {
+                return { status: "OK", user: sessionUser };
+            }
+        } else {
+            // If the session user isn't primary we try and make it one.
+            logDebugMessage("createPrimaryUserIdOrLinkAccounts trying to make session user primary");
+
+            // We could check here if the session user can even become a primary user, but that'd only mean one extra core call
+            // without any added benefits, since the core already checks all pre-conditions
+            if (await this.shouldBecomePrimaryUser(sessionUser, tenantId, session, userContext)) {
```
rishabhpoddar, line 220: this can also be false cause the verification status does not match the requiresVerification from the shouldDoAutomaticAccountLinking function.. in which case we want to throw an email verification claim validation error..

I would suggest that we remove all these new helper functions and just use the ones from before. It's less error prone that way

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -799,4 +844,397 @@ export default class Recipe extends RecipeModule {
...
+                        recipeUserId: newerUser.loginMethods[0].recipeUserId,
+                        userContext,
+                    });
+                    if (createPrimaryUserResult.status !== "OK") {
+                        return createPrimaryUserResult;
+                    }
+                    primaryUser = createPrimaryUserResult.user;
+                    targetUser = olderUser;
+                } else {
+                    // If this also return false, we cannot link
```
rishabhpoddar, line 1148: this could also come here cause of email verification related issues..

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -799,4 +844,397 @@ export default class Recipe extends RecipeModule {
...
+        }
+
+        // If the input user has another user (an it's not the session user) it could be linked to based on account info
+        // then we can't link it to the session user.
+        if (
+            primaryUserThatCanBeLinkedToTheInputUser !== undefined &&
+            sessionUser.id !== primaryUserThatCanBeLinkedToTheInputUser.id
+        ) {
+            return { status: "LINKING_TO_SESSION_USER_FAILED" };
+        }
```
rishabhpoddar, line 920: why do we need to do this check? When we call linkAccounts, it will fail anyway.

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -799,4 +844,397 @@ export default class Recipe extends RecipeModule {
...
+        | { status: "LINKING_TO_SESSION_USER_FAILED" }
+        | { status: "NON_PRIMARY_SESSION_USER" }
+    > {
+        logDebugMessage("tryLinkingBySession called");
+        const usersToLink = await this.getUsersThatCanBeLinkedToRecipeUser({
+            tenantId,
+            user: inputUser,
+            userContext,
+        });
+        const primaryUserThatCanBeLinkedToTheInputUser = usersToLink.primaryUser;
```
rishabhpoddar, line 904: not needed (see comment in place where primaryUserThatCanBeLinkedToTheInputUser is used)

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -799,4 +844,397 @@ export default class Recipe extends RecipeModule {
...
+
+        // If we get here, we know that the session user is primary.
+        // if both of them are primary, they cannot be linked together.
+        // we also check this outside of this function.
+        if (inputUser.isPrimaryUser) {
+            return { status: "LINKING_TO_SESSION_USER_FAILED" };
+        }
```
rishabhpoddar, line 911: why do we need to do this check? Calling linkAccounts will anyway make this fail..

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -799,4 +844,397 @@ export default class Recipe extends RecipeModule {
...
+        inputUser: User;
+        sessionUserLinkingRequiresVerification: boolean;
+        sessionUser: User;
+        authLoginMethod: LoginMethod;
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | { status: "OK"; user: User }
+        | { status: "LINKING_TO_SESSION_USER_FAILED" }
+        | { status: "NON_PRIMARY_SESSION_USER" }
```
rishabhpoddar, line 896: please reuse existing account linking status codes

rishabhpoddar (changes requested):

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,613 @@
...
+                status: "OK",
+                isFirstFactor: false,
+                createsNewLinkForTheSessionUser: false,
+                sessionUser: inputUser,
+            };
+        }
+        logDebugMessage(`checkAuthType loading session user`);
+        // We have to load the session user in order to get the account linking info
+        const sessionUserResult = await getPrimarySessionUser(session, tenantId, userContext);
+        if (sessionUserResult.status !== "OK") {
```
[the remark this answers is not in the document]
porcellus: Yeah, I guess we can do that since that also implies that the app tried to make it a primary user.
This'll slightly change how we used to work even in the case where the app doesn't want to link to the session user. I think we are OK with that as well, but that's basically why I went with first factor sign in in this case.

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    if (validFactorIds.length === 0) {
+        if (!hasSession) {
+            throw new SessionError({
+                type: SessionError.UNAUTHORISED,
+                message: "A valid session is required to authenticate with secondary factors",
+            });
+        } else {
+            return { status: "NOT_LINKING_NON_FIRST_FACTOR" }; // TODO: better name?
```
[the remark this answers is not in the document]
porcellus: We can't throw unauthorised because if that'd make the FE go into a retry loop since it does have a valid session. We need to either throw a normal error (500) or show a support code.
What this status means is:
- There is an active session (otherwise we throw unauthorised)
- For some reason this is still considered a first factor auth (e.g.: the account linking is off)
- This login method is not a valid first factor

lib/ts/recipe/multifactorauth/recipe.ts
```diff
@@ -258,4 +260,52 @@ export default class Recipe extends RecipeModule {
...
+        logDebugMessage("checkIfLinkingAllowed called");
+        let caughtSetupFactorError;
+        const mfaInfo = await getMFARelatedInfoFromSession({
+            session,
+            userContext,
+        });
+        const validFactorIds = [];
+        // In all apis besides passwordless createCode, we know exactly which factor we are signing into, so in those cases,
+        // this is basically just checking if the single factor is allowed to be setup or not.
+        // For createCode (if the FE didn't pass the factor id exactly, which it should for MFA),
```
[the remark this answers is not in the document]
porcellus: That means that we'll disallow (even for custom implementations/for future use) a case where you could use the otp_or_link flow as a secondary factor.

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,613 @@
...
+        logDebugMessage(
+            `checkAuthType session user <-> input user shouldDoAutomaticAccountLinking returned ${JSON.stringify(
+                shouldLink
+            )}`
+        );
+
+        if (shouldLink.shouldAutomaticallyLink === false) {
+            return { status: "OK", isFirstFactor: true };
+        } else {
+            return { status: "OK", isFirstFactor: false, createsNewLinkForTheSessionUser: true, sessionUser };
```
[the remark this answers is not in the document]
porcellus: that isn't needed to decide if this is supposed to be a first factor sign in or not

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -131,184 +134,208 @@ export default class Recipe extends RecipeModule {
     }
 
     // this function returns the user ID for which the session will be created.
-    createPrimaryUserIdOrLinkAccounts = async ({
+    createPrimaryUserIdOrLinkByAccountInfoOrLinkToSessionIfProvided = async ({
         tenantId,
-        user,
+        user: inputUser,
```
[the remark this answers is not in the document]
porcellus: It just feels weird to add an `inputUser` prop to the input, but I guess it's fine.

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -131,184 +134,208 @@ export default class Recipe extends RecipeModule {
...
             // This can come here if the user is using session + email verification
             // recipe with a user ID that is not known to supertokens. In this case,
             // we do not allow linking for such users.
-            return user;
-        }
-
-        if (user.isPrimaryUser) {
-            return user;
+            return inputUser;
         }
```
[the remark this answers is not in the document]
porcellus: `inputUser` cannot be undefined according to types so the compiler may think we can never hit this branch. This is a comment/code left over from the initial account linking impl I think. I'll re-check if it can be removed.

lib/ts/recipe/accountlinking/recipe.ts
porcellus: Yeah, but the point of we cannot link is that at this point the users are not primary and we can't make any of them primary, so we can't link.
The comment is explaining why we do not continue, but I'll expand it.

lib/ts/recipe/multifactorauth/recipe.ts
```diff
@@ -258,4 +260,52 @@ export default class Recipe extends RecipeModule {
         }
         return result;
     };
+
+    async checkIfLinkingAllowed(session: SessionContainerInterface, factorIds: string[], userContext: UserContext) {
```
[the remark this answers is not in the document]
porcellus: Because this is supposed to be called to check if MFA allows linking or not. Which maps to an MFA setup, but I think it's more fitting to do that mapping in the MFA recipe instead of the auth utils.

lib/ts/recipe/multifactorauth/recipe.ts
```diff
@@ -258,4 +260,52 @@ export default class Recipe extends RecipeModule {
...
+        // this is basically just checking if the single factor is allowed to be setup or not.
+        // For createCode (if the FE didn't pass the factor id exactly, which it should for MFA),
+        // we filter whatever is allowed. If any of them are allowed, createCode can happen.
+        // The filtered list can be used to select email templates. As an example:
+        // If the flowType for passwordless is USER_INPUT_CODE_AND_MAGIC_LINK and but only the otp-email factor is allowed to be set up
+        // then we do not want to include a link in the email.
+        for (const id of factorIds) {
+            logDebugMessage(`checkIfLinkingAllowed checking assertAllowedToSetupFactorElseThrowInvalidClaimError`);
+            try {
+                await this.recipeInterfaceImpl.assertAllowedToSetupFactorElseThrowInvalidClaimError({
```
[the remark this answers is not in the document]
porcellus: this function is only called if we are linking something to the session user (as shown by the function name).
We do not check for signup, this is is called whenever a new link will be created.
I think doing this based on the linking status is better since,
- this covers both sign in and up naturally
- it runs this check every time the session user would be changed.
From the MFA perspective linking is the same as factor setup. I've updated the comment where this function is called to say that the link creation is essentially the same as MFA setup.

lib/ts/authUtils.ts
```diff
@@ -0,0 +1,613 @@
...
+            }
+        } else {
+            // This means that the app doesn't want to make the session user primary
+            return { status: "NON_PRIMARY_SESSION_USER_NO_LINK" };
+        }
+    }
+    return { status: "OK", sessionUser };
+}
+
+async function checkAuthType( // TODO: better name
```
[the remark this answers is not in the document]
porcellus: That doesn't really help tell what it does. For now I've renamed it to `checkAuthTypeAndLinkingStatus` but we can discuss this over a call

rishabhpoddar (changes requested):

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -131,184 +136,129 @@ export default class Recipe extends RecipeModule {
...
+        | {
+              status: "LINKING_TO_SESSION_USER_FAILED";
+              reason:
+                  | "EMAIL_VERIFICATION_REQUIRED"
+                  | "RECIPE_USER_ID_ALREADY_LINKED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR"
+                  | "ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR";
+          }
+        | {
+              status: "NON_PRIMARY_SESSION_USER";
+              reason: "ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR";
```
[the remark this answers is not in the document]
porcellus: we could call it something like `NON_PRIMARY_SESSION_USER_WHEN_LINKING`, but this could also come from preAuth checks, where we haven't attempted to link yet. This can happen if the user called with a session, shouldDoAutomaticAccounLinking returns true + the verification status checks out but there is a conflicting primary user.

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -799,4 +850,387 @@ export default class Recipe extends RecipeModule {
...
+    }
+
+    private async tryLinkingBySession({
+        sessionUserLinkingRequiresVerification,
+        authLoginMethod,
+        inputUser,
+        sessionUser,
+        userContext,
+    }: {
+        inputUser: User;
```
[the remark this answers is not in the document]
porcellus: since this can be called by the user manually, it is not always the user we just authenticated.

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -537,12 +532,22 @@ export default function getAPIImplementation(): APIInterface {
                         // create a primary user of the new account, and if it does that, it's OK..
                         // But in most cases, it will end up linking to existing account since the
                         // email is shared.
-                        let linkedToUser = await AccountLinking.getInstance().createPrimaryUserIdOrLinkAccounts({
-                            tenantId,
-                            user: createUserResponse.user,
-                            userContext,
-                        });
-                        if (linkedToUser.id !== existingUser.id) {
+                        const linkRes = await AuthUtils.linkToSessionIfProvidedElseCreatePrimaryUserIdOrLinkByAccountInfo(
```
rishabhpoddar, line 535: this should not take into account the input session

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -885,488 +737,123 @@ export default function getAPIImplementation(): APIInterface {
...
+                    session,
                     userContext,
                 });
+            } else {
+                signUpResponse = await options.recipeImplementation.createNewRecipeUser({
+                    tenantId,
+                    email,
+                    password,
+                    userContext,
+                });
```
rishabhpoddar, line 822: we can just use signUp here instead of using createNewRecipeUser

lib/ts/recipe/passwordless/types.ts
```diff
@@ -151,28 +157,49 @@ export type RecipeInterface = {
                   userInputCode: string;
                   deviceId: string;
                   preAuthSessionId: string;
+                  createRecipeUserIfNotExists: boolean;
```
rishabhpoddar, line 160: this is not required

lib/ts/recipe/passwordless/types.ts
```diff
@@ -151,28 +157,49 @@ export type RecipeInterface = {
...
+              status: "LINKING_TO_SESSION_USER_FAILED";
+              reason:
+                  | "EMAIL_VERIFICATION_REQUIRED"
+                  | "RECIPE_USER_ID_ALREADY_LINKED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR"
+                  | "ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR";
+          }
+        | {
+              status: "NON_PRIMARY_SESSION_USER";
+              reason: "ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR";
+          }
```
rishabhpoddar, line 202: make this into one block

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -594,265 +599,112 @@ export default function getAPIImplementation(): APIInterface {
...
-                if (!isSignInAllowed) {
-                    throw new SignInError({
-                        status: "SIGN_IN_NOT_ALLOWED",
-                        reason:
-                            "Cannot sign in due to security reasons. Please try resetting your password, use a different login method or contact support. (ERR_CODE_008)",
-                    });
-                }
+            const checkCredentialsOnTenant = async (tenantId: string) => {
+                return (
+                    (await options.recipeImplementation.signIn({ email, password, session, tenantId, userContext }))
```
rishabhpoddar, line 624: create a new recipe function called checkCredentials

lib/ts/recipe/passwordless/types.ts
```diff
@@ -209,6 +243,44 @@ export type RecipeInterface = {
           }
     >;
 
+    verifyAndDeleteCode: (
+        input:
+            | {
+                  userInputCode: string;
+                  deviceId: string;
+                  preAuthSessionId: string;
+                  createRecipeUserIfNotExists: boolean;
```
rishabhpoddar, line 252: this boolean should not be here. It should always be false.

lib/ts/recipe/passwordless/types.ts
```diff
@@ -209,6 +243,44 @@ export type RecipeInterface = {
...
+                  tenantId: string;
+                  userContext: UserContext;
+              }
+            | {
+                  linkCode: string;
+                  preAuthSessionId: string;
+                  createRecipeUserIfNotExists: boolean;
```
rishabhpoddar, line 259: this boolean should not be here. It should always be false.

lib/ts/recipe/passwordless/types.ts
```diff
@@ -209,6 +243,44 @@ export type RecipeInterface = {
...
+        | {
+              status: "OK";
+              consumedDevice: {
+                  preAuthSessionId: string;
+                  failedCodeInputAttemptCount: number;
+                  email?: string;
+                  phoneNumber?: string;
+              };
+              createdNewRecipeUser?: boolean;
+              user?: User;
```
rishabhpoddar, line 273:
```suggestion
```

lib/ts/recipe/passwordless/types.ts
```diff
@@ -209,6 +243,44 @@ export type RecipeInterface = {
...
+    ) => Promise<
+        | {
+              status: "OK";
+              consumedDevice: {
+                  preAuthSessionId: string;
+                  failedCodeInputAttemptCount: number;
+                  email?: string;
+                  phoneNumber?: string;
+              };
+              createdNewRecipeUser?: boolean;
```
rishabhpoddar, line 272:
```suggestion
```

lib/ts/recipe/passwordless/types.ts
```diff
@@ -209,6 +243,44 @@ export type RecipeInterface = {
...
+              user?: User;
+              recipeUserId?: RecipeUserId;
```
rishabhpoddar, line 274:
```suggestion
```

lib/ts/recipe/passwordless/types.ts
```diff
@@ -209,6 +243,44 @@ export type RecipeInterface = {
           }
     >;
 
+    verifyAndDeleteCode: (
```
rishabhpoddar, line 246:
```suggestion
    verifyCode: (
```

rishabhpoddar: Pass a non optional flag called `deleteCode` (so this will require a core change).

We expose a `verifyAndDeleteCode` function in the index file, which calls verifyCode with deleteCode: true all the time.

rishabhpoddar (changes requested):

Issue #790, opened 2024-02-05
Failed refresh attempt should remove access token cookie


michael-pxr:
Hi,

I have a NextJS application that calls an API (a NodeJS server) for its data. The API implements the server-side SuperTokens SDK. The NextJS application only implements the client-side SDK. I.e. only the API communicates with the supertokens-core.
When the NextJS app does a server-side request to a protected API endpoint, it is possible that the access token has expired. If so, the NextJS app must communicate this to the client, such that the client can try to refresh the access token.

The NextJS server-side code communicates the 'try refresh token' message back to the client:
```ts
// pages/my-info.tsx
// Based on https://supertokens.com/docs/thirdpartyemailpassword/nextjs/session-verification/in-ssr#1-we-use-getsession-in-getserversideprops
export const getServerSideProps = async () => {
    const props = {};
    try {
        // Try to fetch protected data from the API.
        props.data = await fetchProtectedDataFromAPI();
    } catch (e) {
        if (e.status === 401) {
            // Let the client know that the access token needs to be refreshed.
            props.fromSupertokens = 'needs-refresh';
        }
    }
    return { props };
}
```

If the client needs to refresh, it calls `Session.attemptRefreshingSession` and reloads the page on success. Note that if the refresh attempt fails, it does not redirect the user to the login page. Instead, the page will render different content if the user is logged in or not.
```tsx
// pages/_app.tsx
// Based on https://supertokens.com/docs/thirdpartyemailpassword/nextjs/session-verification/in-ssr#2-doing-manual-refresh-on-the-frontend
function MyApp({ Component, pageProps }: AppProps<{fromSupertokens: string}>) {
    const [needsToRefresh, setNeedsToRefresh] = useState<boolean>(
        pageProps.fromSupertokens === 'needs-refresh',
    );
    useEffect(() => {
        async function doRefresh() {
            if (needsToRefresh) {
                if (await Session.attemptRefreshingSession()) {
                    // post session refreshing, we reload the page. This will
                    // send the new access token to the server, and then 
                    // getServerSideProps will succeed
                    location.reload()
                } else {
                    // the user's session has expired.
                    // I do not want to redirect a user to the login page.
                    setNeedsToRefresh(false);
                }
            }
        }
        doRefresh()
    }, [needsToRefresh]);
    if (needsToRefresh) {
        // in case the frontend needs to refresh, we show nothing.
        // Alternatively, you can show a spinner.
        return null
    }
    return (
[...]

PR #813, opened 2024-03-31
fix: duplicate session token issue when cookieDomain is changed

anku255:
## Summary of change

-   Added `olderCookieDomain` config option in the session recipe. This will allow users to clear cookies from older domain when the cookieDomain is changed.
-   Fixed an issue where the access token wasn't cleared if refresh token API was called without a refresh token

### On changing  `cookieSameSite` or `cookieSecure` config

Changing these values simply overwrites the existing cookies with updated values. 

### On request spike

If an authenticated API is called after cookie domain has changed and token has expired then it would call the authenticated API 3 times while refresh API 2 times to get back to the normal state. Here is a screenshot - 

image

### On decision to clear cookies from refresh api layer

Originally, we planned to remove duplicate cookies from all session functions, including `getSession`, `createSession`, and `refreshSession`. However, some JavaScript frameworks don't permit reading the response cookie header, preventing us from setting multiple cookies with the same name and removing duplicates effectively. This approach would have led to inconsistency. Therefore, we decided to clear cookies only from the refresh API endpoint.

There was also a discussion regarding clearing cookies from `getSession` / `verifySession` along with refresh endpoint API but since this would have led to inconsistent behaviour among different JS frameworks, we decided against it. The saving in API call wasn't worth the inconsistency.

## Related issues

-   https://github.com/supertokens/supertokens-node/issues/790

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 771782a80c44697e8e43d354e3547382dead7e50.  | 
|--------|--------|

### Summary:
This PR adds a new `olderCookieDomain` config option to handle cases where the `cookieDomain` is changed and fixes an issue where the access token wasn't cleared if the refresh token API was called without a refresh token.

**Key points**:
- Added `olderCookieDomain` config option in the session recipe to handle cases where the `cookieDomain` is changed.
- Fixed an issue where the access token wasn't cleared if the refresh token API was called without a refresh token.
- Updated session API implementation, cookie and header handling, and session request functions to accommodate the new changes.
- Added tests to verify the new behavior.

----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)


rishabhpoddar:
# Different cases to consider:

## Case 1
- apiDomain: api.example.com
- User sets cookieDomain to "api.example.com" (used to be <not set>)
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 2
- apiDomain: api.example.com
- User sets cookieDomain to ".example.com" (used to be <not set>)
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 3
- apiDomain: api.example.com
- User changes cookieDomain from "api.example.com" to <not set>
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 4
- apiDomain: api.example.com
- User changes cookieDomain from ".example.com" to <not set>
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 5
- apiDomain: api.example.com
- User changes cookieDomain from "api.example.com" to ".example.com
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 6
- apiDomain: api.example.com
- User changes cookieDomain from ".example.com" to "api.example.com
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 7
- apiDomain changed from example.com to api.example.com
- User changes cookieDomain from ".example.com" to <not set>
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 8
- apiDomain changed from example.com to api.example.com
- User changes cookieDomain from ".example.com" to "api.example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 9
- apiDomain changed from example.com to api.example.com
- User changes cookieDomain from "example.com" to "api.example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 10
- apiDomain changed from example.com to api.example.com
- User changes cookieDomain from <not set> to "api.example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 11
- apiDomain changed from example.com to api.example.com
- User changes cookieDomain from <not set> to ".example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 12
- apiDomain changed from api.example.com to example.com
- User changes cookieDomain from ".example.com" to <not set>
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 13
- apiDomain changed from api.example.com to example.com
- User changes cookieDomain from ".example.com" to "example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 14
- apiDomain changed from api.example.com to example.com
- User changes cookieDomain from "api.example.com" to "example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 15
- apiDomain changed from api.example.com to example.com
- User changes cookieDomain from <not set> to "example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

## Case 16
- apiDomain changed from api.example.com to example.com
- User changes cookieDomain from <not set> to ".example.com"
- Value of olderCookieDomain: TODO

### Behaviour of API call:
- CreateNewSession: TODO 
- GetSession / VerifySession: TODO
- RefreshSession: TODO

lib/ts/recipe/session/api/implementation.ts
```diff
@@ -82,3 +98,83 @@ export default function getAPIInterface(): APIInterface {
         },
     };
 }
+
+/**
```
porcellus, line 102: I think these util functions belong in `cookieAndHeaders.ts`

lib/ts/recipe/session/api/implementation.ts
```diff
@@ -82,3 +98,83 @@ export default function getAPIInterface(): APIInterface {
...
+        return didClearCookies;
+    }
+
+    const cookies = parseCookieStringFromRequestHeader(cookieString);
+
+    const tokenTypes: TokenType[] = ["access", "refresh"];
+    for (const token of tokenTypes) {
+        const cookieName = getCookieNameFromTokenType(token);
+
+        if (cookies[cookieName] && cookies[cookieName].length > 1) {
```
porcellus, line 138:
```suggestion
        if (cookies[cookieName] !== undefined && cookies[cookieName].length > 1) {
```

lib/ts/recipe/session/api/implementation.ts
```diff
@@ -82,3 +98,83 @@ export default function getAPIInterface(): APIInterface {
...
+            didClearCookies = true;
+        }
+    }
+
+    return didClearCookies;
+}
+
+// This function is required because cookies library (and most of the popular libraries in npm)
+// does not support parsing multiple cookies with the same name.
+function parseCookieStringFromRequestHeader(cookieString: string): Record<string, string[]> {
```
porcellus, line 161: not necessary to change, but since we will not be using the value we do not need to parse the header, it could be something like a regexp check

lib/ts/recipe/session/sessionRequestFunctions.ts
```diff
@@ -266,7 +266,12 @@ export async function refreshSessionInRequest({
             setCookie(config, res, LEGACY_ID_REFRESH_TOKEN_COOKIE_NAME, "", 0, "accessTokenPath", req, userContext);
         }
 
-        logDebugMessage("refreshSession: UNAUTHORISED because refresh token in request is undefined");
+        // We need to clear the access token cookie if refresh token is not found.
+        // See: https://github.com/supertokens/supertokens-node/issues/790
+        setCookie(config, res, "sAccessToken", "", 0, "accessTokenPath", req, userContext);
```
porcellus, line 271: this will also clear it if:
- allowedTransferMethod === "header"
- both refresh and access token cookies exist.

we shouldn't clear the access token cookie in that case.

porcellus (changes requested):

lib/ts/recipe/session/sessionRequestFunctions.ts
anku255: Fixed!

lib/ts/recipe/session/api/implementation.ts
anku255: This could prove useful in the future, and since I've already tested this function, I'm choosing to keep it as is.

lib/ts/recipe/session/api/implementation.ts
anku255: Done!

lib/ts/framework/hapi/framework.ts
```diff
@@ -126,17 +126,19 @@ export class HapiResponse extends BaseResponse {
     ) => {
         let now = Date.now();
 
+        const cookieOptions = {
+            isHttpOnly: httpOnly,
+            isSecure: secure,
+            path: path,
+            domain,
+            ttl: expires - now,
+            isSameSite: (sameSite === "lax" ? "Lax" : sameSite === "none" ? "None" : "Strict") as SameSitePolicy,
```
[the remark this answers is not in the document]
anku255: Didn't really work for me. 

image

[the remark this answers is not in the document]
porcellus: Right, because in this case you would have to apply it per branch e.g.: `"Lax" as const`, anyway. See [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggygEQGJQLxQEQEMDOATAMwygB9MAjfIgbgChaBjAewDsdgowsAnLAWzRQA3rShiouQgC4oACgRZgEAHQsmAd1kBKKAD4oABigB+TAA8qARmIyMFwgCYMO3JklFSFKhgA0tAL50jKzsnDz8DoIi4hJUMvKKKmqaOvpGptjesVDMbBy2lITErrnsWn6BtAQAriwMwACWrFBmslgyQrHSsIhI-jrR4qVMADYqI0wA5m3K7lp0-vStXLx887TL4XwO80A)

rishabhpoddar:
Hey @ellipsis, give me a code reivew

lib/ts/recipe/session/sessionRequestFunctions.ts
```diff
@@ -266,7 +266,15 @@ export async function refreshSessionInRequest({
...
 
-        logDebugMessage("refreshSession: UNAUTHORISED because refresh token in request is undefined");
+        // We need to clear the access token cookie if refresh token is not found, provided the allowedTransferMethod is 'cookie' or 'any'
+        // See: https://github.com/supertokens/supertokens-node/issues/790
+        if (allowedTransferMethod === "any" || allowedTransferMethod === "cookie") {
+            setCookie(config, res, "sAccessToken", "", 0, "accessTokenPath", req, userContext);
+        }
+
+        logDebugMessage(
+            "refreshSession: cleared access token and returning UNAUTHORISED because refresh token in request is undefined"
```
porcellus, line 276: This should move into the if as well.

lib/ts/recipe/session/types.ts
```diff
@@ -410,7 +412,10 @@ export type APIInterface = {
      */
     refreshPOST:
         | undefined
-        | ((input: { options: APIOptions; userContext: UserContext }) => Promise<SessionContainerInterface>);
+        | ((input: {
+              options: APIOptions;
+              userContext: UserContext;
+          }) => Promise<SessionContainerInterface | undefined>);
```
porcellus, line 418: I'm not sure if this isn't breaking in some cases, but it definitely doesn't feel like a patch level change.

porcellus (changes requested):

porcellus:
Please explore/define the behaviors if:
- The api used to run on [api.example.com](http://api.example.com/)
- Now they are introducing [api2.example.com](http://api2.example.com/)
- This should use/get the same session, so they update the cookieDomain unset to [.example.com](http://example.com/)
- Plus setting their olderCookieDomain to "" (as I think that'd be the right thing to do.
With this setup consider:
- The user has an active session with cookies set to api.example.com, that is valid for a long time
- They make a call to api2.example.com: they get a 401, so a refresh is called on api.example.com
- This sets up the new cookies with the new cookieDomain
What happens then to requests made to `api.example.com`?
Can this cause any issues? (Token theft detection is not triggered, however other strange things could happen)

porcellus:
Please consider what happens if the user calls `mergeIntoAccessTokenPayload` (or tries to verify their email) in a case similar to Case 5 above.

lib/ts/recipe/session/sessionRequestFunctions.ts
anku255: Fixed

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -205,3 +205,83 @@ export function setCookie(
...
+            didClearCookies = true;
+        }
+    }
+
+    return didClearCookies;
+}
+
+// This function is required because cookies library (and most of the popular libraries in npm)
+// does not support parsing multiple cookies with the same name.
+function parseCookieStringFromRequestHeader(cookieString: string): Record<string, string[]> {
```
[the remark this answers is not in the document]
anku255: Updated!

anku255:
@porcellus Thanks for coming up with these cases.

After testing these cases, I have found two breaking cases:

1. After the cookieDomain change, the updated token payload value won't be used until the older token expires. If the token payload doesn't change then its not an issue but if something like mergeIntoAccessTokenPayload is called then the updated value won't be used until the older token expires (and ultimately gets cleared by refresh call).

2. If there are > 1 session cookies and the olderCookieDomain is set, we will clear the cookies for the older domain and return 200. If the olderCookieDomain is incorrect then it will cause an infinite loop. This can only be fixed by setting olderCookieDomain to the correct value.


NOTE:

I tested the case with multiple API domains. There are no additional issue with this setup. The issue arises when `api2.example.com` updates the token payload. `api.example.com` will still use the old token payload until the token expires while `api2.example.com` will use the updated token payload.

CHANGELOG.md
```diff
@@ -7,11 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
-## [17.0.4] - 2024-04-09
+## [18.0.0] - 2024-04-17
```
porcellus, line 10: Please add a note about verifySession returning 401 if it detects multiple access tokens

lib/build/recipe/session/cookieAndHeaders.js
```diff
@@ -184,17 +184,11 @@ exports.getAuthModeFromHeader = getAuthModeFromHeader;
  */
 function clearSessionCookiesFromOlderCookieDomain({ req, res, config, userContext }) {
     let didClearCookies = false;
-    const cookieString = req.getHeaderValue("cookie");
-    if (cookieString === undefined || config.olderCookieDomain === undefined) {
```
porcellus: How does this handle if `olderCookieDomain` is not defined? (As far as I can see that check has been removed and we just set the cookie with an undefined domain?)

test/session.test.js
```diff
@@ -185,6 +185,117 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
         assert.strictEqual(frontendInfo.up["rsub"], "testuserid");
     });
 
+    // check if access token cookie is cleared if refresh token api is called without the refresh token
```
porcellus, line 188: Please add a test to see what happens if there are multiple cookies, but `olderCookieDomain` is undefined

test/session.test.js
```diff
@@ -185,6 +185,117 @@ describe(`session: ${printPath("[test/session.test.js]")}`, function () {
...
+                })
+        );
+        let cookies = extractInfoFromResponse(res2);
+        assert(res2.status === 401);
+        assert(cookies.accessToken === "");
+        assert(cookies.accessTokenExpiry === new Date(0).toUTCString());
+        assert(cookies.refreshToken === undefined);
+    });
+
+    // check if access and refresh token for olderCookieDomain is cleared if multiple tokens are passed to the refresh endpoint
```
porcellus, line 247: No need to repeat the test title

lib/ts/recipe/session/api/implementation.ts
```diff
@@ -13,7 +14,20 @@ export default function getAPIInterface(): APIInterface {
         }: {
             options: APIOptions;
             userContext: UserContext;
-        }): Promise<SessionContainerInterface> {
+        }): Promise<SessionContainerInterface | undefined> {
```
porcellus, line 17: I think refresh should throw if we detect duplicate tokens but `olderCookieDomain` is undefined

porcellus (changes requested):

lib/build/recipe/session/cookieAndHeaders.js
anku255: Good catch. Missed this while refactoring. 

We will be returning a 500 response from the refreshPOST API, but yeah this function should return early if `olderCookieDomain` is undefined.

lib/ts/recipe/session/api/implementation.ts
anku255: Yeah, I agreed.  Updated the PR to address this.

test/session.test.js
anku255: sure.

test/session.test.js
anku255: done

CHANGELOG.md
anku255: done

lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -205,3 +206,97 @@ export function setCookie(
...
+    }
+
+    if (didClearCookies) {
+        throw new SessionError({
+            message:
+                "The request contains multiple session cookies. We are clearing the cookie from olderCookieDomain. Session will be refreshed in the next refresh call.",
+            type: SessionError.CLEAR_DUPLICATE_SESSION_COOKIES,
+        });
+    }
+    return didClearCookies;
```
porcellus, line 266: We do not need to return anything here, since it can only be false and we ignore the return value anyway.

lib/ts/recipe/session/api/implementation.ts
```diff
@@ -14,6 +15,27 @@ export default function getAPIInterface(): APIInterface {
...
+            // we throw a 500 error asking the user to set 'olderCookieDomain'.
+            if (
+                (hasMultipleCookiesForTokenType(options.req, "access") ||
+                    hasMultipleCookiesForTokenType(options.req, "refresh")) &&
+                options.config.olderCookieDomain === undefined
+            ) {
+                throw new Error(
+                    `The request contains multiple session cookies. This may happen if you've changed the 'cookieDomain' setting in your configuration. To clear tokens from the previous domain, set 'olderCookieDomain' in your config.`
+                );
+            }
```
porcellus, line 30: What do you think about moving this into `clearSessionCookiesFromOlderCookieDomain`? My only reason for it is that then we can't forget this check in the future if `clearSessionCookiesFromOlderCookieDomain` is called from multiple places.

porcellus (changes requested):

lib/ts/recipe/session/types.ts
```diff
@@ -410,7 +415,10 @@ export type APIInterface = {
...
```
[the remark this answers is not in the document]
anku255: Fixed. Thanks for catching it.

lib/ts/recipe/session/cookieAndHeaders.ts
anku255: Agreed. Fixed it.

lib/ts/recipe/session/api/implementation.ts
anku255: Yeah that makes more sense. I have updated the PR.

porcellus (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Rationale
+
+This update addresses an edge case where changing the `cookieDomain` config on the server can lead to session integrity issues. For instance, if the API server URL is 'api.example.com' with a cookie domain of '.example.com', and the server updates the cookie domain to 'api.example.com', the client may retain cookies with both '.example.com' and 'api.example.com' domains.
+
+Previously, verifySession would select one of the access tokens from the incoming request. If it chose the older cookie, it would return a 401 status code, prompting a refresh request. However, the `refreshPOST` API would then set new session token cookies with the updated `cookieDomain`, but older cookies will persist, leading to repeated 401 errors and refresh loops.
+
+With this update, verifySession will return a 401 error if it detects multiple access tokens in the request, prompting a refresh request. The `refreshPOST` API will clear cookies from the old domain if `olderCookieDomain` is specified in the configuration, then return a 200 status. If `olderCookieDomain` is not configured, the `refreshPOST` API will return a 500 error with a message instructing to set `olderCookieDomain`.
```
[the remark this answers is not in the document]
anku255: Done!

porcellus (approved):

PR #835, opened 2024-05-21
fix: Add workaround for unsupported 'cache' field in Cloudflare Workers

anku255:
## Summary of change

Add workaround for unsupported 'cache' field in Cloudflare Workers

If a fetch request fails due to the 'cache' field not being implemented in Cloudflare Workers, retry the request without the 'cache' field. More info: https://github.com/cloudflare/workerd/issues/698

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


examples/cloudflare-workers/with-emailpassword/README.md
```diff
@@ -0,0 +1,9 @@
+![SuperTokens banner](https://raw.githubusercontent.com/supertokens/supertokens-logo/master/images/Artboard%20%E2%80%93%2027%402x.png)
+
+# SuperTokens EmailPassword Demo app
```
rishabhpoddar, line 3: incorrect readme content

examples/cloudflare-workers/with-emailpassword/test/basic.test.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+/*
+ * Imports
+ */
+
+const assert = require("assert");
+const puppeteer = require("puppeteer");
+const { getTestEmail, setInputValues, submitForm, toggleSignInSignUp } = require("../../../../test/exampleTestHelpers");
+
```
rishabhpoddar, line 23: does this test actually run in a cloudflare worker setup>

rishabhpoddar (changes requested):

ngoquoc:
Not sure if it's because of recent changes in Cloudflare runtime or not, but the error returned since couple of days ago:
```
Error: The 'cache' field on 'RequestInitializerDict' is not implemented.
```

It seems that the error is not thrown where Supertoken catches as I debugged and it could never reach `catch` block. I tried the following and it works:
```
const fetchFunction = typeof fetch !== "undefined" ? fetch : crossFetch;
    try {
        delete init.cache;
        return fetchFunction(input, init);
    } catch (e) {
        // it could never reach here
    }
```

@rishabhpoddar @anku255  Would it make sense to have a setting flag to enable/disable adding `cache` to `init` object behavior, so users have control over which behavior to use?

anku255:
@ngoquoc I am not able to reproduce this. Can you share your `supertokens-node` and `wrangler` version? 

We also have a [cf-example app](https://github.com/supertokens/supertokens-node/tree/master/examples/cloudflare-workers/with-emailpassword). You can try that and see if you are able to reproduce this issue there.

ngoquoc:
@anku255 
I'm using `"supertokens-node": "^18.0.0"` and `"wrangler": "^3.60.3"`.

I also deployed the example app (respecting the versions in package-lock) and could reproduce the same error. Please see the full error log:
[cf-init-cache-error-log.json](https://github.com/user-attachments/files/15880752/cf-init-cache-error-log.json)

anku255:
Thank you @ngoquoc 

It seems that error is producible after deployment but not locally using the Wrangler CLI.  I will be taking a took soon. Will keep you updated on the same thread.

anku255:
@ngoquoc We have released a fix in v18.0.1. You can try that and let us know.

ngoquoc:
@anku255 thank you very much. Tried 18.0.1 and it works perfectly.

PR #824, opened 2024-04-29
feat: supporting esm import

anuragmerndev:
add the exports in package.json to support esm import for js and ts

## Summary of change

This pull request addresses issue [#783](https://github.com/supertokens/supertokens-node/issues/783) by adding [subpath patterns](https://nodejs.org/api/packages.html#subpath-patterns) exports to the package.json file. These exports support ESM (ECMAScript Modules) imports for both JavaScript and TypeScript files, improving compatibility and ease of use for developers. Now, imports like import Session from "supertokens-node/recipe/session"; can be used without specifying the index.js file explicitly.

## Related issues

-   [Link to issue #783](https://github.com/supertokens/supertokens-node/issues/783)
-   [Link to issue #459](https://github.com/supertokens/supertokens-node/issues/459)

## Test Plan

1. Update the package.json file with the provided exports configuration.
2. Ensure the project builds without errors.
3. Verify that existing imports using ESM syntax work as expected.
4. Test importing modules in both JavaScript and TypeScript files.
5. Run unit tests to confirm that the changes do not introduce regressions.
6. Check for any compatibility issues with existing codebases using the library.

https://github.com/supertokens/supertokens-node/assets/144275260/197d89a4-2817-4a33-9334-85adccfba59f

## Documentation changes

No documentation changes are required for this PR.

## Checklist for important updates

-   [ ] Changelog has been updated
-   [ ] `coreDriverInterfaceSupported.json` file has been updated (if needed)
    -   Along with the associated array in `lib/ts/version.ts`
-   [ ] `frontendDriverInterfaceSupported.json` file has been updated (if needed)
-   [ ] Changes to the version if needed
    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [ ] Had run `npm run build-pretty`
-   [ ] Had installed and ran the pre-commit hook
-   [ ] If new thirdparty provider is added,
    -   [ ] update switch statement in `recipe/thirdparty/providers/configUtils.ts` file, `createProvider` function.
    -   [ ] add an icon on the user management dashboard.
-   [ ] Issue this PR against the latest non released version branch.
    -   To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.
    -   If no such branch exists, then create one from the latest released branch.
-   [ ] If have added a new web framework, update the `add-ts-no-check.js` file to include that
[...]

PR #850, opened 2024-06-07
feat!: better support for ESM imports

porcellus:
## Summary of change

- Adds export mapping to make ESM imports more comfortable.
- Adds test (added to CI as well) to ensure all files are importable.

## Related issues

-   Original impl: https://github.com/supertokens/supertokens-node/pull/824
-   

## Test Plan

Added a new test file that is run during CI.

## Documentation changes

N/A

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


rishabhpoddar:
@porcellus there is already a 19.0 branch, which @sattvikc is working on, so this can go in that and be released as part of that release.

PR #855, opened 2024-06-11
Feat: anomaly and security service integration

rishabhpoddar:
## Summary of change

TODO

## Related issues

-   https://github.com/supertokens/anomaly-detection-service/issues/1

## Test Plan

TODO

## Documentation changes

TODO

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/types.ts
```diff
@@ -65,6 +67,202 @@ export type TypeInput = {
...
+        googleRecaptcha?: {
+            // if the user provides both, we will use v2
+            v2SecretKey?: string;
+            v1SecretKey?: string;
+        };
+        override?: (
+            originalImplementation: SecurityFunctions,
+            builder?: OverrideableBuilder<SecurityFunctions>
+        ) => SecurityFunctions;
+    };
```
rishabhpoddar, line 80: we can ask the users to pass the api keys for the services here, and then tell the core that these are being used in there. This way, the core would know to send license key ping info.

lib/ts/recipe/session/types.ts
```diff
@@ -241,11 +249,17 @@ export type RecipeInterface = {
         userContext: UserContext;
     }): Promise<SessionContainerInterface | undefined>;
```
rishabhpoddar, line 251: in getSession, if we are doing checkDataabase: true, it should check for the user id / ip  ban as well. It can throw unauthoized error, just like refresh

rishabhpoddar: Also consider that we can utilise the cache for this?

rishabhpoddar: Also session operations like regenerateSession. You can check for user banning and ip here, cause we have access to the request object. This way, we also do these checks for email verification

PR #861, opened 2024-06-18
(DRAFT) feat: Add "USER_BANNED_ERROR" status to email-password recipe

sioquim:
## Summary of change

This PR adds the "USER_BANNED_ERROR" status to the email-password recipe. This change prepares the recipe for future implementation, where we can ban and unban users effectively.

## Test Plan

To verify the changes, the following steps were taken:

1. Updated the sign-in, password reset, and email verification functions to handle the new "USER_BANNED_ERROR" status.
2. Added tests to ensure the new status is correctly returned when a user is banned.
3. Manually tested the sign-in and password reset flows to confirm the new error status is handled as expected.

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2

### DIFF
--- a/lib/ts/recipe/accountlinking/types.ts
+++ b/lib/ts/recipe/accountlinking/types.ts
@@ -201,5 +201,6 @@ export type AccountInfoWithRecipeId = {
 export type RecipeLevelUser = {
     tenantIds: string[];
     timeJoined: number;
+    timeBanned: number | null;
     recipeUserId: RecipeUserId;
 } & AccountInfoWithRecipeId;
--- a/lib/ts/recipe/emailpassword/api/implementation.ts
+++ b/lib/ts/recipe/emailpassword/api/implementation.ts
@@ -63,6 +63,9 @@ export default function getAPIImplementation(): APIInterface {
                   status: "OK";
               }
             | { status: "PASSWORD_RESET_NOT_ALLOWED"; reason: string }
+            | {
+                  status: "USER_BANNED_ERROR";
+              }
             | GeneralErrorResponse
         > {
             const email = formFields.filter((f) => f.id === "email")[0].value;
@@ -350,6 +353,9 @@ export default function getAPIImplementation(): APIInterface {
                   user: User;
                   email: string;
               }
+            | {
+                  status: "USER_BANNED_ERROR";
+              }
             | { status: "RESET_PASSWORD_INVALID_TOKEN_ERROR" }
             | { status: "PASSWORD_POLICY_VIOLATED_ERROR"; failureReason: string }
             | GeneralErrorResponse
@@ -389,6 +395,9 @@ export default function getAPIImplementation(): APIInterface {
                   }
                 | { status: "RESET_PASSWORD_INVALID_TOKEN_ERROR" }
                 | { status: "PASSWORD_POLICY_VIOLATED_ERROR"; failureReason: string }
+                | {
+                      status: "USER_BANNED_ERROR";
+                  }
                 | GeneralErrorResponse
             > {
                 let updateResponse = await options.recipeImplementation.updateEmailOrPassword({
@@ -577,6 +586,9 @@ export default function getAPIImplementation(): APIInterface {
                   status: "OK";
                   session: SessionContainerInterface;
                   user: User;
+              }
+            | {
+                  status: "USER_BANNED_ERROR";
               }
             | {
                   status: "WRONG_CREDENTIALS_ERROR";
--- a/lib/ts/recipe/emailpassword/index.ts
+++ b/lib/ts/recipe/emailpassword/index.ts
@@ -102,7 +102,11 @@ export default class Wrapper {
         password: string,
         session?: undefined,
         userContext?: Record<string, any>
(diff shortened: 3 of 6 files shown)


lib/ts/recipe/accountlinking/types.ts
```diff
@@ -201,5 +201,6 @@ export type AccountInfoWithRecipeId = {
 export type RecipeLevelUser = {
     tenantIds: string[];
     timeJoined: number;
+    timeBanned: number | null;
```
rishabhpoddar, line 204: i dont think it makes sense to add this on a per recipe level user, instead, just on the top level user object is fine

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -63,6 +63,9 @@ export default function getAPIImplementation(): APIInterface {
                   status: "OK";
               }
             | { status: "PASSWORD_RESET_NOT_ALLOWED"; reason: string }
+            | {
+                  status: "USER_BANNED_ERROR";
+              }
```
rishabhpoddar, line 68: the API should not output this, as we already have PASSWORD_RESET_NOT_ALLOWED as a possible output.

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -350,6 +353,9 @@ export default function getAPIImplementation(): APIInterface {
                   user: User;
                   email: string;
               }
+            | {
+                  status: "USER_BANNED_ERROR";
+              }
```
rishabhpoddar, line 358: no need to modify this API, since the token cant be generated in the first place in case the user is banned

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -578,6 +587,9 @@ export default function getAPIImplementation(): APIInterface {
                   session: SessionContainerInterface;
                   user: User;
               }
+            | {
+                  status: "USER_BANNED_ERROR";
```
rishabhpoddar, line 591: no need to modify the API interface as we already have SIGN_IN_NOT_ALLOWED. Instead, modify the recipe interface.

lib/ts/recipe/emailpassword/api/implementation.ts
rishabhpoddar: Instead, modify the recipe interface.

lib/ts/recipe/emailpassword/index.ts
```diff
@@ -112,6 +116,7 @@ export default class Wrapper {
     ): Promise<
         | { status: "OK"; user: User; recipeUserId: RecipeUserId }
         | { status: "WRONG_CREDENTIALS_ERROR" }
+        | { status: "USER_BANNED_ERROR" }
```
rishabhpoddar, line 119: probably also want to output the user object alongside this. Cause this way, devs can know which user id this belongs too

lib/ts/recipe/emailpassword/index.ts
```diff
@@ -153,7 +159,7 @@ export default class Wrapper {
         email: string,
         password: string,
         userContext?: Record<string, any>
-    ): Promise<{ status: "OK" | "WRONG_CREDENTIALS_ERROR" }> {
+    ): Promise<{ status: "OK" | "WRONG_CREDENTIALS_ERROR" | "USER_BANNED_ERROR" }> {
```
rishabhpoddar, line 162: probably also want to output the user object alongside this. Cause this way, devs can know which user id this belongs too

lib/ts/recipe/emailpassword/index.ts
```diff
@@ -199,7 +207,7 @@ export default class Wrapper {
         userContext?: Record<string, any>
     ): Promise<
         | {
-              status: "OK" | "UNKNOWN_USER_ID_ERROR" | "RESET_PASSWORD_INVALID_TOKEN_ERROR";
+              status: "OK" | "USER_BANNED_ERROR" | "UNKNOWN_USER_ID_ERROR" | "RESET_PASSWORD_INVALID_TOKEN_ERROR";
```
rishabhpoddar, line 210: no need to modify this, since token generation wont be possible in case the user is banned

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -237,6 +237,9 @@ export type APIInterface = {
               | {
                     status: "OK";
                 }
```
rishabhpoddar, line 239: you want to modify the recipe interface, and not api interface.

lib/ts/user.ts
```diff
@@ -74,6 +76,7 @@ export class LoginMethod implements RecipeLevelUser {
             phoneNumber: this.phoneNumber,
             thirdParty: this.thirdParty,
             timeJoined: this.timeJoined,
+            timeBanned: this.timeBanned,
```
rishabhpoddar, line 79: dont think we need to return this boolean to the frontend (this function is used to serialise the user object to return to the frontend).

rishabhpoddar (changes requested):

PR #863, opened 2024-06-19
feat: Add recipe functions to update/delete OAuth2Client

anku255:
## Summary of change

This PR adds the recipe functions to update/delete OAuth2Client.

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/recipe/oauth2/recipeImplementation.ts
+++ b/lib/ts/recipe/oauth2/recipeImplementation.ts
@@ -20,10 +20,6 @@ import { toSnakeCase, transformObjectKeys } from "../../utils";
 import { OAuth2Client } from "./OAuth2Client";
 import { RecipeInterface, TypeNormalisedInput } from "./types";
 
-function createOAuth2ClientFromResponse(response: any): OAuth2Client {
-    return new OAuth2Client(transformObjectKeys(response, "camelCase"));
-}
-
 export default function getRecipeInterface(
     querier: Querier,
     _config: TypeNormalisedInput,
@@ -40,7 +36,7 @@ export default function getRecipeInterface(
             if (response.status === "OK") {
                 return {
                     status: "OK",
-                    client: createOAuth2ClientFromResponse(response.data),
+                    client: OAuth2Client.fromAPIResponse(response.data),
                 };
             } else {
                 return {
@@ -73,7 +69,7 @@ export default function getRecipeInterface(
             if (response.status === "OK") {
                 return {
                     status: "OK",
-                    client: createOAuth2ClientFromResponse(response.data),
+                    client: OAuth2Client.fromAPIResponse(response.data),
                 };
             } else {
                 return {
--- a/lib/ts/recipe/oauth2/recipeImplementation.ts
+++ b/lib/ts/recipe/oauth2/recipeImplementation.ts
@@ -16,9 +16,13 @@
 import NormalisedURLPath from "../../normalisedURLPath";
 import { Querier } from "../../querier";
 import { NormalisedAppinfo } from "../../types";
-import { toCamelCase } from "../../utils";
+import { toSnakeCase, transformObjectKeys } from "../../utils";
 import { OAuth2Client } from "./OAuth2Client";
 import { RecipeInterface, TypeNormalisedInput } from "./types";
+
+function createOAuth2ClientFromResponse(response: any): OAuth2Client {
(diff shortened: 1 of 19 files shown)


lib/ts/recipe/oauth2/recipeImplementation.ts
```diff
@@ -16,10 +16,14 @@
...
 import { Querier } from "../../querier";
 import { NormalisedAppinfo } from "../../types";
-import { toCamelCase } from "../../utils";
+import { toSnakeCase, transformObjectKeys } from "../../utils";
 import { OAuth2Client } from "./OAuth2Client";
 import { RecipeInterface, TypeNormalisedInput } from "./types";
 
+function createOAuth2ClientFromResponse(response: any): OAuth2Client {
+    return new OAuth2Client(transformObjectKeys(response, "camelCase"));
+}
```
porcellus, line 25: I'd prefer if this was defined on `OAuth2Client`

lib/ts/recipe/oauth2/types.ts
```diff
@@ -122,3 +151,9 @@ export type OAuth2ClientOptions = {
 };
 
 export type CreateOAuth2ClientInput = Partial<Omit<OAuth2ClientOptions, "createdAt" | "updatedAt">>;
+
+export type UpdateOAuth2ClientInput = Nullable<CreateOAuth2ClientInput>;
```
porcellus, line 155: Why do we need to make this `Nullable`? I thought it is already nullable wherever it is a valid value.

lib/ts/utils.ts
```diff
@@ -428,3 +428,18 @@ export function toCamelCase(str: string): string {
         return match.toUpperCase().replace("-", "").replace("_", "");
     });
 }
+
+export function toSnakeCase(str: string): string {
+    return str.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`);
+}
+
+// Transforms the keys of an object from camelCase to snakeCase or vice versa.
+export function transformObjectKeys<T>(obj: { [key: string]: any }, caseType: "snake-case" | "camelCase"): T {
```
porcellus, line 437: Are there cases where the objects passed to this have embedded objects (not user-defined)? If so you may want to make this recursive. Not sure if we want to keep these helpers in the long run though.

test/oauth2/oauth2client.test.js
```diff
@@ -76,7 +76,107 @@ describe(`OAuth2ClientTests: ${printPath("[test/oauth2/oauth2client.test.js]")}`
...
+            },
+            recipeList: [OAuth2Recipe.init()],
+        });
+
+        // Only run for version >= 2.9
+        let querier = Querier.getNewInstanceOrThrowError(undefined);
+        let apiVersion = await querier.getAPIVersion();
+        if (maxVersion(apiVersion, "2.8") === "2.8") {
+            return;
+        }
```
porcellus, line 102: I don't think 2.8 is the right minimum.

porcellus (changes requested):

lib/ts/utils.ts
anku255: We don't need to make it recursive because only `jwks` and `metadata` fields accept objects and we don't want to alter them.

test/oauth2/oauth2client.test.js
anku255: Removed it, as discussed.

lib/ts/recipe/oauth2/types.ts
anku255: I have updated the types as the per the Hydra API.

For `OAuth2ClientOptions`, there are 3 categories of properties -

1. Always has a non-nullable value.
2. Always exists (either valid/null value)
3. Only exists if a valid value is given


`CreateOAuth2ClientInput` type is basically `OAuth2ClientOptions` with every field being marked as optional.


For `UpdateOAuth2ClientInput`, we wanted to allowed `null` only for the properties that can be set to `null` via a PATCH request. There were only 9 such properties.

lib/ts/recipe/oauth2/recipeImplementation.ts
anku255: Done

PR #877, opened 2024-07-08
feat: Add OAuth2Client recipe

anku255:
## Summary of change

This PR adds the OAuth2Client recipe. It is a trimmed down version of our ThirdParty recipe except that it only allows Supertokens client Ids. There are some TODOs in this PR which are either blocked or require some discussion.

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/oauth2client/api/authorisationUrl.ts
```diff
@@ -0,0 +1,48 @@
...
+    apiImplementation: APIInterface,
+    _tenantId: string,
+    options: APIOptions,
+    userContext: UserContext
+): Promise<boolean> {
+    if (apiImplementation.authorisationUrlGET === undefined) {
+        return false;
+    }
+
+    const redirectURIOnProviderDashboard = options.req.getKeyValueFromQuery("redirectURIOnProviderDashboard");
```
porcellus, line 31: Not sure if the name makes sense, but it should be OK. Please add a TODO note to check this name later.

lib/ts/recipe/oauth2client/api/implementation.ts
```diff
@@ -0,0 +1,65 @@
...
+                status: "OK",
+                ...authUrl,
+            };
+        },
+        signInPOST: async function (input) {
+            const { options, tenantId, userContext } = input;
+
+            const providerConfig = await options.recipeImplementation.getProviderConfig({ userContext });
+
+            let oAuthTokensToUse: any = {};
```
porcellus, line 20: Can we do better than `any`?

lib/ts/recipe/oauth2client/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+import Recipe from "./recipe";
+import { RecipeInterface, APIInterface, APIOptions } from "./types";
+
+export default class Wrapper {
```
porcellus, line 19: We should expose `getAuthorisationRedirectURL`, `exchangeAuthCodeForOAuthTokens` and `getUserInfo` if practicable.
`getProviderConfig` can remain internal and the other functions can call it in the index file if necessary.

lib/ts/recipe/oauth2client/utils.ts
```diff
@@ -0,0 +1,175 @@
...
+        ...config?.override,
+    };
+
+    return {
+        providerConfig: config.providerConfig,
+        override,
+    };
+}
+
+export async function doGetRequest(
```
porcellus, line 60: couldn't we move these into a common util thing?

recipe/openid/index.js
```diff
@@ -0,0 +1,6 @@
+"use strict";
```
porcellus, line 1: This is not necessarily an issue, but why do we need to expose the openid recipe now?

porcellus (changes requested):

lib/ts/recipe/oauth2client/api/authorisationUrl.ts
anku255: done

lib/ts/recipe/oauth2client/api/implementation.ts
anku255: Sure! Added better types.

lib/ts/recipe/oauth2client/index.ts
anku255: done

lib/ts/recipe/oauth2client/utils.ts
anku255: I have moved them to `recipe/utils.ts`.  Let me know if there is a more suitable place.

recipe/openid/index.js
anku255: Initially I assume it was needed but it is not. I have removed it.

lib/ts/recipe/utils.ts
```diff
@@ -0,0 +1,122 @@
+import * as jose from "jose";
```
porcellus, line 1: Please move this file one level up and rename it to something like `thirdpartyutils`

lib/ts/recipe/oauth2client/index.ts
```diff
@@ -14,12 +14,37 @@
...
+
+    static async exchangeAuthCodeForOAuthTokens(input: {
+        providerConfig: ProviderConfigWithOIDCInfo;
+        redirectURIInfo: {
+            redirectURIOnProviderDashboard: string;
+            redirectURIQueryParams: any;
+            pkceCodeVerifier?: string | undefined;
+        };
+    }) {
+        return await Recipe.getInstanceOrThrowError().recipeInterfaceImpl.exchangeAuthCodeForOAuthTokens(input);
```
porcellus, line 34: Ideally, this would call `getProviderConfig` itself (esp. since it isn't exposed) instead of taking `providerConfig` as a param

lib/ts/recipe/oauth2client/index.ts
```diff
@@ -14,12 +14,37 @@
...
+    }
+
+    static async getUserInfo(input: { providerConfig: ProviderConfigWithOIDCInfo; oAuthTokens: OAuthTokens }) {
```
porcellus, line 37: Please add a `userContext` parameter to all functions both here and in the `RecipeInterface`

lib/ts/recipe/oauth2client/index.ts
```diff
@@ -14,12 +14,37 @@
...
+    }
+
+    static async exchangeAuthCodeForOAuthTokens(input: {
+        providerConfig: ProviderConfigWithOIDCInfo;
+        redirectURIInfo: {
+            redirectURIOnProviderDashboard: string;
+            redirectURIQueryParams: any;
+            pkceCodeVerifier?: string | undefined;
+        };
+    }) {
```
porcellus, line 33: Please make the function signature match other index files. (for all other exposed functions in this file as well)
Check for example the `lib/ts/recipe/session/index.ts`

```suggestion
    static async exchangeAuthCodeForOAuthTokens(
        redirectURIInfo: {
            redirectURIOnProviderDashboard: string;
            redirectURIQueryParams: any;
            pkceCodeVerifier?: string | undefined;
        },
        userContext?: Record<string, any>
    ) {
```

porcellus (changes requested):

lib/ts/recipe/utils.ts
anku255: done

lib/ts/recipe/oauth2client/index.ts
anku255: done

lib/ts/recipe/oauth2client/index.ts
anku255: done

lib/ts/recipe/oauth2client/index.ts
porcellus: The userContext passed to this is the wrong type. Please check the linked file on how we handle it in other recipes.

lib/ts/recipe/oauth2client/types.ts
```diff
@@ -78,24 +78,25 @@ export type TypeNormalisedInput = {
...
             fromIdTokenPayload?: { [key: string]: any };
             fromUserInfoAPI?: { [key: string]: any };
-        };
-        tenantId: string;
-        userContext: UserContext;
-    }): Promise<{
+        },
+        tenantId: string,
+        userContext: UserContext
+    ): Promise<{
```
porcellus, line 99: These should still be getting the input object.

lib/ts/recipe/oauth2client/recipeImplementation.ts
```diff
@@ -24,8 +24,12 @@ export default function getRecipeImplementation(_querier: Querier, config: TypeN
...
 
     return {
-        getAuthorisationRedirectURL: async function ({ redirectURIOnProviderDashboard }) {
-            const providerConfig = await this.getProviderConfig();
+        getAuthorisationRedirectURL: async function (
+            this: RecipeInterface,
+            redirectURIOnProviderDashboard,
+            userContext
+        ) {
+            const providerConfig = await this.getProviderConfig(userContext);
```
porcellus, line 32: This isn't really what I meant: the `getProviderConfig` should be called in `index.ts` and passed to this function. We just don't want to require users to call this.

porcellus (changes requested):

lib/ts/recipe/oauth2client/types.ts
anku255: I have fixed it.

lib/ts/recipe/oauth2client/recipeImplementation.ts
anku255: fixed.

porcellus (changes requested):
There is still on open comment (https://github.com/supertokens/supertokens-node/pull/877#discussion_r1677100725)

porcellus (approved):

PR #898, opened 2024-07-26
Support cloudflare (edge) node runtimes

labithiotis:
## Summary of change

Update library to support [Cloudflare Workers](https://developers.cloudflare.com/workers/runtime-apis/nodejs/) (WorkerD/Edge) runtime plus potentially other edge runtimes.

To make the library work a few changes were needed:
1. No global `Buffer`, changed to import buffer from node eg. `import { Buffer } from "node:buffer";`. 
2. No global `process`, changed to use import instead from node eg. `import { env } from "node:process”;`.
3. `querystring` is not bundled with edge runtime so I added `querystringify` which is very popular library to handle the same behaviour.
4. `psl` is using punycode which isn’t included in edge runtime and is also deprecated in node 22, lastly PSL seems to be [unmaintained](https://github.com/lupomontero/psl/issues/318) so switched to favoured alternated `tldts`. 
5. `zlib` is not bundled with edge runtime and is used by `inflation`, opted to use `pako` and replicated the inflation inflate method where it was used to avoid using `zlib`.
6. Hoisted nested module requires to use import instead, initially they were done to resolve circular reference but with esm that's not an issue.
7. CommonJS/Dynamic requires is not supported in edge runtime so it needs to support ESM. Adding type module to package.json and I updated all type import to have `type` prefix and added tsup to compile CJS and ESM files.
8. Removed redirect “helpers” at root in favour of exports definitions (removed: nestjs, framework, recipe, types and index.js at root)

> 👉 **I have not included the build files in this PR just yet to make it easier to review PR.**

**🙋‍♂️ Help needed!! I can’t seem to make tsup work with all the exports paths correctly, I get `Cannot find module` in my test projects but files exist at the expected path so unsure why. I'm pretty sure I tried every combination and change I could think to resolve this and can't so would really appreciate if someone could help resolve this. (I also tried reverting exports to master and restoring redirect helpers but still had same issue)**

## Test Plan

If I compile code with tsc and ignore cjs it works with basic auth flow in a sample project. 

## Documentation changes

Don’t think docs need changing?

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Add build files


rishabhpoddar:
Hi @labithiotis , thanks for the PR and detailed description. We are currently very busy with our roadmap, and will come around to seeing this and helping soon enough.

labithiotis:
Hi @rishabhpoddar,

Unfortunately, I was unable to get the library to work within the CF/Remix ecosystem using both CJS and ES modules. After extensive attempts, I decided to opt for an alternate auth provider that offered out-of-the-box compatibility and smoother integration with Remix, including prebuilt components.

I still believe Supertokens is superior to Auth0 in many respects, but my specific stack choices prevented successful integration. I may still consider Supertokens in future when edge environments are supported.

Would you like me to leave this PR open for further review, or should I close it?

rishabhpoddar:
Hi @labithiotis appreciate the feedback and the attempt to getting it to work. We do plan on working on fixing this integration, so please keep this PR open. Someone from our team will get around to doing it soon enough.

Thanks.

rishabhpoddar:
We have made our SDK edge compatible from version 20.1.0 onwards.

Issue #745, opened 2023-11-17
Build error in next.js when using `runtime = 'edge'`


porcellus:
This causes an error: `Module not found: Can't resolve 'zlib'` (same for querystring and crypto) during building and running in dev mode.

Reproduction:
Using our nextjs example, add
```
export const runtime = 'edge';
```
to the user route https://github.com/supertokens/next.js/blob/canary/examples/with-supertokens/app/api/user/route.ts reproduces the issue.

Reported in: https://discordapp.com/channels/603466164219281420/1173443361550643250

rishabhpoddar:
This issue has been solved in node sdk >= 20

PR #918, opened 2024-08-30
feat: Add edge compatibility for custom frameworks and Next.JS

deepjyoti30Alt:
## Summary of change

This PR adds fixes for various things that break when running this SDK with next.js on edge runtime.

## Related issues

-   #745

### Testing with NextJS app router in edge runtime:

1. Setup the example app from the repo
2. Make all the server code use `edge` runtime by adding `export const runtime = "edge";` to the top of the file.
3. Run `npm run build` or `npm run dev` (live server)
4. Both working without errors would indicate that the fix is working

### Testing with NextJS pages router in edge runtime

1. Setup the example next app using pages
2. Make all the server code use `edge` runtime by adding `export const runtime = "edge";` to the top of the file.
3. Run `npm run build` or `npm run dev` (live server)
4. Both working without errors would indicate that the fix is working

### Test with express to confirm that `inflate` works with different compression methods:

1. Create a demo BE app with express
2. Inflate should work with `br`, `gzip` and `deflate`.
3. Create a JSON file with the body, call it `body.json`.
4. Create compressed versions using one of the following:

```sh
# For gzip
gzip -c body.json > body.gz

# For brotli
brotli -o body.br body.json
```

5. Run cURL request with the created body:

```sh
# For gzip
curl -v -i -X POST http://localhost:3001/auth/signin -H 'Content-Encoding: gzip' --data-binary @body.gz

# For brotli
curl -v -i -X POST http://localhost:3001/auth/signin -H 'Content-Encoding: br' --data-binary @body.br
```

6. Request should succeed successfully in order for the test to be successful.

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/framework/utils.ts
```diff
@@ -190,6 +253,7 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(metho
 }
 
 export async function assertFormDataBodyParserHasBeenUsedForExpressLikeRequest(request: Request) {
+    console.log("Entering assert 2");
```
porcellus, line 266: Please remove console.log

lib/ts/framework/utils.ts
```diff
@@ -179,7 +241,8 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(metho
             try {
                 // parsing it again to make sure that the request is parsed atleast once by a json parser
                 request.body = await parseJSONBodyFromRequest(request);
-            } catch {
+            } catch (err) {
+                console.log("err:", err);
```
porcellus, line 255: Please remove console.log

lib/ts/framework/utils.ts
```diff
@@ -23,7 +22,72 @@ import type { HTTPMethod } from "../types";
...
+export function parseParams(string: string): object {
+    // Set up a new URLSearchParams object using the string.
+    const params = new URLSearchParams(string);
+
+    // Get an iterator for the URLSearchParams object.
+    const entries = params.entries();
+
+    const result: { [key: string]: any } = {};
+
+    // Loop through the URLSearchParams object and add each key/value
```
porcellus, line 88: Was this always done in the library we used to use?

lib/ts/framework/utils.ts
```diff
@@ -23,7 +22,72 @@ import type { HTTPMethod } from "../types";
...
+    for (const [key, value] of entries) {
+        // Split comma-separated values into an array.
+        result[key] = value.split(",");
+
+        // If a key does not have a value, delete it.
+        if (!value) {
```
porcellus, line 94: Would this ever be true? I suspect the line above would throw in that case.

lib/ts/framework/utils.ts
```diff
@@ -23,7 +22,82 @@ import type { HTTPMethod } from "../types";
...
+            inflator.push(chunk, false);
+        }
+
+        if (inflator.err) {
+            throw new Error(`Decompression error: ${inflator.msg}`);
+        }
+
+        decompressedData = inflator.result;
+    } else if (encoding === "br") {
+        if (!brotliDecompress) throw new Error("Brotli decompression not supported on the platform");
```
porcellus, line 58: How does this show up? Does it crash the server or produces a 500 response?

porcellus (changes requested):

lib/ts/framework/utils.ts
deepjyoti30Alt: Done

lib/ts/framework/utils.ts
deepjyoti30Alt: My bad, I missed these. Removed.

lib/ts/framework/utils.ts
deepjyoti30Alt: It is for the case of an empty string. The value would never contain a `null` or `undefined` as it goes through URLSearchParams. I also have a test case for this.

Take the following example:

```js
const p = parseParam("?a=1&b=")

// In this case b's value would evaluate to an empty string and parseParams will remove it
console.log(p).

// p should be { "a": ["1"] }
```

lib/ts/framework/utils.ts
deepjyoti30Alt: It doesn't crash the server but a BAD_INPUT_ERROR is returned. This is handled by the parent function and thrown.

I saw however, that the parent function did not care about the type of error so I did not update it.

Here's the line I am referring to: https://github.com/supertokens/supertokens-node/blob/f51a72cb23490a95b3e69811e16a4cd2684589ac/lib/ts/framework/utils.ts#L182

lib/ts/framework/utils.ts
deepjyoti30Alt: Just checked the source code, they do a `forEach` loop but yeah, this is done it seems.

lib/ts/framework/utils.ts
porcellus: I see, but then `Please make sure to pass a valid JSON input in the request body` is not the correct error message.

lib/ts/framework/utils.ts
porcellus: Are you sure about that? Could you provide a link?

lib/ts/framework/utils.ts
porcellus: I see, my bad. However, that is a very unexpected behaviour.

porcellus (changes requested):

lib/ts/framework/utils.ts
deepjyoti30Alt: I agree but since that was going to require more changes that were not directly related to this change, I thought it was out of scope.

Let me know if I should go ahead and make those changes.

deepjyoti30Alt: Updated the code as per our discussion.

lib/ts/framework/utils.ts
deepjyoti30Alt: Updated it as per our discussion

.github/workflows/test-cf-worker-next.yml
```diff
@@ -0,0 +1,73 @@
+name: "Test edge function compatibility for Next.js on Cloudflare Workers"
+on: push
+jobs:
+    test:
+        runs-on: ubuntu-latest
+        env:
+            CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+            CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+            APP_URL: supertokens-node-b95.pages.dev
```
porcellus, line 9: I think this should also come from the env, otherwise we'll have to update this again after merging to use our own credentials.

.github/workflows/test-cf-worker-next.yml
```diff
@@ -0,0 +1,73 @@
...
+                  find app/api -type f \( -name "*.js" -o -name "*.ts" \) -exec sed -i '1s/^/export const runtime = "edge";\n/' {} +
+                  echo 'export const runtime = "edge";' >> app/auth/[[...path]]/page.tsx
+
+            # Install next on pages to build the app
+            - name: Install next-on-pages
+              run: npm install --save-dev @cloudflare/next-on-pages
+
+            # Setup the compatibility flag to make non edge functions run
+            - name: Create a wrangler.toml
+              run: echo "compatibility_flags = [ "nodejs_compat" ]" >> wrangler.toml
```
porcellus, line 35: I think I'm missing something, why do we need this?

.github/workflows/test-cf-worker-next.yml
```diff
@@ -0,0 +1,73 @@
...
+            - run: echo $GITHUB_SHA
+            - run: npm install git+https://github.com:supertokens/supertokens-node.git#$GITHUB_SHA
+            - run: npm install
+            - run: npm install mocha@6.1.4 jsdom-global@3.0.2 puppeteer@^11.0.0 isomorphic-fetch@^3.0.0
+
+            # Step to update the runtime to edge to all files in app/api/ and remove pages path
+            # to avoid conflicts
+            - name: Add runtime export to API files
+              run: |
+                  rm -rf pages
```
porcellus, line 25: You can just modify the example app itself to use app dir only. I think you'd only use both during migration which is weird in an example.

.github/workflows/test-cf-worker-next.yml
```diff
@@ -0,0 +1,73 @@
...
+            - name: Build using next-on-pages
+              run: npx next-on-pages
+
+            - name: Publish to Cloudflare Pages
+              id: deploy
+              uses: cloudflare/pages-action@v1
+              with:
+                  apiToken: ${{ env.CLOUDFLARE_API_TOKEN }}
+                  accountId: ${{ env.CLOUDFLARE_ACCOUNT_ID }}
+                  projectName: "supertokens-node-pr-check-for-edge-function-compat"
```
porcellus, line 49: I think this should also come from the env.

lib/ts/framework/awsLambda/framework.ts
```diff
@@ -45,7 +44,7 @@ export class AWSRequest extends BaseRequest {
         if (this.event.body === null || this.event.body === undefined) {
             return {};
         } else {
-            const parsedUrlEncodedFormData = parse(this.event.body);
+            const parsedUrlEncodedFormData = Object.fromEntries(new URLSearchParams(this.event.body).entries());
```
porcellus, line 47: If someone passes a malformed body how does that show up? Is the error message correct?

lib/ts/framework/loopback/framework.ts
```diff
@@ -92,7 +92,7 @@ export class LoopbackResponse extends BaseResponse {
     sendHTMLResponse = (html: string) => {
         if (!this.response.writableEnded) {
             this.response.set("Content-Type", "text/html");
-            this.response.status(this.statusCode).send(Buffer.from(html));
+            this.response.status(this.statusCode).send(getBuffer().from(html));
```
porcellus, line 95: Please check that we actually need to make this into a buffer.
If we do, I'd prefer if we used a `getBufferFromString` function instead of `getBuffer().from`

lib/ts/framework/utils.ts
```diff
@@ -173,13 +222,22 @@ export async function assertThatBodyParserHasBeenUsedForExpressLikeRequest(metho
             }
         } else if (
             request.body === undefined ||
-            Buffer.isBuffer(request.body) ||
+            getBuffer().isBuffer(request.body) ||
```
porcellus, line 225: I'd prefer if the utils exported an `isBuffer` function directly instead

test/framework/utils.test.js
```diff
@@ -0,0 +1,14 @@
+let assert = require("assert");
```
porcellus, line 1: This should be removed, right?

porcellus (changes requested):

.github/workflows/test-cf-worker-next.yml
deepjyoti30Alt: This is to run the non edge code of the app. `edge` runtime works on non-static routes but static routes still use it(?). I initially tested without the compatibility flags but cloudflare was throwing an error (build would work fine but CF functions would return error) when trying to access the page.

After I added this, it started working fine. However, I have verified that all the non-static code runs on edge runtime. This can be verified by looking at the build logs from the workflow.

.github/workflows/test-cf-worker-next.yml
deepjyoti30Alt: I did not know the context for keeping both so I left it as is (as the netlify example also uses it). I will make this change.

lib/ts/framework/loopback/framework.ts
deepjyoti30Alt: > Please check that we actually need to make this into a buffer.

I don't think it's necessary as the expected type of `html` is string and express takes care of it automatically.

lib/ts/framework/awsLambda/framework.ts
deepjyoti30Alt: Fixed it, it will throw a input error if that happens.

examples/cloudflare-workers/with-be-emailpassword/index.ts
```diff
@@ -0,0 +1,41 @@
...
+app.use("*", async (c, next) => {
+    return await cors({
+        origin: getWebsiteDomain(),
+        credentials: true,
+        allowHeaders: ["Content-Type", ...supertokens.getAllCORSHeaders()],
+        allowMethods: ["GET", "POST", "PUT", "PATCH", "DELETE"],
+    })(c, next);
+});
+
+// This exposes all the APIs from SuperTokens to the client.
```
porcellus, line 25: And adds the session to the request object if one exists.

lib/ts/framework/constants.ts
```diff
@@ -13,3 +13,6 @@
  * under the License.
  */
 export const COOKIE_HEADER = "Set-Cookie";
+
+// Define error message for brotli decompression not being supported
+export const BROTLI_DECOMPRESSION_ERROR_MESSAGE = "Brotli decompression not implement, use a middleware instead";
```
porcellus, line 18: Please add a middleware that handles decompression before the SuperTokens middleware.

lib/ts/framework/custom/nodeHeaders.ts
```diff
@@ -64,7 +64,7 @@ export default class Headers extends URLSearchParams {
         } else if (init == null) {
             // eslint-disable-line no-eq-null, eqeqeq
             // No op
-        } else if (typeof init === "object" && !types.isBoxedPrimitive(init)) {
+        } else if (typeof init === "object" && isBoxedPrimitive(init)) {
```
porcellus, line 67: why did this change?

examples/cloudflare-workers/with-be-emailpassword/package.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "with-be-email-password-hono",
```
porcellus, line 2: Please rename this as `with-email-password-hono-be-only` (incl the directory)

porcellus (changes requested):

lib/ts/framework/custom/nodeHeaders.ts
deepjyoti30Alt: `util` from where `types` was being imported is not supported on `edge`.

porcellus: right, but the condition is flipped.

deepjyoti30Alt: Ohh, my bad. Let me fix that 😅

examples/cloudflare-workers/with-be-emailpassword/package.json
deepjyoti30Alt: Done

lib/ts/framework/custom/nodeHeaders.ts
deepjyoti30Alt: Done

porcellus (approved):

Issue #447, opened 2022-12-06
Do not call form field validator for sign in and reset password code POST API


rishabhpoddar:
This makes customisations harder to do. Also, it doesn't add any purpose since these APIs are not creating a new user. Just doing operations on existing users.

PR #920, opened 2024-08-31
feat: skip field validations on non signup routes

akashrajum7:
## Summary of change

From my understanding of the ticket, we wanted to avoid doing field validations on form fields for non-signup routes, since it is field validation that emits FORM_FIELD errors. This would mean we would be delegating the validation to the core, for say email and password. Please feel free to let me know if this is not the intended change of the ticket and if it is to be done differently. I am quite new to the repo so my understanding is quite limited. I am happy to do it the right way with little guidance. Thank you for your time and patience.

## Related issues

-   https://github.com/supertokens/supertokens-node/issues/447
-   https://github.com/supertokens/supertokens-auth-react/issues/638

## Test Plan

I have yet to run the test cases and update them if necessary, will do it once I get a go on the implementation

## Documentation changes

This might need a Docs update as mentioned in the ticket.

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [ ] Issue this PR against the latest non-released version branch.

-   [ ] Get clarification on the approach
-   [ ] Maybe update types where necessary (Need some guidance on this part as well)

### DIFF
--- a/lib/ts/recipe/emailpassword/api/generatePasswordResetToken.ts
+++ b/lib/ts/recipe/emailpassword/api/generatePasswordResetToken.ts
@@ -40,7 +40,8 @@ export default async function generatePasswordResetToken(
         options.config.resetPasswordUsingTokenFeature.formFieldsForGenerateTokenForm,
         requestBody.formFields,
         tenantId,
-        userContext
+        userContext,
+        false
     );
 
     let result = await apiImplementation.generatePasswordResetTokenPOST({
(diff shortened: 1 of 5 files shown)


rishabhpoddar:
Hey @akashrajum7 thanks for this PR. Whilst it makes sense, i won't be merging it right now, since this requires changes to our frontend sdk types and docs for which, we don't yet have the bandwidth.

akashrajum7:
> Hey @akashrajum7 thanks for this PR. Whilst it makes sense, i won't be merging it right now, since this requires changes to our frontend sdk types and docs for which, we don't yet have the bandwidth.

Thanks for the review, sure, let me see if I can find some time this week to both update the test cases for these changes as well as make those changes in the frontend sdk

PR #919, opened 2024-08-31
[WIP] feat: improve form validation

akashrajum7:
## Summary of change

This PR aims to improve the form field validation in emailpassword to allow for optional form fields to be truly optional and not have to send empty values. It looks like the rest of the recipes use the UserMetadata recipe to store extra information.

## Related issues

- https://github.com/supertokens/supertokens-node/issues/300
- https://github.com/supertokens/supertokens-auth-react/issues/638

## Test Plan

Added a test case to check the functionality of optional fields missing in the payload does not result in an error

## Documentation changes

Not sure which section of the docs talks about the behaviour of optional fields currently, can update there once this implementation is completed

## Checklist for important updates

-   [x] Changelog has been updated
-   [x] Changes to the version if needed
    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [x] Had run `npm run build-pretty`
-   [x] Had installed and ran the pre-commit hook
-   [ ] Issue this PR against the latest non-released version branch.
    -   To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.
    -   If no such branch exists, then create one from the latest released branch.

## Remaining TODOs for this PR

-   [ ] Get clarity on what it means to "improve error message" for missing form fields

PR #921, opened 2024-09-02
Fix failing tests regarding form validation

deepjyoti30Alt:
## Summary of change

This PR fixes the [failing tests](https://app.circleci.com/pipelines/github/supertokens/supertokens-node/5985/workflows/cfe50d90-2d11-4f6d-bf95-180345c1235c/jobs/3363) that were a side effect of #919 

## Test Plan

Run the `signupFeature` related tests and verify that all of them work.

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/test/emailpassword/signupFeature.test.js
+++ b/test/emailpassword/signupFeature.test.js
@@ -1093,10 +1093,6 @@ describe(`signupFeature: ${printPath("[test/emailpassword/signupFeature.test.js]
                             id: "email",
                             value: "random@gmail.com",
                         },
-                        {
-                            id: "testField",
-                            value: "",
-                        },
                     ],
                 })
                 .expect(400)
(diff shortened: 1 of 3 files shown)


akashrajum7:
Thanks for this, I did not know that objects as values were allowed.

test/emailpassword/signupFeature.test.js
```diff
@@ -1093,10 +1093,6 @@ describe(`signupFeature: ${printPath("[test/emailpassword/signupFeature.test.js]
                             id: "email",
                             value: "random@gmail.com",
                         },
-                        {
-                            id: "testField",
-                            value: "",
-                        },
```
rishabhpoddar, line 1099: why is this removed?

rishabhpoddar (changes requested):

test/emailpassword/signupFeature.test.js
deepjyoti30Alt: To fix the test. This test is to make sure that we get an error when the required fields are not passed.

The fix was either to remove this (which would throw an error) or add 2 new ones (which would also throw an error)

PR #928, opened 2024-09-25
fix: Add changes to support unknown type in formField values

deepjyoti30-st:
## Summary of change

Makes the internal type of formField values to be `unknown` instead of `string`. Also adds test for signup/password reset API's when they get non string type of values for email/password.

> PS: Updated the bad input error for email/password being invalid to keep it same as Go/Python SDK's

> NOTE: Did not update Changelog as that was updated in the previous PR to indicate that `any` support is added for formField values.

## Test Plan

- `signup`, `signin` and `passwordReset` tests for emailpassword should pass.

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

## Remaining TO-DO's

- [x] Run `build-pretty` once the base branch issues are fixed

### DIFF
--- a/lib/build/recipe/emailpassword/api/implementation.js
+++ b/lib/build/recipe/emailpassword/api/implementation.js
@@ -40,7 +40,16 @@ function getAPIImplementation() {
             };
         },
         generatePasswordResetTokenPOST: async function ({ formFields, tenantId, options, userContext }) {
-            const email = formFields.filter((f) => f.id === "email")[0].value;
+            // NOTE: Check for email being a non-string value. This check will likely
+            // never evaluate to `true` as there is an upper-level check for the type
+            // in validation but kept here to be safe.
+            const emailAsUnknown = formFields.filter((f) => f.id === "email")[0].value;
+            if (typeof emailAsUnknown !== "string")
+                return {
+                    status: "GENERAL_ERROR",
+                    message: "email value needs to be a string",
+                };
+            const email = emailAsUnknown;
             // this function will be reused in different parts of the flow below..
             async function generateAndSendPasswordResetToken(primaryUserId, recipeUserId) {
                 // the user ID here can be primary or recipe level.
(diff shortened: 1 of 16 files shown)


lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -65,7 +65,16 @@ export default function getAPIImplementation(): APIInterface {
...
-            const email = formFields.filter((f) => f.id === "email")[0].value;
+            // NOTE: Check for email being a non-string value. This check will likely
+            // never evaluate to `true` as there is an upper-level check for the type
+            // in validation but kept here to be safe.
+            const emailAsUnknown = formFields.filter((f) => f.id === "email")[0].value;
+            if (typeof emailAsUnknown !== "string")
+                return {
+                    status: "GENERAL_ERROR",
+                    message: "email value needs to be a string",
+                };
```
rishabhpoddar, line 76:
```suggestion
                throw new Exception("Should never come here since we already check that the email value is a string in validateFormFieldsOrThrowError");
```

rishabhpoddar: similar changes everywhere else

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/api/implementation.ts
deepjyoti30-st: Okay.

lib/build/recipe/emailpassword/api/implementation.js
```diff
@@ -40,7 +40,15 @@ function getAPIImplementation() {
             };
         },
```
rishabhpoddar, line 41: modify changelog under the breaking change section about this.

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/api/implementation.ts
```diff
@@ -64,7 +65,15 @@ export default function getAPIImplementation(): APIInterface {
             | { status: "PASSWORD_RESET_NOT_ALLOWED"; reason: string }
             | GeneralErrorResponse
         > {
-            const email = formFields.filter((f) => f.id === "email")[0].value;
+            // NOTE: Check for email being a non-string value. This check will likely
+            // never evaluate to `true` as there is an upper-level check for the type
+            // in validation but kept here to be safe.
+            const emailAsUnknown = formFields.filter((f) => f.id === "email")[0].value;
+            if (typeof emailAsUnknown !== "string")
```
rishabhpoddar, line 72: use brackets always for if statements.

rishabhpoddar (changes requested):

PR #942, opened 2024-10-10
feat: Add passkeys support

coolbueb:
## Summary of change

Add passkeys support 

## Related issues

-   [Core Repo Passkeys](https://github.com/supertokens/supertokens-core/issues/1046)

## Test Plan

- Mock core API calls
- Add tests for each method individually

More details here: https://docs.google.com/document/d/1G7tO9_dSNi8wur3ajGg4pq-wiHatKDbHv2sBt-uSbQg/edit

PR #951, opened 2024-10-16
feat: WebAuthN Sign In, Sign Up and Options methods support

victorbojica:
ONLY REVIEW REQUIRED. NOTHING TESTED YET.

## Summary of change
Implement WebAuthN support according to:
https://docs.google.com/document/d/1G7tO9_dSNi8wur3ajGg4pq-wiHatKDbHv2sBt-uSbQg/edit#heading=h.olee876uqu8a

## Related
Check https://github.com/supertokens/supertokens-node/pull/942 for more feedback

## Test Plan
No testing at the moment.

## Documentation changes

Will have to add WebAuthN recipe documentation

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Check support for other authenticators
-   [ ] Account recovery flow
-   [ ] Options testing
-   [ ] Sign In testing
-   [ ] Sign Up testing
-   [ ] Multiple authenticators testing


lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -154,6 +154,7 @@ export type GetPhoneNumbersForFactorsFromOtherRecipesFunc = (
 
 export const FactorIds = {
     EMAILPASSWORD: "emailpassword",
+    WEBAUTHN: "webauthn",
```
rishabhpoddar, line 157: need to also modify the user object and login methods object (in ts/types.ts)

rishabhpoddar: Relate to that, will all webauthn methods have an email associated with that login method?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+
+export type TypeNormalisedInputRelyingPartyId = (input: {
+    request: BaseRequest | undefined;
+    userContext: UserContext;
+}) => string; // should return the domain of the origin
+
+export type TypeNormalisedInputRelyingPartyName = (input: {
+    request: BaseRequest | undefined;
+    userContext: UserContext;
+}) => string; // should return the app name
```
rishabhpoddar, line 51: why do these need to be functinos that take in request? Maybe instead they should take in the tenant id? What kind of use cases do you think exist that users might want to dynamically set relyingPartyId and relyingPartyName?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+
+export type TypeInputGetOrigin = (input: { request: BaseRequest; userContext: UserContext }) => string;
+
+export type RecipeInterface = {
+    registerOptions(input: {
+        email: string;
+        relyingPartyId: string;
+        relyingPartyName: string;
+        origin: string;
+        timeout: number;
```
rishabhpoddar, line 85: As we had discussed on call, timeout here should be optional in the input, since this function can set a default of it to 5sec

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+        attestation: "none" | "indirect" | "direct" | "enterprise";
```
rishabhpoddar, line 86: Can attestation also be optional?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+        status: "OK";
+        webauthnGeneratedOptionsId: string;
+        rp: {
+            id: string;
+            name: string;
+        };
+        user: {
+            id: string;
+            name: string;
+            displayName: string;
```
rishabhpoddar, line 99: where does name and displayName come from? What effect does this have?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+            requireResidentKey: boolean;
+            residentKey: "required" | "preferred" | "discouraged";
+            userVerification: "required" | "preferred" | "discouraged";
+        };
+    }>;
+
+    signInOptions(input: {
+        relyingPartyId: string;
+        origin: string;
+        timeout: number;
```
rishabhpoddar, line 123: timeout is optional

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<{
+        status: "OK";
+        webauthnGeneratedOptionsId: string;
+        challenge: string;
+        timeout: number;
+        userVerification: "required" | "preferred" | "discouraged";
```
rishabhpoddar, line 131: how is the value of userVerification defined?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+        attestation: "none" | "indirect" | "direct" | "enterprise";
+        pubKeyCredParams: {
+            alg: number;
+            type: string;
+        }[];
+        authenticatorSelection: {
+            requireResidentKey: boolean;
+            residentKey: "required" | "preferred" | "discouraged";
+            userVerification: "required" | "preferred" | "discouraged";
+        };
```
rishabhpoddar, line 117: how are these defined?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+            id: string;
+            name: string;
+            displayName: string;
+        };
+        challenge: string;
+        timeout: number;
+        excludeCredentials: {
+            id: string;
+            type: string;
+            transports: ("ble" | "hybrid" | "internal" | "nfc" | "usb")[];
```
rishabhpoddar, line 106: how is the value of type and transports defined?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+        shouldTryLinkingWithSessionUser: boolean | undefined;
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+              user: User;
+              recipeUserId: RecipeUserId;
+          }
+        | { status: "EMAIL_ALREADY_EXISTS_ERROR" }
```
rishabhpoddar, line 159: What about status of webauthnGeneratedOptionsId not existing? Or something going wrong its verification?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+        };
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+              user: User;
+              recipeUserId: RecipeUserId;
+          }
+        | { status: "EMAIL_ALREADY_EXISTS_ERROR" }
```
rishabhpoddar, line 197: See my comments in the sign up recipe function

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+              status: "LINKING_TO_SESSION_USER_FAILED";
+              reason:
+                  | "EMAIL_VERIFICATION_REQUIRED"
+                  | "RECIPE_USER_ID_ALREADY_LINKED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR"
+                  | "ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR"
+                  | "SESSION_USER_ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR";
+          }
+    >;
+
+    // todo uncomment one by one once starting implementation
```
rishabhpoddar, line 251: Please uncomment all now. Types need to be very clear from right at the start.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,520 @@
...
+    req: BaseRequest;
+    res: BaseResponse;
+    emailDelivery: EmailDeliveryIngredient<TypeWebauthnEmailDeliveryInput>;
+};
+
+export type APIInterface = {
+    registerOptionsPOST:
+        | undefined
+        | ((input: {
+              email: string;
```
rishabhpoddar, line 309:
```suggestion
              email?: string; recoveryToken?: string; session?: SessionContainer
```

rishabhpoddar (changes requested):

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: the transports value is set by the authenticator and then stored along with the credentials (see .
the value will always be 'public-key' in our case (i updated the type)- see https://developer.mozilla.org/en-US/docs/Web/API/Credential/type. 
Done

lib/ts/recipe/webauthn/types.ts
victorbojica: added them in the input and will add defaults for them. Done

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: it's the user email for both, as we don't have first/last name as inputs and we shouldn't disclose them either even if we would have them stored.

"The WebAuthn User Display Name (user.displayName) is a user-chosen, readable name for display in the app interface, not for authentication. It enhances user experience by providing a familiar, human-readable identifier, like a real name or nickname. " 

"The WebAuthn User Name (user.name) is a unique identifier within the RP’s system, crucial in scenarios where users might have similar Display Names. It’s used in the authentication process and can be an email, username, or any unique ID, ensuring each account is distinctively linked to its credentials. This name is vital for maintaining authentication integrity, especially when different accounts have similar Display Names. It can be any arbitrary string, offering flexibility in user identification."

See https://medium.com/@corbado_tech/webauthn-credential-id-user-name-user-id-user-handle-afe248a1d7c5#:~:text=The%20WebAuthn%20User%20Display%20Name,a%20real%20name%20or%20nickname.

lib/ts/recipe/webauthn/types.ts
victorbojica: yes. done.

lib/ts/recipe/webauthn/types.ts
victorbojica: done.

lib/ts/recipe/webauthn/types.ts
victorbojica: Relying party id: 
If they don't pass anything, we will use the origin. 
If they pass in a string, we will use that. 
If they pass a function, they might want to make use of the request details such as origin, headers maybe, etc). 
They might want to use dynamically when running multiple tenants on different domains.

Relying party name:
i can drop the inputs and only keep the function without any params. it will default to app name.

Done

lib/ts/recipe/multifactorauth/types.ts
victorbojica: > Relate to that, will all webauthn methods have an email associated with that login method?

don't think i understand.

> need to also modify the user object and login methods object (in ts/types.ts)
Isn't this enough? https://github.com/supertokens/supertokens-node/pull/951/files#diff-f8b1deda227dcaaaf5169c750f94e6e49ad2660d1064b9b46693d708a33e862fL198
As far as i can tell this is the only change needed .

PR #952, opened 2024-10-17
feat: WebAuthN Sign In, Sign Up and Options methods support - NEW

victorbojica:
ONLY REVIEW REQUIRED. NOTHING TESTED YET.

## Summary of change
Implement WebAuthN support according to:
https://docs.google.com/document/d/1G7tO9_dSNi8wur3ajGg4pq-wiHatKDbHv2sBt-uSbQg/edit#heading=h.olee876uqu8a

## Related
Check https://github.com/supertokens/supertokens-node/pull/942 for more feedback

## Test Plan
No testing at the moment.

## Documentation changes

Will have to add WebAuthN recipe documentation

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Check support for other authenticators
-   [ ] Account recovery flow
-   [ ] Options testing
-   [ ] Sign In testing
-   [ ] Sign Up testing
-   [ ] Multiple authenticators testing
-   [ ] Do we need verifyCredentials in recipeInterface?


lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+            origin: string;
+            requireResidentKey: boolean | undefined; // should default to false in order to allow multiple authenticators to be used; see https://auth0.com/blog/a-look-at-webauthn-resident-credentials/
+            // default to 'required' in order store the private key locally on the device and not on the server
+            residentKey: "required" | "preferred" | "discouraged" | undefined;
+            // default to 'preferred' in order to verify the user (biometrics, pin, etc) based on the device preferences
+            userVerification: "required" | "preferred" | "discouraged" | undefined;
+            // default to 'none' in order to allow any authenticator and not verify attestation
+            attestation: "none" | "indirect" | "direct" | "enterprise" | undefined;
+            // default to 5 seconds
+            timeout: number | undefined;
```
rishabhpoddar, line 90:
```suggestion
            timeout?: number;
```

rishabhpoddar: Prefer optional cause otherwise users have to explicitly set it to undefined.

lib/ts/recipe/multifactorauth/types.ts
```diff
@@ -154,6 +154,7 @@ export type GetPhoneNumbersForFactorsFromOtherRecipesFunc = (
 
 export const FactorIds = {
     EMAILPASSWORD: "emailpassword",
+    WEBAUTHN: "webauthn",
```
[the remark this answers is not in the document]
rishabhpoddar: You need to modufy the user object.. For example, we are adding pass keys credentials for them. So now login methods need to have passkeys credentials ids? And the root of the user object needs to also have a passkeys credentials ids array? What else needs to be added?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        apis: (originalImplementation: APIInterface, builder?: OverrideableBuilder<APIInterface>) => APIInterface;
+    };
+};
+
+export type TypeNormalisedInputRelyingPartyId = (input: {
+    request: BaseRequest | undefined;
+    userContext: UserContext;
+}) => string; // should return the domain of the origin
+
+export type TypeNormalisedInputRelyingPartyName = () => string; // should return the app name
```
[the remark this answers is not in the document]
rishabhpoddar: You need to add the tenant id as an input param to these? Also, these should return Promise<string> so as to allow devs to do network calls

lib/ts/recipe/webauthn/types.ts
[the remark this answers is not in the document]
rishabhpoddar: Make it `?` instead of `| undefined`. Cause then devs have to explicitly pass in undefined.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+            relyingPartyId: string;
+            relyingPartyName: string;
+            origin: string;
+            requireResidentKey: boolean | undefined; // should default to false in order to allow multiple authenticators to be used; see https://auth0.com/blog/a-look-at-webauthn-resident-credentials/
+            // default to 'required' in order store the private key locally on the device and not on the server
+            residentKey: "required" | "preferred" | "discouraged" | undefined;
+            // default to 'preferred' in order to verify the user (biometrics, pin, etc) based on the device preferences
+            userVerification: "required" | "preferred" | "discouraged" | undefined;
+            // default to 'none' in order to allow any authenticator and not verify attestation
+            attestation: "none" | "indirect" | "direct" | "enterprise" | undefined;
```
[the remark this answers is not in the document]
rishabhpoddar: If they have to be set to a non default value, who will set them? For example, the frontend making the request to the api will set them based on the type of device? If the frontend will set them, then how will it know what to set it to?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+              webauthnGeneratedOptionsId: string;
+              rp: {
+                  id: string;
+                  name: string;
+              };
+              user: {
+                  id: string;
+                  name: string; // user email
+                  displayName: string; //user email
+              };
```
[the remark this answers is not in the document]
rishabhpoddar: It says `The WebAuthn User Name (user.name) is a unique identifier within the RP’s system` - but email is not a unique ID, since many passkeys (belonging to the same user) can have the same email?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+              }[];
+              authenticatorSelection: {
+                  requireResidentKey: boolean;
+                  residentKey: "required" | "preferred" | "discouraged";
+                  userVerification: "required" | "preferred" | "discouraged";
+              };
+          }
+        | { status: "EMAIL_MISSING_ERROR" } // email is required if not using session or recoverAccountToken
+        | { status: "SESSION_MISSING_ERROR" } // session is required if not using email or recoverAccountToken
+        | { status: "RECOVER_ACCOUNT_TOKEN_MISSING_ERROR" } // recoverAccountToken is required if not using email or session
```
rishabhpoddar, line 138: These can be general JS exceptions since this is a programming error if none of them are provided.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
+        | { status: "RELYING_PARTY_ID_MISSING_ERROR" }
```
rishabhpoddar, line 140: Why would it be missing? The input type of this function is `string` for it, right?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        | { status: "RELYING_PARTY_ID_INVALID_ERROR" } // wrong format; must be domain
```
rishabhpoddar, line 141: This is also a programming error.. so can just throw a JS exception

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        | { status: "RELYING_PARTY_NAME_MISSING_ERROR" } // relyingPartyName is required
```
rishabhpoddar, line 142: Why would it be missing? The input type has it as a string

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        | { status: "ORIGIN_MISSING_ERROR" }
+        | { status: "ORIGIN_INVALID_ERROR" } // wrong format; must be domain url
```
rishabhpoddar, line 144: Why would it be missing? Input type has it as a string. Also, invalid error should just throw a JS error.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+              status: "OK";
+              webauthnGeneratedOptionsId: string;
+              challenge: string;
+              timeout: number;
+              userVerification: "required" | "preferred" | "discouraged";
+          }
+        | { status: "RELYING_PARTY_ID_MISSING_ERROR" }
+        | { status: "RELYING_PARTY_ID_INVALID_ERROR" } // wrong format; must be domain
+        | { status: "ORIGIN_MISSING_ERROR" }
+        | { status: "ORIGIN_INVALID_ERROR" } // wrong format; must be domain url
```
rishabhpoddar, line 165: see comments for registerOptions output please.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+              user: User;
+              recipeUserId: RecipeUserId;
+          }
+        | { status: "EMAIL_ALREADY_EXISTS_ERROR" }
+        | { status: "WRONG_CREDENTIALS_ERROR" }
+        | { status: "CREDENTIAL_MISSING_ERROR" }
```
rishabhpoddar, line 195: Do we need to distinguish between wrong credentials vs credentials missing? Does the frontend need to react in a different way for these two types of outputs?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        | { status: "GENERATED_OPTIONS_ID_MISSING_ERROR" }
```
rishabhpoddar, line 196: Whats the diff between GENERATED_OPTIONS_ID_MISSING_ERROR and CREDENTIAL_MISSING_ERROR?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+        };
+        session: SessionContainerInterface | undefined;
+        shouldTryLinkingWithSessionUser: boolean | undefined;
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | { status: "OK"; user: User; recipeUserId: RecipeUserId }
+        | { status: "WRONG_CREDENTIALS_ERROR" }
+        | { status: "CREDENTIAL_MISSING_ERROR" }
+        | { status: "GENERATED_OPTIONS_ID_MISSING_ERROR" }
```
rishabhpoddar, line 230: see comment in sign up.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+              userId: string;
+          }
+        | { status: "WRONG_CREDENTIALS_ERROR" }
+        | { status: "CREDENTIAL_MISSING_ERROR" }
+        | { status: "GENERATED_OPTIONS_ID_MISSING_ERROR" }
+        | { status: "RECOVER_ACCOUNT_TOKEN_MISSING_ERROR" }
+        | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
+    >;
+
+    // used internally for creating a passkey during password reset flow or when adding a credential to an existing user
```
rishabhpoddar, line 287: password reset flow?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+              user: User;
+              recipeUserId: RecipeUserId;
+          }
+        | { status: "EMAIL_ALREADY_EXISTS_ERROR" }
+        | { status: "WRONG_CREDENTIALS_ERROR" }
+        | { status: "CREDENTIAL_MISSING_ERROR" }
+        | { status: "GENERATED_OPTIONS_ID_MISSING_ERROR" }
+    >;
+
+    verifyCredentials(input: {
```
rishabhpoddar, line 352: this function has identical func signature as registerPasskeys. Are both, this and the other function needed?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,625 @@
...
+    // used for retrieving the user details (email) from the recover account token
+    // should be used in the registerOptions function when the user recovers the account and generates the credentials
+    getUserFromRecoverAccountToken(input: {
+        token: string;
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | { status: "OK"; user: User; recipeUserId: RecipeUserId }
+        | { status: "RECOVER_ACCOUNT_TOKEN_MISSING_ERROR" }
+        | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
```
rishabhpoddar, line 385: Can just be one type.

rishabhpoddar (changes requested):

lib/ts/recipe/webauthn/types.ts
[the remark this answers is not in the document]
victorbojica: i will, but this is how i've seen it done with `session` and `shouldTryLinkingWithSessionUser`. Should i change those as well ?

lib/ts/recipe/webauthn/types.ts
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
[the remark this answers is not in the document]
victorbojica: this the front should not be allowed to do such things. the developer will set them by overriding the methods - same as for the timeout

lib/ts/recipe/webauthn/types.ts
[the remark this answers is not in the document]
victorbojica: it's fine to have multiple passkeys. it's not ok to have multiple users with same email.

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: yes. one logs in the user, the other adds a credential to the database

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done. those are exceptions

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done. credentials missing will be an exception

lib/ts/recipe/webauthn/types.ts
victorbojica: done. will be an exception

lib/ts/recipe/webauthn/types.ts
victorbojica: done. will be an exception

lib/ts/recipe/webauthn/types.ts
victorbojica: done. will be an exception

lib/ts/recipe/webauthn/types.ts
victorbojica: done. will be an exception

lib/ts/recipe/webauthn/types.ts
victorbojica: done. will be an exception

lib/ts/recipe/webauthn/types.ts
victorbojica: done. will be an exception

lib/ts/recipe/webauthn/types.ts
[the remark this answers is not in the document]
victorbojica: nothing to be done

lib/ts/recipe/webauthn/types.ts
victorbojica: nothing to be done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,608 @@
...
+    // used for retrieving the user details (email) from the recover account token
+    // should be used in the registerOptions function when the user recovers the account and generates the credentials
+    getUserFromRecoverAccountToken(input: {
+        token: string;
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        { status: "OK"; user: User; recipeUserId: RecipeUserId } | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
+    >;
+};
```
rishabhpoddar, line 377: need more functions to have CRUD operation on credential IDs? How will a

rishabhpoddar (changes requested):

lib/ts/recipe/multifactorauth/types.ts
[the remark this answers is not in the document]
victorbojica: Yeah. So after discussing went over the code and schema again and decided to add a webauthn object containing the credential ids. Besides this, i've also added the relying party id to the credentials schema (see here: https://docs.google.com/document/d/1G7tO9_dSNi8wur3ajGg4pq-wiHatKDbHv2sBt-uSbQg/edit#heading=h.td4zxgo1675k) because we don't have a way of knowing what credentials are available for a relying party id. 

Also added some more info for the CDI docs to include the webauthn info in the user object (see here: https://docs.google.com/document/d/1G7tO9_dSNi8wur3ajGg4pq-wiHatKDbHv2sBt-uSbQg/edit#heading=h.r1r1xaejv0ga).

Done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,608 @@
...
+            userContext: UserContext;
+        } & (
+            | {
+                  recoverAccountToken: string;
+              }
+            | {
+                  email: string;
+              }
+            | {
+                  session: SessionContainerInterface;
```
rishabhpoddar, line 114: when is only a session needed? And how will we get the email from the session? What if the currently logging in user has no email, or what if the currently logged in user has multiple emails (cause they are a primary user with multiple emails)

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,608 @@
...
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+              user: User;
+              recipeUserId: RecipeUserId;
+          }
+        | { status: "EMAIL_ALREADY_EXISTS_ERROR" }
+        // when the attestation is checked and is not valid or other cases in whcih the authenticator is not correct
+        | { status: "INVALID_AUTHENTICATOR_ERROR" }
```
rishabhpoddar, line 194: is there nothing else we need to tell the user when this error happens? Like why was the authenticator invalid?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,608 @@
...
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+              email: string;
+              userId: string;
+          }
+        | { status: "WRONG_CREDENTIALS_ERROR" }
+        | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
```
rishabhpoddar, line 280: shouldnt this also have `INVALID_AUTHENTICATOR_ERROR` as an output?

lib/ts/recipe/webauthn/types.ts
rishabhpoddar: Thats too high level of an explanation.. what will be the exact diff between the implementation of them? Its not like the signIn will create a session - so it doesn't "log in" the user.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,608 @@
...
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+              user: User;
+              recipeUserId: RecipeUserId;
+          }
+        | { status: "EMAIL_ALREADY_EXISTS_ERROR" }
+        | { status: "WRONG_CREDENTIALS_ERROR" }
```
rishabhpoddar, line 342: need `INVALID_AUTHENTICATOR_ERROR` here?

lib/ts/recipe/webauthn/types.ts
rishabhpoddar: @victorbojica, you missed this comment?

rishabhpoddar (changes requested):

lib/ts/recipe/webauthn/types.ts
victorbojica: If there is no email, it will throw. I've added the error. As for when there are multiple emails i was thinking of only using the email from the webauthn login method. Basically allowing register passkey to be used if the user is logged in with webauthn. If there is no session, they should pass one of the other options...

What do you think ?

lib/ts/recipe/webauthn/types.ts
victorbojica: I was thinking that when this kind of error happens, we inform the user to try using another authenticator: "The authenticator is not supported, please try a different one" I could add some reasons tough, but they might be too technical for the users: "Not supported supported authenticator. Try using an Apple device, Android device or supported FIDO authenticator."

lib/ts/recipe/webauthn/types.ts
victorbojica: They do different things. It's like sign up vs sign in in `emailpassword`. They both take a password and an email but do different things. 

I replicated what was on the emailpassword recipe: 
https://github.com/supertokens/supertokens-node/blob/ced57b11260144d1e1b89039db3c1bdbbc258318/lib/ts/recipe/emailpassword/types.ts#L148

registerCredentials (registerPasskeys) will be used on sign up and for adding a passkey to a logged in user.

Am i missing something ?

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
rishabhpoddar: So verifyCredentials can only be used by options created from signInOptions, whereas registerPasskeys can only be used by options created from registerOptions?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,735 @@
...
+        | {
+              status: "OK";
+          }
+        | AddCredentialErrorResponse
+    >;
+
+    // credentials CRUD
+    removeCredential(input: {
+        webauthnCredentialId: string;
+        tenantId: string;
```
rishabhpoddar, line 423: why does this take a tenant id? credentials are specific to users, and not tenant?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,735 @@
...
+    getUserFromRecoverAccountToken(input: {
+        token: string;
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<{ status: "OK"; user: User; recipeUserId: RecipeUserId } | GetUserFromRecoverAccountTokenErrorResponse>;
+
+    // credentials CRUD
+
+    // this will call registerCredential internally
+    addCredential(input: {
```
rishabhpoddar, line 396: no need to have this since we alredy have regiusterCredentials

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,735 @@
...
+        };
+        tenantId: string;
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+          }
+        | AddCredentialErrorResponse
+    >;
+
```
rishabhpoddar, line 419: need a function to get credentials?

rishabhpoddar: Also need a function to update credentials? Like change name or something like that?

lib/ts/recipe/multitenancy/types.ts
```diff
@@ -173,6 +173,9 @@ export type APIInterface = {
               passwordless: {
                   enabled: boolean;
               };
+              webauthn: {
+                  credentialIds: string[];
+              };
```
rishabhpoddar, line 178: no need for this. There is already a firstFactors array output which will contain webauthn id if its enabled as first factor.

lib/ts/recipe/multifactorauth/types.ts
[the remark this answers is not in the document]
rishabhpoddar: There are no type changes here.. just thje implmenetation of the api has to change. So keeping this comment open

lib/ts/recipe/webauthn/types.ts
rishabhpoddar: I dont think its clear when a session will be an input.. it could mean that the user wants to add a passkey to an existing logged in account. In that case, there could be already a webauthn login method, or there may not be any at all.. so we cant just throw in case an email doesn't exist, and we can't just assume webauthn login method is always there. We need clarity on the flow here! Please think about this case carefully.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -84,6 +84,37 @@ export type TypeInputGetOrigin = (input: {
...
+// centralize error types in order to prevent missing cascading errors
+type RegisterCredentialErrorResponse =
+    | { status: "WRONG_CREDENTIALS_ERROR" }
+    // when the attestation is checked and is not valid or other cases in whcih the authenticator is not correct
+    | { status: "INVALID_AUTHENTICATOR_ERROR" };
+
+type VerifyCredentialsErrorResponse =
+    | { status: "WRONG_CREDENTIALS_ERROR" }
+    // when the attestation is checked and is not valid or other cases in which the authenticator is not correct
+    | { status: "INVALID_AUTHENTICATOR_ERROR" };
```
rishabhpoddar, line 96: why do we need two different variable to have the same type?

rishabhpoddar (changes requested):

lib/ts/recipe/webauthn/types.ts
victorbojica: credentials can't be updated. only removed and added. done.

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done.

victorbojica: yes. they are specific to a user.

lib/ts/recipe/webauthn/types.ts
victorbojica: i think i'm missing context then. doesn't the existence of webauthn in the loginMethods mean that the webauth loging method exists ? Maybe we can chat

lib/ts/recipe/webauthn/types.ts
victorbojica: yes. but maybe let's chat.

lib/ts/recipe/multitenancy/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: we decided to skip using the session

lib/ts/recipe/multifactorauth/types.ts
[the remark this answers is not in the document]
rishabhpoddar: @victorbojica , you missed this comment?

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,858 @@
...
+    }): Promise<
+        | {
+              status: "OK";
+              credential: {
+                  id: string;
+                  rp_id: string;
+                  created_at: number;
+              };
+          }
+        | RemoveCredentialErrorResponse
```
rishabhpoddar, line 488: why does the get function return a RemoveCredentialErrorResponse?

rishabhpoddar: Also, this function should return the recipeUserId that uses this credential as well.

rishabhpoddar (changes requested):

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+type GenerateRecoverAccountTokenPOSTErrorResponse = {
+    status: "ACCOUNT_RECOVERY_NOT_ALLOWED";
+    reason: string;
+};
+
+type RecoverAccountPOSTErrorResponse =
+    | {
+          status: "ACCOUNT_RECOVERY_NOT_ALLOWED";
+          reason: string;
+      }
```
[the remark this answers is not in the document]
victorbojica: done (both)

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+// centralize error types in order to prevent missing cascading errors
+type RegisterCredentialErrorResponse =
+    | { status: "WRONG_CREDENTIALS_ERROR" }
+    // when the attestation is checked and is not valid or other cases in whcih the authenticator is not correct
+    | { status: "INVALID_AUTHENTICATOR_ERROR"; reason: string };
+
+type VerifyCredentialsErrorResponse =
+    | { status: "WRONG_CREDENTIALS_ERROR" }
+    // when the attestation is checked and is not valid or other cases in which the authenticator is not correct
+    | { status: "INVALID_AUTHENTICATOR_ERROR" };
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+              userContext: UserContext;
+          }) => Promise<
+              | {
+                    status: "OK";
+                    user: User;
+                    email: string;
+                }
+              | GeneralErrorResponse
+              | RecoverAccountTokenPOSTErrorResponse
+          >);
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+                    status: "OK";
+                    credential: {
+                        id: string;
+                        rp_id: string;
+                        created_at: number;
+                    };
+                }
+              | GetCredentialGETErrorResponse
+              | GeneralErrorResponse
+          >);
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+                        type: string;
+                    }[];
+                    authenticatorSelection: {
+                        requireResidentKey: boolean;
+                        residentKey: "required" | "preferred" | "discouraged";
+                        userVerification: "required" | "preferred" | "discouraged";
+                    };
+                }
+              | GeneralErrorResponse
+              | RegisterOptionsPOSTErrorResponse
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+}) => Promise<string>;
+
+export type TypeNormalisedInputGetOrigin = (input: {
+    tenantId: string;
+    request: BaseRequest;
+    userContext: UserContext;
+}) => Promise<string>; // should return the app name
+
+export type TypeInput = {
+    emailDelivery?: EmailDeliveryTypeInput<TypeWebauthnEmailDeliveryInput>;
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+                    };
+                }
+              | GeneralErrorResponse
+              | RegisterOptionsPOSTErrorResponse
+          >);
+
+    signInOptionsPOST:
+        | undefined
+        | ((input: {
+              tenantId: string;
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+type VerifyCredentialsErrorResponse =
+    | { status: "WRONG_CREDENTIALS_ERROR" }
+    // when the attestation is checked and is not valid or other cases in which the authenticator is not correct
+    | { status: "INVALID_AUTHENTICATOR_ERROR" };
+
+type CreateNewRecipeUserErrorResponse = RegisterCredentialErrorResponse | { status: "EMAIL_ALREADY_EXISTS_ERROR" };
+
+type GetUserFromRecoverAccountTokenErrorResponse = { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" };
+
+type RegisterOptionsErrorResponse = GetUserFromRecoverAccountTokenErrorResponse | { status: "EMAIL_MISSING_ERROR" };
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+          }) => Promise<
+              | {
+                    status: "OK";
+                    webauthnGeneratedOptionsId: string;
+                    challenge: string;
+                    timeout: number;
+                    userVerification: "required" | "preferred" | "discouraged";
+                }
+              | GeneralErrorResponse
+              | SignInOptionsPOSTErrorResponse
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+          >);
+
+    signUpPOST:
+        | undefined
+        | ((input: {
+              email: string;
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+
+type RemoveCredentialErrorResponse = { status: "CREDENTIAL_NOT_FOUND_ERROR" };
+
+type GetCredentialErrorResponse = { status: "CREDENTIAL_NOT_FOUND_ERROR" };
+
+type DecodeCredentialErrorResponse = { status: "WRONG_CREDENTIALS_ERROR" };
+
+type Base64URLString = string;
+
+export type RecipeInterface = {
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -606,30 +673,17 @@ export type APIInterface = {
...
-                    status: "OK";
-                    email: string;
-                    user: User;
+                    status: "RECOVER_ACCOUNT_NOT_ALLOWED";
+                    reason: string;
                 }
-              | RecoverAccountPOSTErrorResponse
-              | GeneralErrorResponse
+              | { status: "UNKNOWN_USER_ID_ERROR" }
+              | { status: "INVALID_EMAIL_ERROR" }
```
rishabhpoddar, line 683: Both of these are incorrect statuses to return:
- The input is not a user id, so we can't return UNKNOWN_USER_ID_ERROR
- We do not need to check email validation here cause we are not adding a new email to any user based on this api.
- We only return RECOVER_ACCOUNT_NOT_ALLOWED and OK. In case user with the email is not found, we still return OK cause we do not want to reveal info about user's existing or not. We just dont send an email - similar to what we do in password reset API when we create a new token

rishabhpoddar (changes requested):

lib/ts/recipe/webauthn/types.ts
```diff
@@ -532,12 +587,15 @@ export type APIInterface = {
                     };
                 }
               | GeneralErrorResponse
-              | RegisterOptionsPOSTErrorResponse
+              //   | RegisterOptionsPOSTErrorResponse
+              | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
+              | { status: "INVALID_EMAIL_ERROR" }
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -85,36 +92,52 @@ export type TypeInputGetOrigin = (input: {
...
+export type TypeInputValidateEmailAddress = (
+    email: string,
+    tenantId: string
+) => Promise<string | undefined> | string | undefined;
+
 // centralize error types in order to prevent missing cascading errors
-type RegisterCredentialErrorResponse =
+type RegisterOptionsErrorResponse =
+    | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
+    | { status: "INVALID_EMAIL_ERROR" };
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -411,10 +467,42 @@ export type RecipeInterface = {
...
+              status: "OK";
+              id: string;
+              relyingPartyId: string;
+              relyingPartyName: string;
+              origin: string;
+              userName: string;
+              userDisplayName: string;
+              timeout: string;
+              attestation: string;
+              challenge: string;
```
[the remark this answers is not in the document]
victorbojica: done. was using the userName, but i changed to be email - in the schema also.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -550,13 +608,14 @@ export type APIInterface = {
                     userVerification: "required" | "preferred" | "discouraged";
                 }
               | GeneralErrorResponse
-              | SignInOptionsPOSTErrorResponse
+              //   | SignInOptionsPOSTErrorResponse
+              | { status: "WRONG_CREDENTIALS_ERROR" }
+              | { status: "INVALID_EMAIL_ERROR" }
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
```diff
@@ -492,12 +488,9 @@ export type RecipeInterface = {
...
               id: string;
               relyingPartyId: string;
-              relyingPartyName: string;
               origin: string;
-              userName: string;
-              userDisplayName: string;
+              email: string;
               timeout: string;
-              attestation: string;
               challenge: string;
```
rishabhpoddar, line 494: why did we remove relyingPartyName, userName, userDisplayName and attestation?

rishabhpoddar (changes requested):

lib/ts/recipe/webauthn/types.ts
victorbojica: I checked the schema and we don't store them in the generated options. `userName` is the email and i renamed it to be more simple.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,768 @@
...
+export type APIInterface = {
+    registerOptionsPOST:
+        | undefined
+        | ((
+              input: {
+                  tenantId: string;
+                  options: APIOptions;
+                  userContext: UserContext;
+              } & ({ email: string } | { recoverAccountToken: string })
+          ) => Promise<
```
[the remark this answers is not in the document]
victorbojica: Yes, through WebViews starting from iOS 16 and Android 9 (and 14 in some cases): https://passkeys.dev/docs/reference/android/ and https://passkeys.dev/docs/reference/ios/.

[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/types.ts
rishabhpoddar: So ideally, the output of registerOptions function should be similar to the output of this function. For example, in the registerOptions output, we have:
```
user: {
                  id: string;
                  name: string; // user email
                  displayName: string; //user email
              };
```

Now in the output of getGeneratedOptions, it's totally different. So this is strange.

victorbojica: Because we only store what is needed to verify the credentials. See the schema here (`webauthn_user_generated_options`) https://docs.google.com/document/d/1G7tO9_dSNi8wur3ajGg4pq-wiHatKDbHv2sBt-uSbQg/edit?tab=t.0#heading=h.td4zxgo1675k

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,747 @@
...
+
+// type RemoveGeneratedOptionsErrorResponse = { status: "GENERATED_OPTIONS_NOT_FOUND_ERROR" };
+
+// type GetGeneratedOptionsErrorResponse = { status: "GENERATED_OPTIONS_NOT_FOUND_ERROR" };
+
+type Base64URLString = string;
+
+export type RecipeInterface = {
+    // should have a way to access the user email: passed as a param, through session, or using recoverAccountToken
+    // it should have at least one of those 3 options
```
porcellus, line 144: The comment seems off

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,747 @@
...
+    TypeInput as EmailDeliveryTypeInput,
+    TypeInputWithService as EmailDeliveryTypeInputWithService,
+} from "../../ingredients/emaildelivery/types";
+import EmailDeliveryIngredient from "../../ingredients/emaildelivery";
+import { GeneralErrorResponse, NormalisedAppinfo, User, UserContext } from "../../types";
+import RecipeUserId from "../../recipeUserId";
+
+export type TypeNormalisedInput = {
+    relyingPartyId: TypeNormalisedInputRelyingPartyId;
+    relyingPartyName: TypeNormalisedInputRelyingPartyName;
```
porcellus, line 29: since these are functions, I'd prefer if these were prefixed with `get`

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,747 @@
...
+    // should have a way to access the user email: passed as a param, through session, or using recoverAccountToken
+    // it should have at least one of those 3 options
+    registerOptions(
+        input: {
+            relyingPartyId: string;
+            relyingPartyName: string;
+            origin: string;
+            requireResidentKey: boolean | undefined; // should default to false in order to allow multiple authenticators to be used; see https://auth0.com/blog/a-look-at-webauthn-resident-credentials/
+            // default to 'required' in order store the private key locally on the device and not on the server
+            residentKey: "required" | "preferred" | "discouraged" | undefined;
```
porcellus, line 152: Since these lists seem to be reused in a few places (e.g.: both in the input and the return types) I'd prefer if you moved made these into re-usable types.

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,747 @@
...
+              status: "OK";
+              webauthnGeneratedOptionsId: string;
+              rp: {
+                  id: string;
+                  name: string;
+              };
+              user: {
+                  id: string;
+                  name: string; // user email
+                  displayName: string; //user email
```
porcellus, line 182: Is this required by a standard? If so, please add a comment saying that

lib/ts/recipe/webauthn/api/implementation.ts
```diff
@@ -0,0 +1,1095 @@
...
+                LINKING_TO_SESSION_USER_FAILED: {
+                    EMAIL_VERIFICATION_REQUIRED:
+                        "Cannot sign in / up due to security reasons. Please contact support. (ERR_CODE_013)",
+                    RECIPE_USER_ID_ALREADY_LINKED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR:
+                        "Cannot sign in / up due to security reasons. Please contact support. (ERR_CODE_014)",
+                    ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR:
+                        "Cannot sign in / up due to security reasons. Please contact support. (ERR_CODE_015)",
+                    SESSION_USER_ACCOUNT_INFO_ALREADY_ASSOCIATED_WITH_ANOTHER_PRIMARY_USER_ID_ERROR:
+                        "Cannot sign in / up due to security reasons. Please contact support. (ERR_CODE_016)",
+                },
```
porcellus, line 234: I think the error codes will have to be updated. You can add a TODO here and wherever an update like this is pending.

lib/ts/recipe/webauthn/api/implementation.ts
```diff
@@ -0,0 +1,1095 @@
...
+            // check for type is done in a parent function but they are kept
+            // here to be on the safe side.
+            if (!email) {
+                throw new Error(
+                    "Should never come here since we already check that the email value is a string in validateFormFieldsOrThrowError"
+                );
+            }
+
+            const preAuthCheckRes = await AuthUtils.preAuthChecks({
+                authenticatingAccountInfo: {
```
porcellus, line 258: shouldn't this have the credentialId? or is that something we only get later?

lib/ts/recipe/webauthn/api/implementation.ts
```diff
@@ -0,0 +1,1095 @@
...
+            // do the verification before in order to retrieve the user email
+            const verifyCredentialsResponse = await options.recipeImplementation.verifyCredentials({
+                credential,
+                webauthnGeneratedOptionsId,
+                tenantId,
+                userContext,
+            });
+            const checkCredentialsOnTenant = async () => {
+                return verifyCredentialsResponse.status === "OK";
+            };
```
porcellus, line 403: Is this correct? is credential validation independent of the tenant?

lib/ts/recipe/webauthn/api/implementation.ts
```diff
@@ -0,0 +1,1095 @@
...
+                    );
+                }
+                email = loginMethod?.email;
+            } else {
+                return {
+                    status: "WRONG_CREDENTIALS_ERROR",
+                };
+            }
+
+            const authenticatingUser = await AuthUtils.getAuthenticatingUserAndAddToCurrentTenantIfRequired({
```
porcellus, line 424: this function also needs to be updated to take the new account info prop into account (plus I think you'll need to check all places where we use the account info)

lib/ts/recipe/webauthn/types.ts
```diff
@@ -0,0 +1,747 @@
...
+        userContext: UserContext;
+    }): Promise<
+        | {
+              status: "OK";
+              user: User;
+              recipeUserId: RecipeUserId;
+          }
+        // | SignUpErrorResponse
+        | { status: "EMAIL_ALREADY_EXISTS_ERROR" }
+        | { status: "WRONG_CREDENTIALS_ERROR" }
```
porcellus, line 243: This feels weird on a `signUp` endpoint, maybe specify which param is wrong/how?

porcellus (changes requested):

lib/ts/recipe/accountlinking/types.ts
```diff
@@ -192,10 +192,13 @@ export type AccountInfo = {
         id: string;
         userId: string;
     };
+    webauthn?: {
+        credentialIds: string[];
```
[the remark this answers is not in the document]
victorbojica: I haven't taken that into consideration. But as far as ai can tell, no. there is no use like that. There might the possibllity to search by one of the credentialIds. What do you think?

lib/ts/recipe/webauthn/api/recoverAccount.ts
```diff
@@ -0,0 +1,72 @@
...
+import { APIInterface, APIOptions } from "../";
+import { UserContext } from "../../../types";
+
+export default async function recoverAccount(
+    apiImplementation: APIInterface,
+    tenantId: string,
+    options: APIOptions,
+    userContext: UserContext
+): Promise<boolean> {
+    // Logic as per https://github.com/supertokens/supertokens-node/issues/22#issuecomment-710512442
```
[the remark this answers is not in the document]
victorbojica: makes sense. done

lib/ts/recipe/webauthn/api/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+import STError from "../error";
+
+export async function validatewebauthnGeneratedOptionsIdOrThrowError(
```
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/api/signup.ts
```diff
@@ -0,0 +1,86 @@
...
+        throw new STError({
+            type: STError.FIELD_ERROR,
+            payload: [
+                {
+                    id: "email",
+                    error: "This email already exists. Please sign in instead.",
+                },
+            ],
+            message: "Error in input formFields",
+        });
```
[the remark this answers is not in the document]
victorbojica: done (also see first comment about STError)

lib/ts/recipe/webauthn/api/implementation.ts
```diff
@@ -0,0 +1,1095 @@
...
+                },
+            };
+
+            const generatedOptions = await options.recipeImplementation.getGeneratedOptions({
+                webauthnGeneratedOptionsId,
+                tenantId,
+                userContext,
+            });
+            if (generatedOptions.status !== "OK") {
+                return { status: "WRONG_CREDENTIALS_ERROR" };
```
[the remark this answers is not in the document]
victorbojica: The status is a generic one, for whatever error happens because of the Webauthn credentials flow. We could use something more specific, but it wouldn't  really make sense to the end user (they would understand and they couldn't act in any way on it).

Also, as a general note: The credential in the Webauthn recipe represents what is generated by the browser for the webauthn flow (public key). See: https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions.

What do you think ?

lib/ts/recipe/webauthn/api/implementation.ts
```diff
@@ -0,0 +1,1095 @@
...
+            }
+
+            const email = generatedOptions.email;
+
+            // NOTE: Following checks will likely never throw an error as the
+            // check for type is done in a parent function but they are kept
+            // here to be on the safe side.
+            if (!email) {
+                throw new Error(
+                    "Should never come here since we already check that the email value is a string in validateFormFieldsOrThrowError"
```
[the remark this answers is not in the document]
victorbojica: As for the form fields, we are not using them for this recipe.
The validation of the email is happening in the registerOptions. See: https://github.com/supertokens/supertokens-node/blob/feat/webauthn/basic-methods/lib/ts/recipe/webauthn/api/registerOptions.ts
The reason we don't validate here, is that we assume that the email has already been validated in the registerOptions method.

lib/ts/recipe/webauthn/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+    let override = {
+        functions: (originalImplementation: RecipeInterface) => originalImplementation,
+        apis: (originalImplementation: APIInterface) => originalImplementation,
+        ...config?.override,
+    };
+
+    function getEmailDeliveryConfig(isInServerlessEnv: boolean) {
+        let emailService = config?.emailDelivery?.service;
+        console.log("emailService", emailService);
+        console.log("isInServerlessEnv", isInServerlessEnv);
```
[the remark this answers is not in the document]
victorbojica: done.

lib/ts/recipe/webauthn/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+        relyingPartyId,
+        relyingPartyName,
+        validateEmailAddress,
+        getEmailDeliveryConfig,
+    };
+}
+
+function validateAndNormaliseRelyingPartyIdConfig(
+    _: Recipe,
+    __: NormalisedAppinfo,
```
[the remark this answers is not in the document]
victorbojica: done.

lib/ts/recipe/webauthn/index.ts
```diff
@@ -0,0 +1,408 @@
...
+
+        if (!payload) {
+            return { status: "INVALID_EMAIL_ERROR", err: "Email is missing" };
+        }
+
+        return Recipe.getInstanceOrThrowError().recipeInterfaceImpl.registerOptions({
+            requireResidentKey: DEFAULT_REGISTER_OPTIONS_REQUIRE_RESIDENT_KEY,
+            residentKey: DEFAULT_REGISTER_OPTIONS_RESIDENT_KEY,
+            userVerification: DEFAULT_REGISTER_OPTIONS_USER_VERIFICATION,
+            supportedAlgorithmIds: DEFAULT_REGISTER_OPTIONS_SUPPORTED_ALGORITHM_IDS,
```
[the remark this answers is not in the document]
victorbojica: These are overridable only through overrides (by overriding the methods). This should suffice, because the assumption is that only enterprise customers would need to override these. The regular customers would use our defaults.

lib/ts/recipe/webauthn/index.ts
```diff
@@ -0,0 +1,408 @@
...
+    static Error = SuperTokensError;
+
+    static async registerOptions(
+        email: string | undefined,
+        recoverAccountToken: string | undefined,
+        relyingPartyId: string,
+        relyingPartyName: string,
+        origin: string,
+        timeout: number,
+        attestation: "none" | "indirect" | "direct" | "enterprise" = "none",
```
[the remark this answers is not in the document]
victorbojica: If your talking about the recipe config, then yes. The default values are set in the API implementation, but the customer should be able to send them as input as well.

lib/ts/recipe/webauthn/index.ts
```diff
@@ -0,0 +1,408 @@
...
+        | { status: "RECOVER_ACCOUNT_TOKEN_INVALID_ERROR" }
+    > {
+        return Recipe.getInstanceOrThrowError().recipeInterfaceImpl.consumeRecoverAccountToken({
+            token,
+            tenantId: tenantId === undefined ? DEFAULT_TENANT_ID : tenantId,
+            userContext: getUserContext(userContext),
+        });
+    }
+
+    static registerCredential(input: {
```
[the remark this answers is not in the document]
victorbojica: done.

lib/ts/recipe/webauthn/types.ts
victorbojica: done.

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: Done

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: What should be the update? Or am i missing something ?

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: I don't think so. We are identifying the user by email. And since we are signing up, the credential isn't stored yet.

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: Done. 

Was doing it like this, because at the moment of implementation, we hadn't decided on having a method for retrieving the `webauthGeneratedOptions` (which contained the email) and i had to retrieve the email before calling `AuthUtils.getAuthenticatingUserAndAddToCurrentTenantIfRequired`. 
In the last few chats with Rishabh we deciede on adding the new method, but didn't update the API implmentation yet.

As a side note:
The credential can be shared across tenants and is not linked to a tenant.

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: maybe we could have a chat about this.

lib/ts/recipe/webauthn/types.ts
victorbojica: The status is a generic one, for whatever error happens because of the Webauthn credentials flow. If we'd need to be more specific here (not in the API implementation - we should keep it generic there), then we'd need to do the actual core implementation, as there might lots of stuff that we can only find out when we start implementing.

There is another type of status "INVALID_AUTHENTICATOR_ERROR" (along with the reason) which means that the credential has been generated by an "unsupported" authenticator (this could happen by overriding the registerOptions method and overriding the default values for `attestation`, `requireResidentKey`, `residentKey`, `userVerification`, `supportedAlgorithmIds`, etc.).

What do you think ?

lib/ts/recipe/webauthn/recipe.ts
```diff
@@ -0,0 +1,365 @@
...
+                        ];
+                    }
+
+                    // If the list is empty we generate an email address to make the flow where the user is never asked for
+                    // an email address easier to implement. In many cases when the user adds an email-password factor, they
+                    // actually only want to add a password and do not care about the associated email address.
+                    // Custom implementations can choose to ignore this, and ask the user for the email anyway.
+                    if (result.length === 0) {
+                        result.push(`${sessionRecipeUserId.getAsString()}@stfakeemail.supertokens.com`);
+                    }
```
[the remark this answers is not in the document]
victorbojica: maybe we could chat about this.

lib/ts/recipe/webauthn/api/emailExists.ts
```diff
@@ -0,0 +1,51 @@
+/* Copyright (c) 2021, VRAI Labs and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
victorbojica: Done

lib/ts/recipe/webauthn/types.ts
porcellus: The `INVALID_AUTHENTICATOR_ERROR` sounds good. My main issue with `WRONG_CREDENTIALS_ERROR` is that it overlaps with the error status we have for people submitting wrong passwords I guess.

lib/ts/recipe/webauthn/api/implementation.ts
porcellus: I see. So the point of `checkCredentialsOnTenant` is to find out if the user can use the current set of credentials to log in on the given `tenantId` (it can take it as a first parameter). This is tenant independent for pwless and thirdparty, but not for emailpassword for example.

lib/ts/recipe/webauthn/api/implementation.ts
porcellus: well the error codes are login method specific, so we'll have to have new ones I think.

lib/ts/recipe/webauthn/index.ts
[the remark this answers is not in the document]
porcellus: What I meant is that couldn't the ones that appear in the config default to those values? or couldn't we remove those from the interface? do we expect multiple values for these depending on the context?

lib/ts/recipe/webauthn/api/implementation.ts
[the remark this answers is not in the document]
porcellus: I see. Then please update the error message.

lib/ts/recipe/webauthn/api/implementation.ts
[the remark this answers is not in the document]
porcellus: It does make sense for the sign in/sign in options and also kind of makes sense for registerOptions, but I think INVALID_WEBAUTHN_OPTIONS_ID would make more sense here and in cases where we can know that that is the reason it failed (which is the case here I think).

lib/ts/recipe/webauthn/api/implementation.ts
porcellus: so in no cases would this be relevant to this or other linking related functions? if so then the types may require some cleanup

lib/ts/recipe/accountlinking/types.ts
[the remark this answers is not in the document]
porcellus: I think searching by a single credentialId makes sense.

lib/ts/recipe/webauthn/api/implementation.ts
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/api/implementation.ts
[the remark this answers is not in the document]
victorbojica: We'll be using a bit more context in the errors. Also don't need to be overly specific - something like the example above should be fine.

lib/ts/recipe/webauthn/index.ts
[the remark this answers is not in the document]
victorbojica: will use objects as params in all the methods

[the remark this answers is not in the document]
victorbojica: figure out a way of sending request to this method )possibly split the types by req input)

lib/ts/recipe/webauthn/recipe.ts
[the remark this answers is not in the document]
victorbojica: the flow is correct, but:
when using webauthn as an second factor and the first factor doesn't have an email adress, the credential generation would use the webuathn fake email as display - which would look unprofessional. because of this we could use display name as a paramenter for register options and default it to a sane value if its missing.

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: done. added to do.

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: will check if we need to remove webauthn credential ids from the type - it is not used atm

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: reverted to previous version because we don't need to check in the scope of the tenant.

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: same as previous comments. also need to check method to see if ti works with webauthn

lib/ts/recipe/webauthn/types.ts
victorbojica: same as previous error discussion

lib/ts/recipe/webauthn/api/implementation.ts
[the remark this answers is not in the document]
victorbojica: done

lib/ts/recipe/webauthn/recipe.ts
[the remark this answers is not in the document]
victorbojica: done.

lib/ts/recipe/webauthn/api/implementation.ts
victorbojica: done

lib/ts/recipe/webauthn/types.ts
victorbojica: done

lib/ts/recipe/webauthn/index.ts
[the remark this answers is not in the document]
victorbojica: done

PR #964, opened 2024-11-27
Added API testing for missing methods

victorbojica:
## Summary of change

Added basic tests for registerOptions, signInOptions, signIn, signUp.
These tests are for the happy flow. More will be added when the core will be implemented.

The core has been mocked and is using the simplewebauthn library for handling webauthn logic.

For emulating the `navigator.credentials.create` and `navigator.credentials.get` we used a virtual authenticator: https://www.descope.com/blog/post/virtual-webauthn . This has been ported using WebAssembly.


test/webauthn/apis.test.js
```diff
@@ -46,104 +42,730 @@ describe(`apisFunctions: ${printPath("[test/webauthn/apis.test.js]")}`, function
...
+                appInfo: {
+                    apiDomain: "api.supertokens.io",
+                    appName: "SuperTokensplm",
+                    websiteDomain: "supertokens.io",
+                },
+                recipeList: [WebAuthn.init()],
+            });
+
+            // run test if current CDI version >= 2.11
+            if (!(await isCDIVersionCompatible("2.11"))) return;
```
porcellus, line 62: I don't think this is right, it should be something like 5.12?

test/webauthn/apis.test.js
```diff
@@ -46,104 +42,730 @@ describe(`apisFunctions: ${printPath("[test/webauthn/apis.test.js]")}`, function
...
+                request(app)
+                    .post("/auth/webauthn/options/register")
+                    .send({
+                        email: "test@example.com",
+                    })
+                    .expect(200)
+                    .end((err, res) => {
+                        if (err) {
+                            console.log(err);
+                            resolve(undefined);
```
porcellus, line 79: In this case we should be rejecting, resolving with undefined makes debugging tests rather confusing.

test/webauthn/wasm_exec.js
```diff
@@ -0,0 +1,639 @@
+// Copyright 2018 The Go Authors. All rights reserved.
```
porcellus, line 1: Why do we need to use wasm here? Are there no libraries that'd do this natively or abstract this away?

test/webauthn/apis.test.js
```diff
@@ -12,27 +12,23 @@
  * License for the specific language governing permissions and limitations
```
porcellus, line 12: I'm fine with simple and/or node specific tests here, but other than those I'd prefer if these were moved into the backend-sdk-testing repo. This is fine for now if you think it'd be a lot of overhead because of mocking/using nock, but we'll need to move them.

test/webauthn/apis.test.js
```diff
@@ -46,104 +42,730 @@ describe(`apisFunctions: ${printPath("[test/webauthn/apis.test.js]")}`, function
...
+                            resolve(undefined);
+                        } else {
+                            resolve(JSON.parse(res.text));
+                        }
+                    })
+            );
+            assert(signInOptionsResponse.status === "OK");
+
+            const { createAndAssertCredential } = await getWebauthnLib();
+            const credential = createAndAssertCredential(registerOptionsResponse, signInOptionsResponse, {
```
porcellus, line 626: why does `createAndAssertCredential` take both the register and sign in options?
what exactly does it assert?

porcellus (changes requested):

test/webauthn/apis.test.js
victorbojica: It basically both creates  (used for signup)  and verifies a credential (used for login). 
I do it like this because the virtual authenticator doesn't store state and in order to verify (sign in) a credential, the virtual authenticator needs to have the credential stored (similar to how the phone stores the credential after creation). Doing it like this, i generate the credential and also verify using a single method, avoiding storing/handling state.

test/webauthn/apis.test.js
victorbojica: Did it like this because i wanted to have some kind of feedback on the implementation and making sure it works.
We'll move them to the SDK testing repo

test/webauthn/apis.test.js
porcellus: you can just add a TODO to update it, we don't want to skip it right now. I think it's going to be 5.3 (the current one is 5.2)

test/webauthn/apis.test.js
porcellus: I see, that makes it easier to work with wasm. One thing that still confuses me is that we re-use the same attestation for both sign up and in, but I'll try running the tests and check the actual object structure.

test/webauthn/apis.test.js
victorbojica: done

PR #973, opened 2025-01-16
Split listUsersByAccountInfo input type from AccountInfo

victorbojica:
## Summary of change

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2

### DIFF
--- a/lib/ts/recipe/accountlinking/recipe.ts
+++ b/lib/ts/recipe/accountlinking/recipe.ts
@@ -18,7 +18,13 @@ import type { BaseRequest, BaseResponse } from "../../framework";
 import normalisedURLPath from "../../normalisedURLPath";
 import RecipeModule from "../../recipeModule";
 import type { APIHandled, HTTPMethod, NormalisedAppinfo, RecipeListFunction, User, UserContext } from "../../types";
-import type { TypeNormalisedInput, RecipeInterface, TypeInput, AccountInfoWithRecipeId } from "./types";
+import type {
+    TypeNormalisedInput,
+    RecipeInterface,
+    TypeInput,
+    AccountInfoWithRecipeId,
+    AccountInfoInput,
+} from "./types";
 import { validateAndNormaliseUserInput } from "./utils";
 import OverrideableBuilder from "supertokens-js-override";
 import RecipeImplementation from "./recipeImplementation";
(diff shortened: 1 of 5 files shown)


lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -146,13 +152,22 @@ export default class Recipe extends RecipeModule {
...
+            ...user.loginMethods[0],
+            webauthn: undefined,
+        };
+        // any credentialId will identify the same user
+        if (user.loginMethods[0]?.webauthn?.credentialIds.length) {
+            accountInfo.webauthn = {
+                credentialId: user.loginMethods[0].webauthn.credentialIds[0],
+            };
+        }
+        // then, we try and find a primary user based on the email / phone number / third party ID / credentialId.
```
porcellus, line 165: Does it make sense to search by `credentialId` here? I don't think so: we are looking for a user to link to the current recipe user, but any search using the credentialId of the current user "will identify the same user" which is the current one.

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -202,10 +217,20 @@ export default class Recipe extends RecipeModule {
...
+        const accountInfo: AccountInfoInput = {
+            ...user.loginMethods[0],
+            webauthn: undefined,
+        };
+        // any credentialId will identify the same user
+        if (user.loginMethods[0]?.webauthn?.credentialIds.length) {
+            accountInfo.webauthn = {
+                credentialId: user.loginMethods[0].webauthn.credentialIds[0],
+            };
+        }
```
porcellus, line 229: Here it may also not make sense to search using the `credentialId`, since it'll only match the current user which is already a part of the results because the email address is already present.

lib/ts/recipe/accountlinking/recipe.ts
```diff
@@ -322,9 +347,21 @@ export default class Recipe extends RecipeModule {
...
+        const accountInfoInput: AccountInfoInput = {
+            ...accountInfo,
+            webauthn: undefined,
+        };
+        // any credentialId will identify the same user
+        if (accountInfo.webauthn?.credentialIds.length) {
+            accountInfoInput.webauthn = {
+                credentialId: accountInfo.webauthn.credentialIds[0],
+            };
+        }
```
porcellus, line 361: Here it may also not make sense to search using the credentialId, since it'll only match the current user which is already a part of the results because the email address is already present.

porcellus (changes requested):

lib/ts/recipe/accountlinking/recipe.ts
victorbojica: Done.

lib/ts/recipe/accountlinking/recipe.ts
victorbojica: Done

lib/ts/recipe/accountlinking/recipe.ts
victorbojica: Done

Issue #972, opened 2025-01-10
Allow extra form fields in sign in and password reset flows


porcellus:

vitusan:
🚀 Feature
The recipe of EmailPassword should allow for signin form field customizations, so additional fields can be supplied and accessed where needed.

Implementation details
In my use case, there's a need to know additional data from the user. The frontend js sdk allows for signin formfield customization, but the backend only allows for signup formfield customization.

To achieve my desired use case, i'm forced to overwrite the signin POST endpoint and perform the signin function on my own. Not a big deal, but it would be nice to use the native function from the Supertokens backend SDK (NodeJS), cause it has more support to other features.

PR #976, opened 2025-02-17
feat: Adds signInFeature type and validation for form fields in EmailPassword Recipe

vitusan:
## Summary of change

Introduce a new `signInFeature` type with validation for form fields, ensuring proper handling of email and password fields. Update related functions to accommodate the new feature. Why? Because my use case needs to send additional data to the SignIn POST endpoint and that made things complicated for the only way to achieve this was to override the SignInPost.

## Related issues

- https://github.com/supertokens/supertokens-node/issues/447
- https://github.com/supertokens/supertokens-node/issues/972

## Test Plan

npm run test. All tests have passed. 

## Documentation changes

- [ ] Mention in EmailPassword Recipe that is also possible to change the signIn Form fields, not just the signUp Form fields.

-   [X] Changelog has been updated
-   [X] Changes to the version if needed
    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [X] Had run `npm run build-pretty`
-   [X] Had installed and ran the pre-commit hook


lib/ts/version.ts
```diff
@@ -12,7 +12,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-export const version = "21.1.0";
+export const version = "21.1.1";
```
porcellus, line 15: This will be a minor (not patch) level change.

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -114,25 +119,58 @@ function validateAndNormaliseResetPasswordUsingTokenConfig(
...
-    return formFields
-        .filter((filter) => filter.id === FORM_FIELD_EMAIL_ID || filter.id === FORM_FIELD_PASSWORD_ID)
-        .map((field) => {
-            return {
-                id: field.id,
-                // see issue: https://github.com/supertokens/supertokens-node/issues/36
-                validate: field.id === FORM_FIELD_EMAIL_ID ? field.validate : defaultValidator,
-                optional: false,
-            };
+function normaliseSignInFormFields(formFields?: TypeInputFormField[]) {
```
porcellus, line 122: If done this way, it'd be a breaking change: if someone added an override for their email validators in the sign up config, they'd expect it to be applied here. I'd prefer if we kept it that way as well.

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -114,25 +119,58 @@ function validateAndNormaliseResetPasswordUsingTokenConfig(
...
+    let normalisedFormFields: NormalisedFormField[] = [];
+    if (formFields !== undefined) {
+        formFields.forEach((field) => {
+            if (field.id === FORM_FIELD_PASSWORD_ID) {
+                normalisedFormFields.push({
+                    id: field.id,
+                    validate: field.validate === undefined ? defaultPasswordValidator : field.validate,
```
porcellus, line 129: In general, you shouldn't have a validator for the password field during sign-in, since we'd want to check the DB in most cases anyway to prevent cases where a password validation change is preventing people from logging in. I don't mind giving people the option to define this specifically, but by default it should let everything pass.

```suggestion
                    validate: field.validate === undefined ? defaultValidator : field.validate,
```

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -114,25 +119,58 @@ function validateAndNormaliseResetPasswordUsingTokenConfig(
...
+            } else {
+                normalisedFormFields.push({
+                    id: field.id,
+                    validate: field.validate === undefined ? defaultValidator : field.validate,
+                    optional: field.optional === undefined ? false : field.optional,
+                });
+            }
         });
+    }
+    if (normalisedFormFields.filter((field) => field.id === FORM_FIELD_PASSWORD_ID).length === 0) {
```
porcellus, line 147:
```suggestion
    if (!normalisedFormFields.some((field) => field.id === FORM_FIELD_PASSWORD_ID)) {
```

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -114,25 +119,58 @@ function validateAndNormaliseResetPasswordUsingTokenConfig(
...
+        // no password field give by user
+        normalisedFormFields.push({
+            id: FORM_FIELD_PASSWORD_ID,
+            validate: defaultPasswordValidator,
+            optional: false,
+        });
+    }
+    if (normalisedFormFields.filter((field) => field.id === FORM_FIELD_EMAIL_ID).length === 0) {
```
porcellus, line 155:
```suggestion
    if (!normalisedFormFields.some((field) => field.id === FORM_FIELD_EMAIL_ID)) {
```

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -114,25 +119,58 @@ function validateAndNormaliseResetPasswordUsingTokenConfig(
...
+                    optional: field.optional === undefined ? false : field.optional,
+                });
+            }
         });
+    }
+    if (normalisedFormFields.filter((field) => field.id === FORM_FIELD_PASSWORD_ID).length === 0) {
+        // no password field give by user
+        normalisedFormFields.push({
+            id: FORM_FIELD_PASSWORD_ID,
+            validate: defaultPasswordValidator,
```
porcellus, line 151:
```suggestion
            validate: defaultValidator,
```

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -114,25 +119,58 @@ function validateAndNormaliseResetPasswordUsingTokenConfig(
...
+            optional: false,
+        });
+    }
+    if (normalisedFormFields.filter((field) => field.id === FORM_FIELD_EMAIL_ID).length === 0) {
+        // no email field give by user
+        normalisedFormFields.push({
+            id: FORM_FIELD_EMAIL_ID,
+            validate: defaultEmailValidator,
```
porcellus, line 159: This should fall back to the (normalized) email field validator for sign-up to avoid breaking apps.

porcellus (changes requested):

porcellus:
Hi, thanks for your contribution. I think we should be able to get this merged and released next week :)

PR #987, opened 2025-04-03
feat: inferring pkce based on OIDC response

deepjyoti30-st:
## Summary of change

This PR adds support for inferring whether pkce should be used or not based on the OIDC response.

## Test Plan

All OIDC related tests and existing tests should pass

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/version.ts
+++ b/lib/ts/version.ts
@@ -12,7 +12,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-export const version = "22.0.1";
+export const version = "22.1.0";
 
 export const cdiSupported = ["5.3"];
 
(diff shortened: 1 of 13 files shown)


lib/ts/version.ts
```diff
@@ -12,7 +12,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-export const version = "21.1.1";
+export const version = "21.1.2";
```
porcellus, line 15: This should be a minor version bump, plus please rebase this on top of latest.

porcellus (changes requested):

deepjyoti30-st (commented):
Leaving reply

Issue #1007, opened 2025-06-11
Session Refresh doesn't work if request has an unrelated, unencoded cookie in it


shankj3:
Session refresh and sign out will fail if the request has a cookie attached to it that has a character that would fail `decodeURIComponent`, e.g. a cookie value with a `%` in it. 

https://github.com/supertokens/supertokens-node/blob/8fabd099f11bf4d375dacf2ccfea028d598f9261/lib/build/recipe/session/cookieAndHeaders.js#L259

Though it is common practice to have cookie values be safely uri decoded, that isn't in the spec for cookies. Supertokens shouldn't really be attempting to do anything with cookies that are unrelated to any of its logic, so ideally this won't be processed in this map at all. 

If a user gets into this state, their only solution is to clear their cookies and hope that another cookie with this value doesn't get added again. 

sdk version: 20.1.5, but i see this code snippet in all the recent versions.

PR #1009, opened 2025-06-17
fix: session cookie parsing optimizations

deepjyoti30-st:
## Summary of change

This PR refactors the function used for parsing incoming request cookies and optimizes it to avoid parsing cookies (since we don't care about the value).

## Related issues

- #1007 

## Test Plan

- All tests in backend-sdk-testing should pass

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/session/cookieAndHeaders.ts
```diff
@@ -286,24 +286,19 @@ function parseCookieStringFromRequestHeaderAllowingDuplicates(cookieString: stri
     const cookiePairs = cookieString.split(";");
```
porcellus, line 286: Since this function is right above and since we do not decode the values, I'd prefer if this just returned the list of cookie names to avoid later issues if someone were to use this function without knowing the full context.
Something like:
`getCookieNamesFromRequestHeaderAllowingDuplicates`

porcellus (changes requested):

lib/ts/recipe/session/cookieAndHeaders.ts
deepjyoti30-st: I can also refactor the function to accept the name as a parameter and check and return a bool instead. I think that will be more suitable since we don't really use the cookies in the parent at all.

porcellus (approved):

porcellus:
### Merge activity

* **Jun 19, 2:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/supertokens/supertokens-node/1009).

